PR #28, opened 2020-02-26
Metrics endpoint

jframe:
Metrics endpoint that provides metrics in prometheus format. Includes standard Prometheus JVM and system metrics plus http metrics.

Includes http request_time metric that gives the time taken to process a http request.

### DIFF
--- a/commandline/build.gradle
+++ b/commandline/build.gradle
@@ -19,6 +19,7 @@ dependencies {
   implementation 'org.apache.logging.log4j:log4j-api'
   implementation 'org.apache.logging.log4j:log4j-core'
   implementation 'org.hyperledger.besu:plugin-api'
+  implementation 'org.hyperledger.besu.internal:metrics-core'
 
   runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
 
--- a/commandline/build.gradle
+++ b/commandline/build.gradle
@@ -18,6 +18,7 @@ dependencies {
   implementation 'com.google.guava:guava'
   implementation 'org.apache.logging.log4j:log4j-api'
   implementation 'org.apache.logging.log4j:log4j-core'
+  implementation 'org.hyperledger.besu:plugin-api'
 
   runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
 
--- a/commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
+++ b/commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
@@ -12,14 +12,9 @@
  */
 package tech.pegasys.eth2signer.commandline;
 
-import tech.pegasys.eth2signer.commandline.convertor.MetricCategoryConverter;
-import tech.pegasys.eth2signer.core.metrics.Eth2SignerMetricCategory;
-
 import java.io.PrintWriter;
 
 import org.apache.logging.log4j.Level;
-import org.hyperledger.besu.metrics.StandardMetricCategory;
-import org.hyperledger.besu.plugin.services.metrics.MetricCategory;
 import picocli.CommandLine;
 import picocli.CommandLine.Help.Ansi;
 import picocli.CommandLine.ParameterException;
@@ -49,16 +44,7 @@ public class CommandlineParser {
     commandLine.setExecutionExceptionHandler(this::handleExecutionException);
     commandLine.setParameterExceptionHandler(this::handleParseException);
 
-    registerConverters(commandLine);
-
     return commandLine.execute(args);
-  }
-
-  private void registerConverters(final CommandLine commandLine) {
-    final MetricCategoryConverter metricCategoryConverter = new MetricCategoryConverter();
-    metricCategoryConverter.addCategories(Eth2SignerMetricCategory.class);
-    metricCategoryConverter.addCategories(StandardMetricCategory.class);
-    commandLine.registerConverter(MetricCategory.class, metricCategoryConverter);
   }
 
   private int handleParseException(final ParameterException ex, final String[] args) {
--- a/commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
+++ b/commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
@@ -12,9 +12,14 @@
  */
 package tech.pegasys.eth2signer.commandline;
 
+import tech.pegasys.eth2signer.commandline.convertor.MetricCategoryConverter;
+import tech.pegasys.eth2signer.core.metrics.Eth2SignerMetricCategory;
+
 import java.io.PrintWriter;
 
 import org.apache.logging.log4j.Level;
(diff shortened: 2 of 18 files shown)


commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
```diff
@@ -24,6 +29,7 @@
   private final Eth2SignerCommand baseCommand;
   private final PrintWriter outputWriter;
   private final PrintWriter errorWriter;
+  private final MetricCategoryConverter metricCategoryConverter = new MetricCategoryConverter();
```
rain-on: nit: does this need to be a class level variable? Or can it be constructed in registerConverters?

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -56,27 +59,43 @@ public Runner(final Config config) {
...
+            config.getMetricCategories(),
+            vertx);
     try {
-      final Handler<HttpServerRequest> requestHandler = createRouter(vertx);
+      final Handler<HttpServerRequest> requestHandler =
+          createRouter(vertx, metricsEndpoint.getMetricsSystem());
       final HttpServer httpServer = createServerAndWait(vertx, requestHandler);
       LOG.info("Server is up, and listening on {}", httpServer.actualPort());
 
+      metricsEndpoint.start();
```
rain-on: nit: is there a reason you star the metrics endpoint _after_ the server is started? I'd have thought we'd start metrics first ... 🤷‍♂

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -56,27 +59,43 @@ public Runner(final Config config) {
...
       persistPortInformation(httpServer.actualPort());
     } catch (final ExecutionException | InterruptedException e) {
       vertx.close();
       throw new RuntimeException("Failed to create Http Server", e.getCause());
     } catch (final Throwable t) {
+      metricsEndpoint.stop();
```
rain-on: nit:metricsEndpoing _may_ not have started - is that a problem?

core/src/main/java/tech/pegasys/eth2signer/core/http/ResponseTimeMetricsHandler.java
```diff
@@ -0,0 +1,43 @@
...
+            Eth2SignerMetricCategory.HTTP,
+            "request_time",
+            "Time taken to process a http request",
+            "endpoint");
+  }
+
+  @Override
+  public void handle(final RoutingContext context) {
+    final TimingContext timingContext = labelledTimer.labels(context.normalisedPath()).startTimer();
+    context.addBodyEndHandler(handler -> timingContext.stopTimer());
```
rain-on: nit: there appears to be a risk that this won't get called - will that make the timingContext create very a bad metric?

core/src/main/java/tech/pegasys/eth2signer/core/http/ResponseTimeMetricsHandler.java
```diff
@@ -0,0 +1,43 @@
...
+    context.next();
```
rain-on: nit: do we need to put context.next() on all of our handlers?

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
jframe: no particular reason. don't think the order matters at all. is there a good reason to start it first?

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
jframe: No shouldn't be a problem. The Besu MetricsHttpService does a null check on the internal Vertx httpServer before attempting to do anything.

core/src/main/java/tech/pegasys/eth2signer/core/http/ResponseTimeMetricsHandler.java
jframe: No only special handlers like one and logging etc. where you want the router to try continue matching and handling off to other handlers. If we are completely handling the request there is no need to call next and when combining them.

core/src/main/java/tech/pegasys/eth2signer/core/http/ResponseTimeMetricsHandler.java
jframe: not sure on this one. In what cases would the endBodyHandler not be called? I'll have to think about this case some more.

commandline/build.gradle
```diff
@@ -18,6 +18,8 @@ dependencies {
   implementation 'com.google.guava:guava'
   implementation 'org.apache.logging.log4j:log4j-api'
   implementation 'org.apache.logging.log4j:log4j-core'
+  implementation 'org.hyperledger.besu:plugin-api'
```
rain-on: why plugin-api? That seems ... strange ..

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
```diff
@@ -44,9 +49,18 @@ public int parseCommandLine(final String... args) {
     commandLine.setExecutionExceptionHandler(this::handleExecutionException);
     commandLine.setParameterExceptionHandler(this::handleParseException);
 
+    registerConverters(commandLine);
+
     return commandLine.execute(args);
   }
 
+  private void registerConverters(final CommandLine commandLine) {
+    final MetricCategoryConverter metricCategoryConverter = new MetricCategoryConverter();
```
rain-on: nit: not sure how else to do this - but having the converters registered somewhere apart from the Eth2SignerCommand feels odd - i.e. it's the types in Eth2SignerCommand which require converts, so having the logic here (instead of there) - feels messy (happy to be wrong, and not something I'm concerned about, just a smell)

commandline/build.gradle
jframe: The underlying type MetricCategory is defined in the plugin-api so we need both this and Besu metrics module unfortunately.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
jframe: It is a bit messy, but the registration needs to happen on the CommandLine. Perhaps passing in convertors to the CommandlineParser would be better? At least then the definitions could stay in the Eth2SignerCommand.

jframe: Found a better solution using the converter property on the Option annotation in the Eth2SignerCommand

PR #29, opened 2020-03-02
Multi key yaml

jframe:
Change the toml validator key loading to use a yaml file format instead.


acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/UnencryptedTomlKeyLoadAndSignAcceptanceTest.java
```diff
@@ -33,7 +33,7 @@
 
 public class UnencryptedTomlKeyLoadAndSignAcceptanceTest extends AcceptanceTestBase {
```
rain-on: rename to remove "toml" from classname

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/UnencryptedTomlKeyLoadAndSignAcceptanceTest.java
```diff
@@ -46,9 +46,9 @@
   @ValueSource(strings = {"/signer/block", "/signer/attestation"})
   public void signDataWithKeyLoadedFromUnencryptedFile(final String artifactSigningEndpoint)
       throws Exception {
-    final String tomlKeyFilename = keyPair.publicKey().toString().substring(2);
-    final Path keyConfigFile = testDirectory.resolve(tomlKeyFilename + ".toml");
-    tomlHelpers.createUnencryptedTomlFileAt(keyConfigFile, privateKeyString);
+    final String metadataKeyFilename = keyPair.publicKey().toString().substring(2);
```
rain-on: nit: if this were configFilename, then there's no impression that it needs to be YAML (it was badly named before too)

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
```diff
@@ -30,23 +30,23 @@
 
   private static final Logger LOG = LogManager.getLogger();
 
-  private final SigningMetadataTomlConfigLoader signingMetadataTomlConfigLoader;
+  private final SigningMetadataConfigLoader signingMetadataConfigLoader;
```
rain-on: nit: can we genericise this further - eg configLoader?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/SigningMetadataConfigLoader.java
```diff
@@ -0,0 +1,88 @@
...
+  public SigningMetadataConfigLoader(
+      final Path rootDirectory,
+      final String fileExtension,
+      final SigningMetadataFileProvider metadataFileProvider) {
+    this.configsDirectory = rootDirectory;
+    this.metadataFileProvider = metadataFileProvider;
+    this.fileExtension = fileExtension;
+  }
+
+  Optional<SigningMetadataFile> loadMetadataForAddress(final String address) {
```
rain-on: few things:
1. does this need explicit visibility (public)?
2. Can we abstract away from "address"? I.e. eth2signer doens't know about addresses ... its more identifiers (I think?)
3. There's a smell, that this returns a SigningMetadataFILE - but is a loader ... shouldn't it provide just SigningMetadata? Either the return type is not quite right, or the name of the function is more "getMetadataFile(final String identifier)

core/src/main/java/tech/pegasys/eth2signer/core/multikey/SigningMetadataConfigLoader.java
```diff
@@ -0,0 +1,88 @@
...
+  Collection<SigningMetadataFile> loadAvailableSigningMetadataConfigs() {
+    final Collection<SigningMetadataFile> metadataConfigs = new HashSet<>();
+
+    try (final DirectoryStream<Path> directoryStream =
+        Files.newDirectoryStream(configsDirectory, "**." + fileExtension)) {
+      for (final Path file : directoryStream) {
+        metadataFileProvider.getMetadataInfo(file).ifPresent(metadataConfigs::add);
+      }
+      return metadataConfigs;
+    } catch (final IOException e) {
```
rain-on: do you want to do this check _inside_ the loop, such that you load all valid files - even if one in the middle is 'bad'?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SigningMetadataFile.java
```diff
@@ -19,19 +19,19 @@
 
   protected String baseFilename;
 
-  public SigningMetadataFile(final String filename) {
-    this.baseFilename = getFilenameWithoutExtension(filename);
+  public SigningMetadataFile(final String filename, final String fileExtension) {
```
rain-on: smell: maybe filename should be renamed basename? This class seems to do validation that the input data was correct .. which, feels like you're expecting people to use it incorrectly - which should be hard to do.

Maybe it should be constructed with a Path - then use org.apache.commons.io.FilenameUtils?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/YamlSigningMetadataFileProvider.java
```diff
@@ -0,0 +1,71 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+
+public class YamlSigningMetadataFileProvider implements SigningMetadataFileProvider {
+  private static final Logger LOG = LogManager.getLogger();
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
```
rain-on: nit: must this be static?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/YamlSigningMetadataFileProvider.java
```diff
@@ -0,0 +1,71 @@
...
+
+  static {
+    OBJECT_MAPPER.enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS);
+  }
+
+  @Override
+  public Optional<SigningMetadataFile> getMetadataInfo(final Path file) {
+    try {
+      final MetadataFileBody metaDataInfo =
+          OBJECT_MAPPER.readValue(file.toFile(), MetadataFileBody.class);
```
rain-on: nit: seeing file.toFile() seems odd - is the function parameter well named?

core/src/test/java/tech/pegasys/eth2signer/core/multikey/MetadataFileFixture.java
```diff
@@ -0,0 +1,30 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.eth2signer.core.multikey;
+
+public class MetadataFileFixture {
+
+  public static final String CONFIG_FILE_EXTENSION = ".yaml";
+  public static String LOWERCASE_ADDRESS = "627306090abab3a6e1400e9345bc60c78a8bef57";
+
+  public static String PREFIX_MIXEDCASE_ADDRESS = "f17f52151ebef6c7334fad080c5704d77216b732";
```
rain-on: nit: the content of this doesn't _appear_ to be mixed_case, is it meant to be?

core/src/test/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProviderTest.java
```diff
@@ -0,0 +1,82 @@
...
+  void getSignerForAvailableMetadataReturnsSigner() {
+    when(loader.loadMetadataForAddress(PUBLIC_KEY)).thenReturn(Optional.of(metadataFile));
+
+    final Optional<ArtifactSigner> signer = signerFactory.getSigner(PUBLIC_KEY);
+    assertThat(signer).isNotEmpty();
+    assertThat(signer.get().getIdentifier()).isEqualTo("0x" + PUBLIC_KEY);
+  }
+
+  @Test
+  void getAddresses() {
```
rain-on: nit:rename? eg getAddressesReturnsAllAvailableKeyIdentifiers?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/UnencryptedTomlKeyLoadAndSignAcceptanceTest.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: It's still specific to signing metadata config files so don't think it can be

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SigningMetadataFile.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/SigningMetadataConfigLoader.java
jframe: I think we want this and an additional check around the directory stream iterator and perhaps around around the metadataFileProvider.getMetadataInfo too.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/SigningMetadataConfigLoader.java
jframe: 1. It's got package level access, don't see how we can restrict this any further
2. Think it's more public keys than addresses? Would that be better?
3. I guess it is a bit odd. This is existing code, didn't think too much about the code I copied.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/YamlSigningMetadataFileProvider.java
jframe: No but it's better to be static as creating the object mapper is a little bit expensive. Not something we want create everytime ideally

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/YamlSigningMetadataFileProvider.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
```diff
@@ -12,72 +12,140 @@
...
+import org.apache.commons.io.FilenameUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-public class MultiKeyArtifactSignerProvider implements ArtifactSignerProvider, MultiSignerFactory {
+public class MultiKeyArtifactSignerProvider implements ArtifactSignerProvider {
 
   private static final Logger LOG = LogManager.getLogger();
+  private final Path configsDirectory;
+  private SignerParser signerParser;
```
rain-on: nit: final?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
```diff
@@ -12,72 +12,140 @@
...
-        .map(metadataFile -> metadataFile.createSigner(this));
+    final String normalisedIdentifier = normaliseIdentifier(signerIdentifier);
+    final Optional<ArtifactSignerWithFileName> signer =
+        loadSignerForIdentifier(normalisedIdentifier);
+    if (signer.isEmpty()) {
+      LOG.error("No matching metadata file found for the identifier {}", signerIdentifier);
+      return Optional.empty();
+    }
+    final ArtifactSignerWithFileName signerWithFileName = signer.get();
+    if (!signerMatchesIdentifier(signerWithFileName, signerIdentifier)) {
```
rain-on: nit: I suspect this is the check which is wrong - the real verification is "does loaded Signer have the same identifier as that requested?"
The concept of "filenames" is irrelevant here.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
```diff
@@ -12,72 +12,140 @@
...
+
+  private Collection<ArtifactSignerWithFileName> findSigners(
+      final DirectoryStream.Filter<? super Path> filter) {
+    final Collection<ArtifactSignerWithFileName> signers = new HashSet<>();
+
+    try (final DirectoryStream<Path> directoryStream =
+        Files.newDirectoryStream(configsDirectory, filter)) {
+      for (final Path file : directoryStream) {
+        try {
+          final ArtifactSignerWithFileName artifactSignerWithFileName =
```
rain-on: confirming - the filter above means we don't attempt to load every signer?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
```diff
@@ -12,72 +12,140 @@
...
-    // strip leading 0x from the address.
-    final String signerAddress = signer.getIdentifier().substring(2).toLowerCase();
-    if (!metadataFile.getBaseFilename().toLowerCase().endsWith(signerAddress)) {
-      LOG.error(
-          String.format(
-              "Signer's Ethereum Address (%s) does not align with metadata filename (%s)",
-              signerAddress, metadataFile.getBaseFilename()));
-      return false;
+  private String normaliseIdentifier(final String signerIdentifier) {
+    return signerIdentifier.startsWith("0x") ? signerIdentifier.substring(2) : signerIdentifier;
```
rain-on: do we care about lowercase?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 ConsenSys AG.
+ * Copyright 2019 ConsenSys AG.
```
rain-on: ? should this be 2020? I.e. when was this file created?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
```diff
@@ -12,26 +12,28 @@
...
-  public Bytes getPrivateKeyBytes() {
-    return privateKeyBytes;
+  @JsonProperty(value = "privateKey")
+  public String getPrivateKey() {
+    return privateKey;
   }
 
   @Override
-  public ArtifactSigner createSigner(final MultiSignerFactory factory) {
+  public ArtifactSigner createSigner(final ArtifactSignerFactory factory) {
```
rain-on: ? still doing double dispatch?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SignerParser.java
```diff
@@ -10,11 +10,13 @@
...
-import tech.pegasys.eth2signer.core.multikey.metadata.UnencryptedKeyMetadataFile;
 import tech.pegasys.eth2signer.core.signing.ArtifactSigner;
 
-public interface MultiSignerFactory {
-  ArtifactSigner createSigner(UnencryptedKeyMetadataFile metadataFile);
+import java.nio.file.Path;
+
+public interface SignerParser {
+
+  ArtifactSigner parse(final Path file) throws SigningMetadataException;
```
rain-on: no need for "final" on an interface (apparently)

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
```diff
@@ -12,72 +12,140 @@
...
+
+    try (final DirectoryStream<Path> directoryStream =
+        Files.newDirectoryStream(configsDirectory, filter)) {
+      for (final Path file : directoryStream) {
+        try {
+          final ArtifactSignerWithFileName artifactSignerWithFileName =
+              new ArtifactSignerWithFileName(file, signerParser.parse(file));
+          signers.add(artifactSignerWithFileName);
+        } catch (Exception e) {
+          LOG.error("Error parsing signing metadata file {}", file, e);
```
rain-on: nit: wondering - if you have 2 files which match the filter, but one of them fails - should you keep working? Or is the fact that more than 1 file matched a "fail" condition?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: It's not actually checking the filename, just poorly named variable. I've made a tuple with the signer and filename used internally so I can display details of the failure to the user.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: yes that's right, it creates signers that match the identifier

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
jframe: Yeah it's simpler to test with the double dispatch in place, but I can get rid of it.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SignerParser.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: I think you should keep working and attempt to load the next file

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: Yes I think we do. done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/MultiKeyArtifactSignerProvider.java
jframe: yes that's right, it creates signers that match the identifier

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
jframe: got rid it

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SigningMetadata.java
```diff
@@ -0,0 +1,27 @@
+/*
```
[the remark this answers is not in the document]
rain-on: actually - don't care enough. 👍

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
```diff
@@ -0,0 +1,44 @@
...
+  private final String privateKey;
+
+  @JsonCreator
+  public FileRawSigningMetadata(
+      @JsonProperty(value = "privateKey", required = true) final String privateKey) {
+    this.privateKey = privateKey;
+  }
+
+  @JsonProperty(value = "privateKey")
+  public String getPrivateKey() {
```
rain-on: is this getter required?

rain-on: is this getter required?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
```diff
@@ -0,0 +1,44 @@
...
+    return privateKey;
+  }
+
+  @Override
+  public ArtifactSigner createSigner() {
+    final Bytes privateKey1 = Bytes.fromHexString(getPrivateKey());
```
rain-on: maybe privateKeyBytes rather than privateKey1?

If this fails, because the string is invalid - it won't be found at parse, but at usage - is that bad?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
```diff
@@ -0,0 +1,44 @@
...
+    final KeyPair keys = new KeyPair(SecretKey.fromBytes(privateKey1));
```
rain-on: same as above.

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
```diff
@@ -0,0 +1,209 @@
...
+  private static final String PRIVATE_KEY =
+      "000000000000000000000000000000003ee2224386c82ffea477e2adf28a2929f5c349165a4196158c7f3a2ecca40f35";
+
+  private ArtifactSigner artifactSigner;
+  private DirectoryBackedArtifactSignerProvider signerProvider;
+
+  @BeforeEach
+  void setup() {
+    signerProvider = new DirectoryBackedArtifactSignerProvider(configsDirectory, signerParser);
+    artifactSigner = createArtifactSigner(PRIVATE_KEY);
```
rain-on: nit: move this to an initialiser

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
```diff
@@ -0,0 +1,209 @@
...
+    assertThat(identifiers).hasSize(3);
+    assertThat(identifiers).containsOnly(publicKey1, publicKey2, publicKey3);
+  }
+
+  private Path pathEndsWith(final String endsWith) {
+    return argThat((Path path) -> path != null && path.endsWith(endsWith + FILE_EXTENSION));
+  }
+
+  @SuppressWarnings("ResultOfMethodCallIgnored")
+  private void createFile(final String filename) throws IOException {
```
rain-on: nit: rename to "createFileInConfigsDirectory" (to be explicit about where it is being created (or else, createConfigFile)

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
```diff
@@ -0,0 +1,209 @@
...
+    when(signerParser.parse(any())).thenReturn(artifactSigner);
+
+    final Optional<ArtifactSigner> signer = signerProvider.getSigner("0x" + PUBLIC_KEY);
+
+    assertThat(signer).isNotEmpty();
+    verify(signerParser).parse(pathEndsWith(metadataFilename));
+  }
+
+  @Test
+  void signerReturnedWhenHasUpperCaseHexPrefix() throws IOException {
```
rain-on: nit: is this really a valid test? I.e. hex is deliberately 0x (not 0X), I thought ...

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
jframe: whoops, I that was meant to be privateKeyBytes

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
jframe: done

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
jframe: done

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
jframe: It should be, but I was just thinking about case insensitive filesystems where it might be returned as 0X being part of the filename

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/FileRawSigningMetadata.java
jframe: Including the conversion as part of the constructor now so if it's invalid we throw a SigningMetadataException

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -0,0 +1,155 @@
...
+    this.signerParser = signerParser;
+  }
+
+  @Override
+  public Optional<ArtifactSigner> getSigner(final String signerIdentifier) {
+    final String normalisedIdentifier = normaliseIdentifier(signerIdentifier);
+    final Optional<ArtifactSignerWithFileName> signer =
+        loadSignerForIdentifier(normalisedIdentifier);
+    if (signer.isEmpty()) {
+      LOG.error("No matching metadata file found for the identifier {}", signerIdentifier);
```
rain-on: nit: no _valid_ matching metadata?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SigningMetadataModule.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public SigningMetadataModule() {
+    super("SigningMetadata");
+    addDeserializer(SecretKey.class, new PrivateKeyDeserializer());
+  }
+
+  private static class PrivateKeyDeserializer extends JsonDeserializer<SecretKey> {
+
+    @Override
+    public SecretKey deserialize(JsonParser p, DeserializationContext ctxt) {
```
rain-on: nit final on params (can we turn this back on in gradle/errorprone?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
jframe: yeah that is more correct. done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SigningMetadataModule.java
jframe: done

PR #39, opened 2020-03-20
log request path

jframe:


jframe:
> TBH - its probably a debug log - otherwise it could spam us to death

done

PR #51, opened 2020-03-26
Include public key in url path

jframe:
Includes public key in url path so that the API can be more easily used by load balancers.

See issue https://github.com/PegaSysEng/eth2signer/issues/49

This will update the API to be:
```
/signer/block/<publickey>
/signer/attestation/<publickey>
/signer/randao_reveal/<publickey>
```

```
{
	“signingRoot”: <hex encoded string>
}
```

### DIFF
--- a/core/src/main/java/tech/pegasys/eth2signer/core/http/SigningRequestHandler.java
+++ b/core/src/main/java/tech/pegasys/eth2signer/core/http/SigningRequestHandler.java
@@ -30,6 +30,8 @@ import org.apache.tuweni.bytes.Bytes;
 
 public class SigningRequestHandler implements Handler<RoutingContext> {
 
+  public static final String SIGNER_PATH_REGEX =
+      "/signer/(?<signerType>attestation|block|randao_reveal)/(?<publicKey>.*)";
   private static final Logger LOG = LogManager.getLogger();
   private final ArtifactSignerProvider signerProvider;
   private final JsonDecoder jsonDecoder;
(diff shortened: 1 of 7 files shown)


core/src/main/java/tech/pegasys/eth2signer/core/http/SigningRequestHandler.java
```diff
@@ -43,11 +43,12 @@ public SigningRequestHandler(
   @Override
   public void handle(final RoutingContext context) {
     LOG.debug("Received a request for {}", context.normalisedPath());
-    generateResponseFromBody(context.response(), context.getBody());
+    final String publicKey = context.pathParam("param1");
```
usmansaleem: where does the "param1" comes from? Is it specific with VertX?

jframe: yes, that's how vertx names the params

usmansaleem: 👍

rain-on: I think the regex needs to live in this class as a static variable.

core/src/main/java/tech/pegasys/eth2signer/core/http/SigningRequestHandler.java
```diff
@@ -61,14 +62,14 @@ private void generateResponseFromBody(
           .end("Request body illegally formatted for signing operation.");
       return;
     }
-    final Optional<ArtifactSigner> signer = signerProvider.getSigner(signingRequest.publicKey());
+    final Optional<ArtifactSigner> signer = signerProvider.getSigner(publicKey);
```
rain-on: nit: you _could_ move this up into the "handle" function - i.e. this function doesn't need to know about the public key, just a signer.

PR #56, opened 2020-04-01
Cache signing keys

jframe:
Caches keys so that they don't need to be retrieved on every signing request.

* pre-loads all the keys into the cache on startup
* adds a new config option for the cache limit defaulting to 1000
* will cache signers if they are not already in the cache
* uses a LRU for evicting signers after reaching the size limit

### DIFF
--- a/core/src/integrationTest/java/tech/pegasys/eth2signer/core/DirectoryBackedArtifactSigningProviderIntegrationTest.java
+++ b/core/src/integrationTest/java/tech/pegasys/eth2signer/core/DirectoryBackedArtifactSigningProviderIntegrationTest.java
@@ -46,13 +46,10 @@ import org.junit.jupiter.api.io.TempDir;
 public class DirectoryBackedArtifactSigningProviderIntegrationTest {
 
   @TempDir Path configsDirectory;
-  private SignerParser signerParser;
   private static final String FILE_EXTENSION = "yaml";
   private static ObjectMapper YAML_OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
 
   private DirectoryBackedArtifactSignerProvider signerProvider;
-  private ArtifactSignerFactory artifactSignerFactory;
-  private HashicorpConnectionFactory hashicorpConnectionFactory;
   private Vertx vertx;
   private TrackingLogAppender logAppender = new TrackingLogAppender();
   private final Logger logger =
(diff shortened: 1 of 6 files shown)


core/src/integrationTest/java/tech/pegasys/eth2signer/core/DirectoryBackedArtifactSigningProviderIntegrationTest.java
```diff
@@ -64,14 +61,16 @@
...
-    artifactSignerFactory =
+    final ArtifactSignerFactory artifactSignerFactory =
         new ArtifactSignerFactory(
             configsDirectory, new NoOpMetricsSystem(), hashicorpConnectionFactory);
-    signerParser = new YamlSignerParser(artifactSignerFactory);
+    final SignerParser signerParser = new YamlSignerParser(artifactSignerFactory);
     signerProvider =
-        new DirectoryBackedArtifactSignerProvider(configsDirectory, FILE_EXTENSION, signerParser);
+        new DirectoryBackedArtifactSignerProvider(
+            configsDirectory, FILE_EXTENSION, signerParser, 0);
```
usmansaleem, line 73: This 0 is cache size?

jframe: Yeah 0 for cache size disables the cache, I don't want the cache interfering with the integration tests.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -40,47 +47,79 @@
...
+      final Path rootDirectory,
+      final String fileExtension,
+      final SignerParser signerParser,
+      final long maxSize) {
     this.configsDirectory = rootDirectory;
     this.fileExtension = fileExtension;
     this.signerParser = signerParser;
+    this.artifactSignerCache =
+        CacheBuilder.newBuilder()
+            .maximumSize(maxSize)
```
usmansaleem, line 62: what would be the behavior of -ve size? same as 0?

jframe: negative values aren't allowed, Guava cache checks for this and will throw an IllegalArgumentException. So Eth2Signer will fail to startup in this case.

usmansaleem: 👍

Issue #45, opened 2020-03-23
Support Env Variables for CLI options


arash009:
Enable the use of environment variables for all the CLI options used

Issue #46, opened 2020-03-23
Support a config file for CLI options


arash009:
Support the CLI options through a config file

To support YAML

CLI overrides config file

PR #58, opened 2020-04-01
[#45] - Support cli options from env variables and yaml config file

usmansaleem:
Fix issues #45 and #46.

Read default values from config file specified using `--config-file` or environment variables prefixed with `ETH2SIGNER_`. The precedence order is cli, env, config file


commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
```diff
@@ -0,0 +1,30 @@
...
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Unmatched;
+
+@Command(mixinStandardHelpOptions = true)
+class ConfigFileCommand {
+  @Option(names = CONFIG_FILE_OPTION_NAME, description = "...")
+  File configPath = null;
+
+  @Unmatched List<String> unmatched;
```
jframe, line 29: why is this needed? doesn't look it is ever used?

commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.File;
+import java.util.List;
+
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Unmatched;
+
+@Command(mixinStandardHelpOptions = true)
+class ConfigFileCommand {
+  @Option(names = CONFIG_FILE_OPTION_NAME, description = "...")
```
jframe, line 26: can a description be added

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/CascadingDefaultProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+import picocli.CommandLine.IDefaultValueProvider;
+import picocli.CommandLine.Model.ArgSpec;
+
+public class CascadingDefaultProvider implements IDefaultValueProvider {
+  private final IDefaultValueProvider defaultValueProvider1;
+  private final IDefaultValueProvider defaultValueProvider2;
+
+  public CascadingDefaultProvider(
+      final IDefaultValueProvider defaultValueProvider1,
```
jframe, line 23: Why limit this to just two defaultProviders? The version in Besu can handle multiple and the code is simpler

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
```diff
@@ -0,0 +1,136 @@
...
+  @Override
+  public String defaultValue(final ArgSpec argSpec) {
+    loadConfigurationFromFile();
+
+    // only options can be used in config because a name is needed for the key
+    // so we skip default for positional params
+    return argSpec.isOption() ? getConfigurationValue(((OptionSpec) argSpec)) : null;
+  }
+
+  private void loadConfigurationFromFile() {
```
jframe, line 57: seems a bit odd that the function doesn't return a value since it is loading something. Thinking it should return the Map<String, Object>. I don't think the stateful field variable result is needed at all

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
```diff
@@ -0,0 +1,136 @@
...
+
+    final Set<String> unknownOptionsList =
+        result.keySet().stream()
+            .filter(option -> !commandSpec.optionsMap().containsKey("--" + option))
+            .collect(Collectors.toSet());
+
+    if (!unknownOptionsList.isEmpty()) {
+      final String options = unknownOptionsList.size() > 1 ? "options" : "option";
+      final String csvUnknownOptions =
+          unknownOptionsList.stream().collect(Collectors.joining(", "));
```
jframe, line 101: can be replaced with String.join(", ", unknownOptionsList);

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
```diff
@@ -12,41 +12,79 @@
...
+      final Map<String, String> environment) {
     this.baseCommand = baseCommand;
     this.outputWriter = outputWriter;
     this.errorWriter = errorWriter;
+    this.environment = environment;
   }
 
   public int parseCommandLine(final String... args) {
+    // first pass to obtain config file if specified
+    ConfigFileCommand configFileCommand = new ConfigFileCommand();
```
jframe, line 50: nit: final

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
```diff
@@ -12,41 +12,79 @@
...
+    final CommandLine configFileCommandLine = new CommandLine(configFileCommand);
+    configFileCommandLine.parseArgs(args);
+    if (configFileCommandLine.isUsageHelpRequested()) {
+      new CommandLine(baseCommand).usage(outputWriter);
+      return 0;
```
jframe, line 55: Nicer to use commandLine.getCommandSpec().exitCodeOnUsageHelp

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
```diff
@@ -12,41 +12,79 @@
...
+    } else if (configFileCommandLine.isVersionHelpRequested()) {
+      new CommandLine(baseCommand).printVersionHelp(outputWriter);
+      return 0;
```
jframe, line 58: Nicer to use commandLine.getCommandSpec().exitCodeOnVersionHelp

commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.File;
+import java.util.List;
+
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Unmatched;
+
+@Command(mixinStandardHelpOptions = true)
+class ConfigFileCommand {
```
jframe, line 25: Found this a bit confusing at first as this isn't the real config file command just there for the initial parse. Is there a better name can be used for this or perhaps just make this an internal class for the CommandLineParser?

commandline/src/main/java/tech/pegasys/eth2signer/commandline/Eth2SignerCommand.java
```diff
@@ -58,6 +61,13 @@
 
   private static final Logger LOG = LogManager.getLogger();
 
+  @SuppressWarnings("UnusedVariable")
+  @CommandLine.Option(
```
jframe, line 65: Could this be done as a mixin perhaps, so we could use the same command for both the initial and base command?

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
```diff
@@ -0,0 +1,136 @@
...
+    if (result != null) {
+      return;
+    }
+
+    try {
+      final String configYaml = Files.readString(configFile.toPath());
+      final Map<String, Object> result = new Yaml().load(configYaml);
+      checkConfigurationValidity(result == null || result.isEmpty());
+      checkUnknownOptions(result);
+      this.result = result;
```
jframe, line 67: Would be nicer to just return the result instead of assigning to a field variable

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
```diff
@@ -0,0 +1,136 @@
...
+  public String defaultValue(final ArgSpec argSpec) {
+    loadConfigurationFromFile();
+
+    // only options can be used in config because a name is needed for the key
+    // so we skip default for positional params
+    return argSpec.isOption() ? getConfigurationValue(((OptionSpec) argSpec)) : null;
+  }
+
+  private void loadConfigurationFromFile() {
+    if (result != null) {
```
jframe, line 58: Why is this needed?

commandline/src/test/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProviderTest.java
```diff
@@ -0,0 +1,119 @@
...
+    final String extraYamlOptions = "";
+    final File configFile =
+        Files.writeString(tempDir.resolve("config.yaml"), extraYamlOptions).toFile();
+    final Eth2SignerCommand eth2SignerCommand = new Eth2SignerCommand();
+    final CommandLine commandLine = new CommandLine(eth2SignerCommand);
+    commandLine.registerConverter(Level.class, Level::valueOf);
+    commandLine.setDefaultValueProvider(new YamlConfigFileDefaultProvider(commandLine, configFile));
+
+    final String[] args = CmdlineHelpers.validBaseCommandOptions().split(" ");
+    assertThatExceptionOfType(CommandLine.ParameterException.class)
```
jframe, line 101: Do we care if the config file is empty? Since we have default options for everything having an empty config could be considered valid?

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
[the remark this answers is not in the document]
usmansaleem: 👍

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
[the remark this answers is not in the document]
usmansaleem: done, extracted out separate method as well.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
```diff
@@ -12,41 +12,79 @@
...
     return commandLine.execute(args);
   }
 
+  private IDefaultValueProvider defaultValueProvider(
+      final CommandLine commandLine, final Optional<File> configFile) {
+    final IDefaultValueProvider defaultValueProvider;
+    if (configFile.isPresent()) {
+      defaultValueProvider =
+          new CascadingDefaultProvider(
+              new EnvironmentVariableDefaultProvider(environment),
```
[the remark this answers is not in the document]
usmansaleem: If config file is specified, both env variables provider and config file provider will be used (with env have higher precedence than config file). Else only env variables provider will be used.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
usmansaleem: This is the "two pass" approach. We first create `commandLine(ConfigFileCommand)` which allows us to read optional "config file" option, rest of the command options will be treated as unmatched. This will avoid invoking any option validations by picocli which would be done in the second pass, which is against `Eth2SignerCommand`.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
usmansaleem: We don't print usage from "ConfigFileCommand". It is purely used as a way to parse config file option in the "first pass". The actual description comes in Eth2SignerCommand's configFile option which gets printed when user ask from usage help.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
usmansaleem: It was copied from Besu. This is only required to be executed once during parsing (not initial object creation). I'll modify it slightly to avoid modify instance variable.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
usmansaleem: It was originally internal command ... I think I should move it back.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
usmansaleem: The section for "extra invalid options" require the configuration file to be initialized during parse method as the command line options are fully parsed and initialised. During the initial object creation, we don't have commandLine.parse yet.

usmansaleem: Or may be we can perform "invalid options" check once in parse method and load configuration file when we create object?

commandline/src/test/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProviderTest.java
usmansaleem: The check was there in Besu's "toml" configuration. This implementation is roughly based on that.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
usmansaleem: we need the "result" in consecutive calls of "defaultValue".

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/YamlConfigFileDefaultProvider.java
usmansaleem: 👍 Updated test cases as well.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/valueprovider/CascadingDefaultProvider.java
usmansaleem: refactored.

commandline/src/main/java/tech/pegasys/eth2signer/commandline/ConfigFileCommand.java
usmansaleem: Based on https://github.com/remkop/picocli/blob/master/picocli-examples/src/main/java/picocli/examples/defaultprovider/TwoPass.java

commandline/src/main/java/tech/pegasys/eth2signer/commandline/CommandlineParser.java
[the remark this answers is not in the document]
usmansaleem: slightly refactored to make it more readable.

PR #65, opened 2020-05-06
Hashicorp key retrieval metric

jframe:
Add hashicorp to the key retrieval metrics. This unfortunately got missed when Hashicorp support for key retrieval was added.

### DIFF
--- a/core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
+++ b/core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
@@ -74,7 +74,28 @@ public class ArtifactSignerFactory {
     }
   }
 
-  public ArtifactSigner create(final HashicorpSigningMetadata metadata) {
+  public ArtifactSigner create(final HashicorpSigningMetadata hashicorpMetadata) {
+    try (TimingContext ignored = privateKeyRetrievalTimer.labels("hashicorp").startTimer()) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
```diff
@@ -74,7 +74,28 @@ public ArtifactSigner create(final FileKeyStoreMetadata fileKeyStoreMetadata) {
...
   }
 
-  public ArtifactSigner create(final HashicorpSigningMetadata metadata) {
+  public ArtifactSigner create(final HashicorpSigningMetadata hashicorpMetadata) {
+    try (TimingContext ignored = privateKeyRetrievalTimer.labels("hashicorp").startTimer()) {
+      return createHashicorpArtifact(hashicorpMetadata);
+    }
+  }
+
+  private ArtifactSigner createKeystoreArtifact(final FileKeyStoreMetadata fileKeyStoreMetadata) {
```
usmansaleem, line 83: nit: should this method be remain below createHashicorpArtifact?

jframe: I think I prefer all the private functions together after the public functions. So that's what I've done here.

usmansaleem: 👍

PR #72, opened 2020-05-26
publicKeys endpoint

jframe:
Add public key endpoint that will list all the currently identifier public keys

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/PublicKeysAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/PublicKeysAcceptanceTest.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2020 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.eth2signer.tests;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import tech.pegasys.artemis.bls.BLSKeyPair;
+import tech.pegasys.artemis.bls.BLSPublicKey;
+import tech.pegasys.artemis.bls.BLSSecretKey;
+import tech.pegasys.eth2signer.dsl.HttpResponse;
+import tech.pegasys.eth2signer.dsl.signer.SignerConfigurationBuilder;
+import tech.pegasys.eth2signer.dsl.utils.MetadataFileHelpers;
+import tech.pegasys.signers.bls.keystore.model.KdfFunction;
+
+import java.nio.file.Path;
+import java.util.List;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import org.apache.tuweni.bytes.Bytes;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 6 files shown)


acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/PublicKeysAcceptanceTest.java
```diff
@@ -0,0 +1,123 @@
...
+    startSigner(builder.build());
+
+    final HttpResponse response = signer.getRawRequest("/signer/publicKeys");
+    assertThat(response.getStatusCode()).isEqualTo(HttpResponseStatus.OK.code());
+    assertThat(publicKeysFromResponse(response)).isEmpty();
+  }
+
+  @Test
+  public void onlyValidKeysAreReturnedInPublicKeyResponse() throws Exception {
+    final BLSKeyPair keyPair1 =
```
rain-on, line 71: are these two lines maybe a helper function?

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -99,27 +112,22 @@ public void run() {
...
+    return new DirectoryBackedArtifactSignerProvider(
+        config.getKeyConfigPath(),
+        "yaml",
+        new YamlSignerParser(artifactSignerFactory),
+        config.getKeyCacheLimit());
   }
 
   private Handler<HttpServerRequest> createRouter(
-      final Vertx vertx, final SigningRequestHandler signingHandler) {
+      final Vertx vertx,
```
rain-on, line 128: suspect this is now not such a great function, seeing as every new route requires new parameters to this function - should it be refactoredd so that the router gets passed into each handler, which the registers itself? (or similar?)

core/src/main/java/tech/pegasys/eth2signer/core/http/PublicKeyRequestHandler.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.logging.log4j.Logger;
+
+public class PublicKeyRequestHandler implements Handler<RoutingContext> {
+
+  private static final Logger LOG = LogManager.getLogger();
+  private final ArtifactSignerProvider signerProvider;
+  private final ObjectMapper objectMapper;
+
+  public PublicKeyRequestHandler(
+      final ArtifactSignerProvider signerProvider, final ObjectMapper objectMapper) {
```
rain-on, line 33: does this need a specific objectmapper, or is any object mapper ok?

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -85,9 +86,21 @@ public void run() {
     try {
       metricsEndpoint.start(vertx);
 
-      final SigningRequestHandler signingHandler = createSigningHandler(metricsSystem, vertx);
+      final DirectoryBackedArtifactSignerProvider signerProvider =
+          createSignerProvider(metricsSystem, vertx);
```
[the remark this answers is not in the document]
jframe: It's the cache isn't been loaded and don't want to create the server until that is done so that the application blocks until the cache is ready.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/PublicKeysAcceptanceTest.java
```diff
@@ -0,0 +1,123 @@
...
+        new BLSKeyPair(BLSSecretKey.fromBytes(Bytes.fromHexString(PRIVATE_KEY_1)));
+    final Path keyConfigFile1 = configFileName(keyPair1.getPublicKey());
+    metadataFileHelpers.createUnencryptedYamlFileAt(keyConfigFile1, PRIVATE_KEY_1);
```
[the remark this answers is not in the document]
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
jframe: done

core/src/main/java/tech/pegasys/eth2signer/core/http/PublicKeyRequestHandler.java
jframe: Doesn't need anything specific from the object mapper

PR #91, opened 2020-06-08
Serve static Swagger ui

usmansaleem:
Serve static swagger ui which can be accessed via /openapi/


core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -164,6 +166,11 @@ private void registerPublicKeysRoute(
         .failureHandler(errorHandler);
   }
 
+  private void registerStaticRoute(final Router router) {
+    router.route("/*").handler(StaticHandler.create());
```
jframe, line 170: Do we want to be a bit more selective here, just everything under swagger-ui/*?

core/src/main/resources/webroot/swagger-ui/absolute-path.js
```diff
@@ -0,0 +1,14 @@
+/*
```
jframe, line 1: Instead of copying swagger-ui is it possible to use a webjar (https://www.webjars.org/) instead so we have easy versioning and don't need to copy so many resource files into the project?

usmansaleem: Hmm ... I can give it a go ... it would mean we would have to enable VertX CP loading.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
usmansaleem: done, `/openapi/*` will be loaded from `openapi` local folder

core/src/main/resources/webroot/swagger-ui/absolute-path.js
usmansaleem: Instead of using webjars (which doesn't allow to load our custom openapi yaml), use unpkg interface as suggested by official swagger-ui installation.

core/src/main/resources/openapi/eth2signer.yaml
```diff
@@ -0,0 +1,92 @@
...
+              properties:
+                signingRoot:
+                  type: string
+              example:
+                signingRoot: '0x48656c6c6f2c20776f726c6421'
+      responses:
+        '200':
+          description: 'hex encoded string of signature'
+          content:
+            application/json:
```
jframe, line 43: response type is actually text/plain as the value is returned as is without any extra formatting

core/src/main/resources/openapi/eth2signer.yaml
```diff
@@ -0,0 +1,92 @@
+openapi: 3.0.0
+info:
+  title: 'Eth2Signer'
+  description: 'Sign Eth2 Artifacts'
+  version: '1.0.0'
```
jframe, line 5: Think it would be simpler if this matched the application version. i.e. 0.0.1-SNAPSHOT. Looks like another piece we have to keep in sync though.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -164,6 +166,11 @@ private void registerPublicKeysRoute(
         .failureHandler(errorHandler);
   }
 
+  private void registerStaticRoute(final Router router) {
+    router.route("/openapi/*").handler(StaticHandler.create("openapi"));
+    LOG.info("Registered static route /openapi");
```
jframe, line 171: nit: not sure is much value in this log statement

core/src/main/resources/openapi/eth2signer.yaml
usmansaleem: hmm ... yes, otherwise we can do some token replacement but this sits in source code.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
usmansaleem: agreed!

core/src/main/resources/openapi/eth2signer.yaml
usmansaleem: Implemented token replacement to keep the project version sync with yaml version.

core/src/main/resources/openapi/eth2signer.yaml
```diff
@@ -0,0 +1,92 @@
...
+              example:
+                signingRoot: '0x48656c6c6f2c20776f726c6421'
+      responses:
+        '200':
+          description: 'hex encoded string of signature'
+          content:
+            text/plain; charset=utf-8:
+              schema:
+                type: string
+                example: '0xb3baa751d0a9132cfe93e4e3d5ff9075111100e3789dca219ade5a24d27e19d16b3353149da1833e9b691bb38634e8dc04469be7032132906c927d7e1a49b414730612877bc6b2810c8f202daf793d1ab0d6b5cb21d52f9e52e883859887a5d9'
```
jframe, line 46: When I try the example, I get back a 406 "Error: Not Acceptable"

core/src/main/resources/openapi/eth2signer.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+  /signer/publicKeys:
+    get:
+      tags:
+        - 'Public Key'
+      summary: 'List of available Public Keys'
+      description: 'Returns the public keys for the private keys that have been loaded into Eth2Signer'
+      operationId: 'getPublicKeys'
+      responses:
+        '200':
```
jframe, line 62: When use the try it now, I'm getting message in the response body area "Unrecognized response type; displaying content as text."

core/src/main/resources/openapi/eth2signer.yaml
usmansaleem: Do you have valid primary keys loaded?

core/src/main/resources/openapi/eth2signer.yaml
usmansaleem: Yes, I noticed that as well, I will debug it with curl to figure out whats vertx sending (we do set json type in server).

core/src/main/resources/openapi/eth2signer.yaml
jframe: I did, it's working now on latest code

PR #93, opened 2020-06-09
use redocly instead of swagger-ui to display eth2signer openapi spec

usmansaleem:


usmansaleem:
Swagger UI is preferred for local serving.

Issue #94, opened 2020-06-09
Publish Eth2Signer OpenApi spec to gh_pages branch


usmansaleem:
Publish Eth2Signer OpenApi spec (via CI) to `gh_pages` branch so that it can be accessed via https://pegasyseng.github.io/eth2signer

PR #95, opened 2020-06-10
Publish openapi to gh-pages branch

usmansaleem:
Publish Eth2Signer OpenAPI spec to gh-pages branch using CI.

## Issue
fixes Issue #94


.circleci/config.yml
```diff
@@ -189,7 +193,23 @@ jobs:
...
+      - add_ssh_keys:
+          fingerprints:
+            - "8c:7f:3d:66:05:d9:12:a2:af:58:59:bf:97:c1:c7:84"
+      - attach_workspace:
+          at: ~/project
+      - run:
+          name: Build Folder
+          working_directory: .openapidoc 
+          command: |
+            npm install
```
NicolasMassart, line 210: favour the use of `npm ci` instead of `npm install` and use cache for npm. see https://docs.google.com/document/d/1LUJ5UpJq94fCS7Hnx4Mm3efWzIVymyx3cI-2wrYMwM0

benjamincburns: +1

.openapidoc/index.js
```diff
@@ -0,0 +1,167 @@
...
+  "https://github.com/PegaSysEng/eth2signer/raw/gh-pages/versions.json";
+const distDir = "./dist";
+const versionsFile = distDir + "/versions.json";
+
+/**
+ * Downloads the given source URL and return it as a string
+ * @param source
+ */
+async function downloadAsString(source) {
+  return new Promise((resolve, reject) => {
```
benjamincburns, line 20: I'd just use `fetch` as it returns a promise, so you can `await` it.

.openapidoc/index.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+async function downloadAndUpdateVersionsFile(specVersion) {
+  return new Promise((resolve, reject) => {
+    if (!isReleaseVersion(specVersion)) {
+      console.log("Not a release version, bypassing updating versions.json");
+      return resolve(false);
+    }
+
+    return downloadAsString(versionsFileUrl)
+      .then((value) => {
```
benjamincburns, line 107: use `await`, not `.then`

.openapidoc/index.js
```diff
@@ -0,0 +1,167 @@
...
+        try {
+          updateVersionJson(specVersion, value);
+          return resolve(true);
+        } catch (err) {
+          return reject(err);
+        }
+      })
+      .catch((err) => {
```
benjamincburns, line 115: use a try/catch block, not `.catch`

.openapidoc/index.js
```diff
@@ -0,0 +1,167 @@
...
+  });
+}
+
+/**
+ * Publish dist folder to gh-pages branch
+ * @param specVersion
+ */
+async function publishToGHPages(specVersion) {
+  return new Promise((resolve, reject) => {
+    ghpages.publish(
```
benjamincburns, line 127: does this return a Promise? if so, just use the `await` keyword

.openapidoc/index.js
```diff
@@ -0,0 +1,167 @@
...
+  console.log("Spec Version: " + specVersion);
+
+  copySpecToDist(specVersion);
+  await downloadAndUpdateVersionsFile(specVersion);
+  await publishToGHPages(specVersion);
+
+  console.log("OpenAPI specs published to gh-pages successfully.");
+}
+
+run().catch((error) => {
```
benjamincburns, line 163: put the catch code inside of a try/catch block in the `run` function

.openapidoc/package.json
```diff
@@ -0,0 +1,17 @@
+{
```
benjamincburns, line 1: I always use `husky` and `lint-staged` to fire off pre-commit lint checks. I'd recommend it even more strongly in this case as this is code that isn't going to be touched very often, so the temptation to write bad/unmaintainable code will be higher than code that people are reading all the time. 

See this as an example for how I set it up for `circle-build-monitor`. That's a TypeScript project, but all the tools I'm using stay the same for JS, just the eslint config would need to change a little... https://github.com/PegaSysEng/circle-build-monitor/blob/master/package.json#L40-L50

.openapidoc/index.js
```diff
@@ -0,0 +1,167 @@
...
+      .get(source, (response) => {
+        const { statusCode, headers } = response;
+        if (statusCode >= 400)
+          return reject(
+            new Error(
+              `Failed to download file "${source}". Status: ${statusCode}`
+            )
+          );
+
+        response.on("data", (chunk) => (output += chunk));
```
benjamincburns, line 32: Yeah, definitely just use `fetch` - it'll invalidate the need for this entire function.

.openapidoc/index.js
benjamincburns: Here's an example of how I use `fetch` in the `circle-build-monitor` - it's TypeScript, but TS is just a superset of JavaScript, so it should be easy enough to port over... https://github.com/PegaSysEng/circle-build-monitor/blob/master/src/polling.ts#L122

.openapidoc/index.js
benjamincburns: As an example, I could've written this code with `.then` instead of `await` and `.catch` instead of try/catch, but it's a lot more readable this way:  https://github.com/PegaSysEng/circle-build-monitor/blob/master/src/polling.ts#L122

.openapidoc/index.js
usmansaleem: Unfortunately not, hence wrapped it around Promise.

.openapidoc/index.js
usmansaleem: I've switched to `node-fetch` npm module instead, this method is removed.

.openapidoc/index.js
usmansaleem: method is removed.

.openapidoc/index.js
usmansaleem: moved to try/catch block

.openapidoc/index.js
usmansaleem: 👍

.openapidoc/index.js
usmansaleem: the last run() (which I've renamed to main()) cannot be part of try/catch because it is not part of an async method.

.openapidoc/index.js
```diff
@@ -0,0 +1,142 @@
...
+function calculateVersionFromSpec() {
+  const data = yaml.safeLoad(fs.readFileSync(config.specPath, "utf8"));
+  config.specVersion = data.info.version;
+}
+
+/**
+ * Determine if its a release version from spec version. It contains -dev- as a result from our gradle build
+ */
+function isReleaseVersion() {
+  return !config.specVersion.includes("-dev-");
```
NicolasMassart, line 49: Don't you base your releases on tags? A doc for a released version should not change once published. Here it means that you will have the doc tied to updates on the branch and not only on the final tagging action. It may be better, I don't know, but it must be clearly expected.

.circleci/config.yml
```diff
@@ -201,6 +229,15 @@ workflows:
       - buildDocker:
           requires:
             - build
+      - publishOpenApiSpec:
+          filters:
+            branches:
+              only:
+                - master
+                - /^release-.*/
```
NicolasMassart, line 237: see my comment on the `jsReleaseVersion` function.

.openapidoc/index.js
```diff
@@ -0,0 +1,142 @@
...
+      config.specPath,
+      `${config.distDir}/eth2signer-${config.specVersion}.yaml`
+    );
+  }
+}
+
+// start execution of main method
+main()
+  .then(() => {
+    process.stdout.write(
```
benjamincburns, line 133: Might as well put this in the `main`, no? Also use `console.log` instead of `process.stdout`... Also `console.error` for writing to stderr.

.openapidoc/index.js
```diff
@@ -0,0 +1,142 @@
...
+      `OpenAPI specs [${config.specVersion}] published to [${config.branch}] using user [${ghPagesConfig().user.name}].\n`
+    );
+  })
+  .catch((err) => {
```
benjamincburns, line 137: might as well do this in a `catch` block inside of the `main` function

PR #99, opened 2020-06-22
Metrics allow hostlist

### LINKED ISSUE #97: 403 forbidden on metrics
hello
when i try  to connect to metrics from another pod, i get 403 forbidden back.
i have started eth2signer with:
```
--http-listen-port=9000                                                                                                                                                                
--http-listen-host=0.0.0.0                                                                                                                                                                  --key-store-path=/data                                                                                                                                                                      --logging=ALL                                                                                                                                                                          
--metrics-enabled                                                                                                                                                                     
--metrics-port=9001                                                                                                                                                                          --metrics-host=0.0.0.0
```
and i get metrics if i get a shell in the container, install curl and use it. 
unfortunately, when i try to connect from another pod/container i get 403. prometheus also reports 403

cheers

jframe:
Adds a new command line option --metrics-host-allowlist to allow configuration of the metrics allow host list. The metrics endpoint always supported the allow host list functionality but it wasn't possible to configure this. The metrics allow host list functionality restricts access to the list hostnames.

fixes #97

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/Signer.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/Signer.java
@@ -68,6 +68,10 @@ public class Signer {
   }
 
   public String getUrl() {
-    return String.format(urlFormatting, hostname, runner.httpJsonRpcPort());
+    return String.format(urlFormatting, hostname, runner.httpPort());
+  }
+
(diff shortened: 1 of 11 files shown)


jframe:
> Looks good to me. Is there a need to add any unit test for AllowListHostProperty class?

Think we should, there is some special logic around combining of * and none with other values. I've added unit test for that class.

Issue #102, opened 2020-06-23
Restrict rest API access to specified hostnames


jframe:
Add a host allow list that will restrict access to the rest API to the specified hostnames. This should follow the same approach as the --metrics-host-allowlist but for the rest API and default to allow access to localhost and 127.0.0.1

bgravenorst:
See:
- https://docs.eth2signer.pegasys.tech/en/latest/Reference/CLI/CLI-Syntax/#http-host-allowlist
- https://docs.eth2signer.pegasys.tech/en/latest/Reference/CLI/CLI-Syntax/#metrics-host-allowlist

PR #104, opened 2020-06-24
Http allow hostlist

jframe:
Adds a host allow list that will restrict access to the rest API to the specified hostnames.

* new command line --http-host-allowlist which defaults to "localhost, 127.0.0.1"
* by default http requests will be restricted to only localhost and 127.0.0.1 (checking is based on host header)

#102


core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -140,6 +142,10 @@ private DirectoryBackedArtifactSignerProvider createSignerProvider(
         config.getKeyCacheLimit());
   }
 
+  private void registerCheckAllowListHostHeader(final Router router) {
+    router.route().handler(new HostAllowListHandler(config.getHttpHostAllowList()));
+  }
+
```
usmansaleem, line 148: There might be a merge conflict here, this needs to be placed right besides `registerOpenApiSpecRoute` I think

jframe: There is, i've resolve it now. Also had to install this as a global handler on the openapi router factory for it work.

Issue #80, opened 2020-05-29
TLS support


jframe:
Add TLS support to Eth2Signer. This will likely follow the approach taken by EthSigner and so we can copy the implementation from there.

bgravenorst:
Documented here:
https://docs.eth2signer.pegasys.tech/en/latest/HowTo/Configure-TLS/

PR #110, opened 2020-06-29
TLS

jframe:
Adds support for TLS for connecting to Eth2Signer. This also includes client authentication.

#80 

The following CLI options are added for TLS support
* --tls-keystore-file Path to a PKCS#12 formatted keystore; used to enable TLS on inbound connections.
* --tls-keystore-password-file Path to a file containing the password used to decrypt the keystore.
* --tls-allow-any-client If defined, any client may connect, regardless of presented certificate. This cannot be set if either a whitelist or CA clients have been enabled.
* --tls-known-clients-file Path to a file containing the fingerprints of authorized clients.
* --tls-allow-ca-clients If defined, allows clients authorized by the CA to connect to EthSigner.


acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerProcessRunner.java
```diff
@@ -107,6 +82,17 @@ protected void startExecutor(final List<String> params) {
...
   }
 
+  private void addTrustStoreOptions(final StringJoiner javaOpts) {
+    if (getSignerConfig().getOverriddenCaTrustStore().isPresent()) {
+      final TlsCertificateDefinition caTrustStore =
+          getSignerConfig().getOverriddenCaTrustStore().get();
+      final Path overriddenCaTrustStorePath = createJksCertFile(caTrustStore);
+      javaOpts.add(
+          "-Djavax.net.ssl.trustStore=" + overriddenCaTrustStorePath.toAbsolutePath().toString());
+      javaOpts.add("-Djavax.net.ssl.trustStorePassword=" + caTrustStore.getPassword());
```
rain-on, line 92: can we do something nicer - like return the javaopts, arther than modifying the passed in one?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
```diff
@@ -88,17 +84,7 @@ public void start() {
...
-    } finally {
-      if (signerConfig.isHttpDynamicPortAllocation()) {
-        try {
-          MoreFiles.deleteRecursively(dataPath, RecursiveDeleteOption.ALLOW_INSECURE);
-        } catch (final IOException e) {
-          LOG.info("Failed to clean up temporary file: {}", dataPath, e);
-        }
-      }
-    }
+    shutdownExecutor();
```
rain-on, line 87: is it worth just renaming shutdownExecutor to "shutdown" and making it public?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
```diff
@@ -219,4 +205,23 @@ public int metricsPort() {
       return signerConfig.getMetricsPort();
     }
   }
+
+  private Path createTempDirectory(final String prefix) {
+    try {
+      final Path certificateDirectory = Files.createTempDirectory(prefix);
```
rain-on, line 211: nit: not necessarily the certdirectory - could be the datadir ... maybe rename to just "tempDir"?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
```diff
@@ -219,4 +205,23 @@ public int metricsPort() {
...
+      throw new RuntimeException("Unable to create temporary directory", e);
+    }
+  }
+
+  public Path createJksCertFile(final TlsCertificateDefinition caTrustStore) {
+    final Path certificateDirectory = createTempDirectory("acceptance-test-jks-cert");
+    return createJksTrustStore(certificateDirectory, caTrustStore);
+  }
+
+  public SignerConfiguration getSignerConfig() {
```
rain-on, line 224: does this need to be public, or just protected?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
jframe: this can be protected. done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerProcessRunner.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
jframe: Prefer that the shutdownExecutor wasn't made public as the startup and shutdown are the public functions. For startup there is extra work to be done, but shutdown there isn't. So could get rid of the shutdownExecutor and rename that to shutdown but think it would be odd  to not have a shutdownExecutor. Maybe it doesn't matter though?

PR #114, opened 2020-07-03
Fix TLS allow hostlist not working at runtime due to missing bouncycastle lib

jframe:
Fix TLS allow hostlist not working at runtime

* Added missing bouncycastle library as runtime dependency
* Changed gradle build so that the process runner is used when run as part of the build
* Fixed java opts not being set correctly in the process runner

### DIFF
--- a/acceptance-tests/build.gradle
+++ b/acceptance-tests/build.gradle
@@ -46,7 +46,6 @@ task acceptanceTest(type: Test) {
   description = 'Runs Eth2Signer acceptance tests.'
   group = 'verification'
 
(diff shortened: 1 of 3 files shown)


acceptance-tests/build.gradle
```diff
@@ -45,6 +46,9 @@ task acceptanceTest(type: Test) {
   description = 'Runs Eth2Signer acceptance tests.'
   group = 'verification'
 
+  setSystemProperties(test.getSystemProperties())
```
rain-on, line 49: nit: what's this required for? Does it effectively copy a test's properties to the child process?

Issue #116, opened 2020-07-07
 eth2signer HTTP server doesn't correctly set the Content-Type if there is no "Accept" header


usmansaleem:
If Accept header is empty in the client accessing eth2signer, for instance:

`curl -v -H 'Accept: ' http://127.0.0.1:9000/signer/publicKeys` 

The output doesn't contain Content-Type header.
~~~
*   Trying 127.0.0.1:9000...
* Connected to 127.0.0.1 (127.0.0.1) port 9000 (#0)
> GET /signer/publicKeys HTTP/1.1
> Host: 127.0.0.1:9000
> User-Agent: curl/7.70.0
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 405
< 
* Connection #0 to host 127.0.0.1 left intact
["0x8e8c841eb95aebff6a39bc63739f5463397d42cefb602e36ccf1286ea379357c29468fa114751e4851231d858705228b","0xa1ed1f6e807a89b4e72809982f131010a464b72bf77a714ddff6a6feab33d5cfa7ea3cb35c552d6409faade8887d385b","0x95deaa8ef09f158ab99d4cc213c5a05b9ca04c8326b80e5fe8f14045bd29e1985996a2d417979e59ae29bdff28245b79","0xa802afef4f6dca154fca83401ead04d912ca14b85de5d3a527bcb5b4ac5e2d90baa332fefd70e13eface2e36cb499e8b"]%
~~~

PR #117, opened 2020-07-07
[#116] - Manually set content type header

usmansaleem:
-- Disable automatic content type response handler and manually set content type from each handler

Fixes issue #116 

### DIFF
--- a/core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/GetPublicKeysHandler.java
+++ b/core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/GetPublicKeysHandler.java
@@ -12,6 +12,9 @@
  */
 package tech.pegasys.eth2signer.core.http.handlers;
 
+import static com.google.common.net.MediaType.JSON_UTF_8;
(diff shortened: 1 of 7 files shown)


core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/GetPublicKeysHandler.java
```diff
@@ -29,6 +32,6 @@ public GetPublicKeysHandler(final ArtifactSignerProvider signerProvider) {
   public void handle(final RoutingContext context) {
     final JsonArray jsonArray = new JsonArray();
     signerProvider.availableIdentifiers().forEach(jsonArray::add);
-    context.response().end(jsonArray.encode());
+    context.response().putHeader(CONTENT_TYPE, JSON_UTF_8.toString()).end(jsonArray.encode());
```
rain-on, line 35: can we not somehow default the route to always set the content-type?

rain-on: i.e. route.produces(TEXT)?

PR #130, opened 2020-07-21
Secp256k1 signing

jframe:
Adds secp256k1 signing to the signing endpoint. This uses the existing config format used in EthSigner using a toml file to describe the key configuration.

The endpoint /sign/signing can be used for both bls signing and secp256k1 signing. When used for secp signing the address must be used as the identifier.


core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/SignForIdentifierHandler.java
```diff
@@ -51,11 +51,11 @@ public SignForPublicKeyHandler(
...
     final RequestParameters params = routingContext.get("parsedParameters");
-    final String publicKey = params.pathParameter("publicKey").toString();
-    final Optional<ArtifactSigner> signer = signerProvider.getSigner(publicKey);
+    final String identifier = params.pathParameter("identifier").toString();
+    final Optional<ArtifactSigner> signer = signerProvider.getSigner(identifier);
     if (signer.isEmpty()) {
-      LOG.error("Unable to find an appropriate signer for request: {}", publicKey);
-      routingContext.fail(404);
+      LOG.trace("Unable to find an appropriate signer for request: {}", identifier);
+      routingContext.next();
```
usmansaleem, line 58: why doing next() instead of failing with 404?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariables;
+import org.junit.jupiter.api.io.TempDir;
+import org.web3j.crypto.Sign.SignatureData;
+
+public class SecpSigningAcceptanceTest extends AcceptanceTestBase {
+  private static final byte[] DATA_TO_SIGN = "42".getBytes(UTF_8);
+  private static final String PRIVATE_KEY =
+      "8f2a55949038a9610f50fb23b5883af3b4ecb3c3bb792cbcefbd1542c692be63";
+
+  private static final String SIGN_ENDPOINT = "/signer/sign/{publicKey}";
```
usmansaleem, line 54: The parameter should be change to `{identifier}` to keep in line with openapi spec changes.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -0,0 +1,178 @@
...
+    final SignerConfigurationBuilder builder = new SignerConfigurationBuilder();
+    builder.withKeyStoreDirectory(tomlDirectory);
+    startSigner(builder.build());
+
+    final Response response =
+        given()
+            .baseUri(signer.getUrl())
+            .filter(getOpenApiValidationFilter())
+            .contentType(ContentType.JSON)
+            .pathParam("publicKey", FILENAME)
```
usmansaleem, line 84: continuing to previous comment. `pathParam("identifier")`

core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/SignForIdentifierHandler.java
jframe: So that we can fall through the other instance of this handler. We have one registered with a ArtifactSignerProvider that handles BLS keys and one that has a secp ArtifactSignerProvider. This way if the key isn't a BLS key we will try find it as a secp key. And if that fails vertx will just return a 404 anyway.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -148,15 +164,18 @@ private OpenAPI3RouterFactory createOpenApiRouterFactory(
...
         GET_PUBLIC_KEYS_OPERATION_ID, errorHandler);
 
     openAPI3RouterFactory.addHandlerByOperationId(
-        SIGN_FOR_PUBLIC_KEY_OPERATION_ID,
-        new SignForPublicKeyHandler<>(signerProvider, this::formatBlsSignature, BLS));
+        SIGN_FOR_IDENTIFIER_OPERATION_ID,
+        new SignForIdentifierHandler<>(blsSignerProvider, this::formatBlsSignature, BLS));
+    openAPI3RouterFactory.addHandlerByOperationId(
+        SIGN_FOR_IDENTIFIER_OPERATION_ID,
+        new SignForIdentifierHandler<>(secpSignerProvider, this::formatSecpSignature, SECP256K1));
```
usmansaleem, line 176: got it! the same identifier is used to register two handlers.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -0,0 +1,178 @@
...
+    final byte[] s = signature.slice(32, 32).toArray();
+    final byte[] v = signature.slice(64).toArray();
+    final BigInteger messagePublicKey = recoverPublicKey(new SignatureData(v, r, s));
+    assertThat(messagePublicKey).isEqualTo(expectedPublicKey);
+  }
+
+  private BigInteger recoverPublicKey(final SignatureData signature) {
+    try {
+      return signedMessageToKey(DATA_TO_SIGN, signature);
+    } catch (SignatureException e) {
```
usmansaleem, line 174: nit: final

jframe: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/BlsSigningAcceptanceTest.java
```diff
@@ -0,0 +1,143 @@
...
+  public void signDataWithKeyLoadedFromUnencryptedFile() {
+    final String configFilename = publicKey.toString().substring(2);
+    final Path keyConfigFile = testDirectory.resolve(configFilename + ".yaml");
+    metadataFileHelpers.createUnencryptedYamlFileAt(keyConfigFile, PRIVATE_KEY);
+
+    final SignerConfigurationBuilder builder = new SignerConfigurationBuilder();
+    builder.withKeyStoreDirectory(testDirectory);
+    startSigner(builder.build());
+
+    given()
```
rain-on, line 67: Is it worth hiding this behind a DSL? and being able to write "final Bytes signature = RemoteSigner.sign(key, DATA)"
assertThat(signature).equals("aaaa")?

rain-on: just that MOST of this "given" is repeated

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  static final String FILENAME = "fe3b557e8fb62b89f4916b721be55ceb828dbd73";
+
+  @Test
+  @EnabledIfEnvironmentVariables({
+    @EnabledIfEnvironmentVariable(named = "AZURE_CLIENT_ID", matches = ".*"),
+    @EnabledIfEnvironmentVariable(named = "AZURE_CLIENT_SECRET", matches = ".*"),
+    @EnabledIfEnvironmentVariable(named = "AZURE_KEY_VAULT_NAME", matches = ".*")
+  })
+  @EnabledIfEnvironmentVariable(named = "AZURE_KEY_VAULT_NAME", matches = ".*")
```
rain-on, line 64: can probably make these more simple - i.e. we KNOW that the multitoml file loading stuff works (due to tests in Signer) - just need to prove we've wired the HTTP endpoint up to the signerprovider

acceptance-tests/src/test/resources/secp256k1/rich_benefactor_one.json
```diff
@@ -0,0 +1 @@
+{"address":"fe3b557e8fb62b89f4916b721be55ceb828dbd73","crypto":{"cipher":"aes-128-ctr","ciphertext":"c330870080626a6cf84c36eeb89f6376316637fbc16c86400eb6c394de5aed81","cipherparams":{"iv":"7f2526feb38ac355c644a84e78ec5692"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"b0da609bb541df918eb55bbeaf237445aa406a0c2d6e975663c4fcb221bc54e0"},"mac":"1ed0ef9e2092d0f5db8d16409ee2aaee9af3f521a5eafd910ce19dbc76e67ed7"},"id":"43f8e00a-c6ec-4a62-b7cb-f54cb299c3db","version":3}
```
rain-on, line 1: do we really need a "rich benefactor" for eth2signer? Or is this simply a validator?

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -165,6 +184,16 @@ private String formatBlsSignature(final BlsArtifactSignature signature) {
     return signature.getSignatureData().toString();
   }
 
+  private String formatSecpSignature(final SecpArtifactSignature signature) {
```
rain-on, line 187: static?

core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/SignForIdentifierHandler.java
```diff
@@ -51,11 +51,11 @@ public SignForPublicKeyHandler(
...
   public void handle(RoutingContext routingContext) {
     final RequestParameters params = routingContext.get("parsedParameters");
-    final String publicKey = params.pathParameter("publicKey").toString();
-    final Optional<ArtifactSigner> signer = signerProvider.getSigner(publicKey);
+    final String identifier = params.pathParameter("identifier").toString();
+    final Optional<ArtifactSigner> signer = signerProvider.getSigner(identifier);
     if (signer.isEmpty()) {
-      LOG.error("Unable to find an appropriate signer for request: {}", publicKey);
-      routingContext.fail(404);
+      LOG.trace("Unable to find an appropriate signer for request: {}", identifier);
```
rain-on, line 57: log message is probably more like "Unsuitable handler, invoking next"?

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
jframe: Not seeing the benefit making static since this a private method. Can of course, is there a good reason to make it static?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/SecpSigningAcceptanceTest.java
jframe: Could possibly test most of this in an integration from the http endpoint to the SignerProvider. But I like having this basic AT for these providers as it also shows that the Runner and our ArtifactSignerProvider adapters are working.

core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/SignForIdentifierHandler.java
jframe: It's not necessarily that it's inappropriate. The key might not exist at all.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/BlsSigningAcceptanceTest.java
jframe: This would make the signer ATs cleaner. Would prefer to do this in another PR though.

core/src/main/java/tech/pegasys/eth2signer/core/http/handlers/SignForIdentifierHandler.java
rain-on: but if you do a context.next, then this handler can't make a global statement about the appropriateness of the identifier or not - only that _this_ handler can't handle it

jframe: done

Issue #125, opened 2020-07-16
Support Azure keyvault


arash009:
Add in support for the Azure key vault to eth2signer to store and retrieve BLS keys to be used to sign Blocks/Attestation etc for use by Teku

PR #147, opened 2020-07-28
Support Azure Key Vault for BLS Multisigner

usmansaleem:
#125 

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/BlsSigningAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/BlsSigningAcceptanceTest.java
@@ -16,6 +16,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Collections.singletonMap;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import tech.pegasys.eth2signer.core.multikey.metadata.AzureSigningMetadata;
 import tech.pegasys.eth2signer.dsl.HashicorpSigningParams;
 import tech.pegasys.eth2signer.dsl.utils.MetadataFileHelpers;
 import tech.pegasys.signers.bls.keystore.model.KdfFunction;
@@ -31,6 +32,8 @@ import java.nio.file.Path;
 import io.restassured.response.Response;
 import org.apache.tuweni.bytes.Bytes;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariables;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
```diff
@@ -135,6 +142,30 @@ private ArtifactSigner createHashicorpArtifact(final HashicorpSigningMetadata me
...
+              metadata.getClientSecret(),
+              metadata.getTenantId(),
+              metadata.getVaultName());
+
+      final Optional<String> secret = azureVault.fetchSecret(metadata.getSecretName());
+      if (secret.isEmpty()) {
+        throw new SigningMetadataException(
+            "secret '" + metadata.getSecretName() + "' doesn't exist");
+      }
+      final BLSKeyPair keyPair =
```
rain-on, line 160: do we want to wrap this up in "caching" signer - such that the secret key doesn't stay in memory - but rather fetched everytime? 
Or is that implicit in our whole design? I.e. we ALWAYS go via this class to get a signer? (and thus reconstruct it everytime)

core/src/test/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParserTest.java
```diff
@@ -227,4 +228,38 @@ private FileRawSigningMetadata hasPrivateKey(final String privateKey) {
     final BLSSecretKey blsSecretKey = BLSSecretKey.fromBytes(Bytes.fromHexString(privateKey));
     return argThat((FileRawSigningMetadata m) -> m.getSecretKey().equals(blsSecretKey));
   }
+
+  @Test
+  void azureMetaDataInfoReturnsMetadata() throws IOException {
```
rain-on, line 233:
```suggestion
  void azureMetadataInfoReturnsMetadata() throws IOException {
```

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/BlsSigningAcceptanceTest.java
```diff
@@ -86,6 +89,29 @@ public void ableToSignUsingHashicorp() {
...
+    final String clientSecret = System.getenv("AZURE_CLIENT_SECRET");
+    final String tenantId = System.getenv("AZURE_TENANT_ID");
+    final String keyVaultName = System.getenv("AZURE_KEY_VAULT_NAME");
+    final String secretName = "TEST-KEY";
+
+    final String configFilename = keyPair.getPublicKey().toString().substring(2);
+    final Path keyConfigFile = testDirectory.resolve(configFilename + ".yaml");
+    metadataFileHelpers.createAzureYamlFileAt(
+        keyConfigFile,
+        new AzureSigningMetadata(clientId, clientSecret, tenantId, keyVaultName, secretName));
```
jframe, line 110: Seems unnecessary to create a AzureSigningMetadata, could just pass the params through since we just pull all values out into a map again to create the yaml file.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/ArtifactSignerFactory.java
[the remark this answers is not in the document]
usmansaleem: This is part of overall "caching". It will ultimately be called from 

tech/pegasys/eth2signer/core/Runner.java:133

~~~
final DirectoryBackedArtifactSignerProvider blsSignerProvider =
          createSignerProvider(metricsSystem, vertx);
      blsSignerProvider.cacheAllSigners();
~~~

At the moment, it is only called once at the startup.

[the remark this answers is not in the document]
jframe: It is expected that the factory creates the signer each time as the caching is done at the DirectoryBackedArtifactSignerProvider layer. If caching is enabled the DirectoryBackedArtifactSignerProvider will ensure that this only gets created once. It is possible to disable caching in which case this will retrieve the key each time.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/SigningMetadata.java
```diff
@@ -23,7 +23,8 @@
 @JsonSubTypes({
   @Type(value = FileRawSigningMetadata.class, name = "file-raw"),
   @Type(value = FileKeyStoreMetadata.class, name = "file-keystore"),
-  @Type(value = HashicorpSigningMetadata.class, name = "hashicorp")
+  @Type(value = HashicorpSigningMetadata.class, name = "hashicorp"),
```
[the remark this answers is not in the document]
usmansaleem: This looks cleaner though. We would have to maintain a registry of subclasses anyways as discussed in solution-6.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/signing/BlsSigningAcceptanceTest.java
usmansaleem: applied.

PR #148, opened 2020-07-29
Secp keys

jframe:
Public key endpoint also now returns secp public keys.

* Changed the publicKey endpoint so that we can now specify which kind of key to return using /signer/publicKeys/{keyType} where keyType = BLS | SECP256K1
* Added keyType field to the RPC method public_keys which similarly will return either bls or secp public keys


acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTestBase.java
```diff
@@ -28,52 +29,117 @@
...
+
+  private void createInvalidFile(final Path keyConfigFile) {
+    try {
+      Files.createFile(keyConfigFile);
+    } catch (final IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  private void createSecpKey(final String blsPrivateKey) {
```
rain-on, line 119: why blsPrivateKey, for a secp Function?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTestBase.java
```diff
@@ -28,52 +29,117 @@
...
   }
 
-  private Path configFileName(final BLSPublicKey publicKey) {
-    final String configFilename2 = publicKey.toString().substring(2);
-    return testDirectory.resolve(configFilename2 + ".yaml");
+  protected String[] createSecpKeys(boolean isValid, final String... privateKeys) {
+    return Stream.of(privateKeys)
+        .map(
+            privateKey -> {
+              final ECKeyPair ecKeyPair =
```
rain-on, line 97: is there not a cleaner way to get to the Public Key without going via ECKeyPair?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTestBase.java
```diff
@@ -28,52 +29,117 @@
...
+    try {
+      Files.createFile(keyConfigFile);
+    } catch (final IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  private void createSecpKey(final String blsPrivateKey) {
+    final Bytes privateKey = Bytes.fromHexString(blsPrivateKey);
+    final ECKeyPair ecKeyPair = ECKeyPair.create(Numeric.toBigInt(privateKey.toArray()));
```
rain-on, line 121: this was already a bit done in the previous function - do we have to do it again?

core/src/main/java/tech/pegasys/eth2signer/core/service/operations/PublicKeys.java
```diff
@@ -12,19 +12,28 @@
...
+
 import tech.pegasys.eth2signer.core.signing.ArtifactSignerProvider;
+import tech.pegasys.eth2signer.core.signing.KeyType;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class PublicKeys {
-  private final ArtifactSignerProvider signerProvider;
+  private final ArtifactSignerProvider blsSignerProvider;
```
rain-on, line 24: Wondering if this is the right split (not sure) - i.e. should we maintain 2 different sets of keys everywhere - and build required handlers around each set.
I.e. this solution has a single handler, which then asks a single-keyhandler get me "bls" or "secp" keys - do we want to have 2 handlers, which are initialised with "SECP" and SecpSIgners, and another with "BLS" and BlsSigners?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTestBase.java
jframe: it's not, bad naming. fixed

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTestBase.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTestBase.java
jframe: don't think so

core/src/main/java/tech/pegasys/eth2signer/core/service/operations/PublicKeys.java
jframe: It was easier to do the split so that the rpc service and the rest api handler didn't need to have any knowledge about the signer provider split.

PR #154, opened 2020-08-10
Terminate Idle Connections

rain-on:
It was identified that Teku/Eth2Signer could fail due to neither side removing idle connections, with Teku creating a new connection for every signing operation.

This change means that Eth2Signer will  automatically close idle connections after 5 seconds.

### DIFF
--- a/core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
+++ b/core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
@@ -279,7 +279,6 @@ public class Runner implements Runnable {
         new HttpServerOptions()
             .setPort(config.getHttpListenPort())
             .setHost(config.getHttpListenHost())
-            .setTcpKeepAlive(false)
             .setIdleTimeout(config.getIdleConnectionTimeoutSeconds())
             .setIdleTimeoutUnit(TimeUnit.SECONDS)
             .setReuseAddress(true)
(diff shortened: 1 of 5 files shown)


core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -278,6 +279,9 @@ private HttpServer createServerAndWait(
         new HttpServerOptions()
             .setPort(config.getHttpListenPort())
             .setHost(config.getHttpListenHost())
+            .setTcpKeepAlive(false)
```
usmansaleem, line 282: should this be configurable or these defaults are reasonable?

rain-on: oops - configurable.

jframe, line 282: The default is false isn't it? So don't think needs to be set

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -278,6 +279,9 @@ private HttpServer createServerAndWait(
...
+            .setIdleTimeout(5)
```
jframe, line 283: Default seems a bit low. I think 30s at least would be a better default

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
rain-on: +1

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
rain-on: made this configurable

PR #157, opened 2020-08-12
Speed up file caching

rain-on:


core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -110,26 +114,20 @@ public DirectoryBackedArtifactSignerProvider(
   }
 
   public void cacheAllSigners() {
-    final Set<String> identifiers = allIdentifiers();
-    LOG.info("Loading {} signers", identifiers.size());
-    identifiers.parallelStream().forEach(this::cacheSigner);
+    final Collection<ArtifactSigner> allSigners =
+        findSigners(this::matchesFileExtension, signerParser::parse);
+    allSigners.forEach(
+        signer -> artifactSignerCache.put(normaliseIdentifier(signer.getIdentifier()), signer));
```
jframe, line 120: allIdentifiers does this the same way I think, so could probably moved allSigners into a single function that both can use. Then allIdentifiers can pull the identifiers out and this can just add them to the cache.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -143,26 +141,25 @@ private void cacheSigner(final String identifier) {
...
+          renderException(e, signerConfigFile.getFileName().toString());
         }
-      }
-      return signers;
-    } catch (final IOException | SecurityException e) {
-      LOG.warn("Error searching for signing metadata files: {}", e.getMessage());
-      return Collections.emptySet();
+      });
+    } catch (final IOException e) {
+      LOG.error("Didn't want this one", e);
```
jframe, line 157: Why has the error handling changed here? Don't we still want to return an empty set?

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
```diff
@@ -120,6 +120,7 @@ public void shutdown() {
       params.add("--data-path");
       params.add(dataPath.toAbsolutePath().toString());
     }
+    params.add("--key-cache-limit=5");
```
usmansaleem, line 123: was it temporary or do you want it to make part of acceptance test?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -57,117 +54,73 @@ public DirectoryBackedArtifactSignerProvider(
...
-        .collect(Collectors.toSet());
-  }
-
-  private Set<String> allIdentifiers() {
-    final Function<Path, String> getSignerIdentifier =
-        file -> FilenameUtils.getBaseName(file.toString());
-    return findSigners(this::matchesFileExtension, getSignerIdentifier).stream()
-        .filter(Objects::nonNull)
-        .map(identifier -> "0x" + normaliseIdentifier(identifier))
+        .map(id -> "0x" + id)
```
usmansaleem, line 76: nit: `!id.startsWith("0x") || !is.startsWith("0X") ? "ox" + id : id`. This will handle cases where id already have ox prefix. (or better use hasPrefix kind of method instead).

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -57,117 +54,73 @@ public DirectoryBackedArtifactSignerProvider(
...
     }
-  }
 
-  private Filter<Path> signerIdentifierFilenameFilter(final String signerIdentifier) {
-    return entry -> {
-      final String baseName = FilenameUtils.getBaseName(entry.toString());
-      return matchesFileExtension(entry)
-          && baseName.toLowerCase().endsWith(signerIdentifier.toLowerCase());
-    };
+    try (final Stream<Path> fileStream = Files.list(configsDirectory)) {
```
usmansaleem, line 105: you could have above logging statement here instead of running Files.list twice (L98, L105)

core/src/integrationTest/java/tech/pegasys/eth2signer/core/DirectoryBackedArtifactSigningProviderIntegrationTest.java
```diff
@@ -127,6 +128,7 @@ void missingHashicorpFingerprintFileWhichCannotBeCreatedShowsUsefulErrorLog() th
       final Path filename = createFileWithContent(signingMetadata);
 
       configsDirectory.toFile().setWritable(false);
+      signerProvider.loadSigners();
```
jframe, line 131: Is this needed now because we don't lazily load signers anymore? Can the loadSigners be moved into the setup method?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
jframe, line 76: why remove the normalising of the identifier?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
```diff
@@ -57,117 +54,73 @@ public DirectoryBackedArtifactSignerProvider(
...
+                  return mapper.apply(signerConfigFile);
+                } catch (final Exception e) {
+                  renderException(e, signerConfigFile.getFileName().toString());
+                  return null;
+                }
+              })
+          .filter(Objects::nonNull)
+          .collect(Collectors.toSet());
+    } catch (final IOException e) {
+      LOG.error("Didn't want this one", e);
```
jframe, line 121: Can improve this error message a bit I think. Perhaps something like "Error searching for signing metadata files"?

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
```diff
@@ -83,6 +81,7 @@ void signerReturnedForValidMetadataFile() throws IOException {
     final String filename = PUBLIC_KEY1;
     createFileInConfigsDirectory(filename);
     when(signerParser.parse(any())).thenReturn(artifactSigner);
+    signerProvider.loadSigners();
```
jframe, line 84: can this be moved into the setup?

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
```diff
@@ -208,13 +216,15 @@ void signerIdentifiersReturnedForMetadataFile() throws IOException {
     createFileInConfigsDirectory(PUBLIC_KEY1);
     when(signerParser.parse(any())).thenReturn(artifactSigner);
 
+    signerProvider.loadSigners();
```
jframe, line 219: A test that shows the loadSigners loads the signers into cache would be good

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
rain-on: agreed - we really should remove cache size etc from EVERYWHERE

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTest.java
```diff
@@ -114,4 +121,23 @@ public void azureKeysReturnAppropriatePublicKey() {
     final Response response = callApiPublicKeysWithoutOpenApiClientSideFilter(SECP256K1);
     validateApiResponse(response, containsInAnyOrder("0x" + PUBLIC_KEY_HEX_STRING));
   }
+
+  @Test
+  public void loadKeys() {
```
jframe, line 126: Can this test be given a more descriptive name to indicate it is about testing loading large number of keys? Perhaps  loadsLargeNumberOfKeys although maybe that's not even good enough 🤔

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTest.java
```diff
@@ -114,4 +121,23 @@ public void azureKeysReturnAppropriatePublicKey() {
...
+    final int keyCount = 10000;
+    final String[] publicKeys = new String[keyCount];
+    for (int i = 0; i < keyCount; i++) {
+      final Bytes bytes = Bytes.fromHexString(String.format("%064X", i + 1));
+      final BLSSecretKey key = BLSSecretKey.fromBytes(bytes);
+      final BLSKeyPair keyPair = new BLSKeyPair(key);
+      BLSPublicKey publicKey = keyPair.getPublicKey();
```
jframe, line 133: nit: final

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
rain-on: going to ignore 0X as I'd argue that's not a valid representation of a hex prefix.
Also - we _probably_ shouldn't ever mod the ID reported by the signer ... it should be passed through ... but don't want to deal with that now.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/signer/runner/Eth2SignerRunner.java
```diff
@@ -170,7 +171,7 @@ private void loadPortsFile() {
   }
 
   private void awaitPortsFile(final Path dataDir) {
-    final int secondsToWait = Boolean.getBoolean("debugSubProcess") ? 3600 : 30;
+    final int secondsToWait = Boolean.getBoolean("debugSubProcess") ? 3600 : 1000;
```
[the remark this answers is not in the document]
rain-on: removed

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTest.java
rain-on: done

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTest.java
rain-on: done

core/src/integrationTest/java/tech/pegasys/eth2signer/core/DirectoryBackedArtifactSigningProviderIntegrationTest.java
rain-on: Alas, it cannot be moved, as the config file(s) need to be created before loading.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
rain-on: because the signer was _created_ with a normalisedIdentifier.

rain-on: Given it is known that the artifacSigners are indexed by normalisedIdentifier - which definitively does not have an 0x, is there any need to do this check?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
rain-on: Will fall through into the emptySet returns (now)

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
rain-on: the above log line has been removed, but I found you couldn't embed it here, as it makes the stream "dirty" and it therefore can't then be used to access the files (as the stream is already deemed "used" by the call to count()).

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
rain-on: agreed - updated

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
rain-on: I _THINK_ no, as the signerParser.parse() needs to be configured prior to loading.

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
rain-on: The cache is irrelevant - if you load files, and ask for public keys - you should get back the same number - do you _care_ how/where they are stored?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProvider.java
jframe: If that's the assumption being made, is loadSigners is now mandatory? Should loadSigners be called in the constructor as without this it's not a valid ArtifactSignerProvider?

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryBackedArtifactSignerProviderTest.java
jframe: No I don't care how they are stored. But there is a function loadSigners that does something we should test that it does what we expect.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return result;
+  }
+
+  @Override
+  public Set<String> availableIdentifiers() {
+    return signers
+        .keySet()
+        .parallelStream()
+        .map(id -> "0x" + id)
```
usmansaleem, line 63: suggestion, only add 0x if it doesn't start with 0x? exactly reverse of normalizeIdentifier? i.e. `id -> id.toLowerCase().startsWith("0x") ? id : "0x" + id`

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoaderTest.java
```diff
@@ -0,0 +1,335 @@
...
+    final List<ArtifactSigner> signerList = Lists
+        .newArrayList(DirectoryLoader.loadFromDisk(configsDirectory, FILE_EXTENSION, signerParser));
+
+    assertThat(signerList.size()).isOne();
+    assertThat(signerList.get(0).getIdentifier()).isEqualTo("0x" + PUBLIC_KEY1);
+    verify(signerParser).parse(pathEndsWith(filename));
+  }
+
+  // TODO(tmm) - MOVE THIS TO ImmuntableArtifactSignerProvider
+//  @Test
```
usmansaleem, line 104: remove commented code??

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class ImmutableArtifactSignerProvider implements ArtifactSignerProvider {
```
jframe, line 28: Isn't this just a ArtifactSignerProvider perhaps the interface is meaningless now too. The fact it's immutable isn't really the distinguishing feature. Just a DefaultArtifactSignerProvider?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static final Logger LOG = LogManager.getLogger();
+  private final Map<String, ArtifactSigner> signers;
+
+  private ImmutableArtifactSignerProvider(final Map<String, ArtifactSigner> signers) {
+    this.signers = signers;
+  }
+
+  public static ImmutableArtifactSignerProvider wrap(final Collection<ArtifactSigner> signers) {
```
jframe, line 37: create instead of wrap? This is really the constructor it's not really wrapping as it's converting from a collection to a map.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
```diff
@@ -0,0 +1,66 @@
...
+    final Map<String, ArtifactSigner> signerMap =
+        signers
+            .parallelStream()
+            .collect(Collectors.toMap(ArtifactSigner::getIdentifier, Function.identity()));
+    return new ImmutableArtifactSignerProvider(signerMap);
+  }
+
+  @Override
+  public Optional<ArtifactSigner> getSigner(String identifier) {
```
jframe, line 46: nit: final

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private ImmutableArtifactSignerProvider(final Map<String, ArtifactSigner> signers) {
+    this.signers = signers;
+  }
+
+  public static ImmutableArtifactSignerProvider wrap(final Collection<ArtifactSigner> signers) {
+    final Map<String, ArtifactSigner> signerMap =
+        signers
+            .parallelStream()
+            .collect(Collectors.toMap(ArtifactSigner::getIdentifier, Function.identity()));
```
jframe, line 41: Probably need to normalise the identifier when it gets added to the map as well I'm thinking

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoader.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public static Collection<ArtifactSigner> loadFromDisk(
+      final Path rootDirectory, final String fileExtension, final SignerParser signerParser) {
+    final DirectoryLoader loader = new DirectoryLoader(rootDirectory, fileExtension, signerParser);
+    final Collection<ArtifactSigner> signers = loader.operateOnFileSubset();
+    LOG.info("Loaded {} signers", signers.size());
+    return signers;
+  }
+
+  public Collection<ArtifactSigner> operateOnFileSubset() {
```
jframe, line 55: This can be private

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoader.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.logging.log4j.Logger;
+
+public class DirectoryLoader {
+
+  private static final Logger LOG = LogManager.getLogger();
+  private final Path configsDirectory;
+  private final String fileExtension;
+  private final SignerParser signerParser;
+
+  public DirectoryLoader(
```
jframe, line 40: No real benefit of a constructor since the operateOnFileSubset is only used from the static method

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoaderTest.java
```diff
@@ -0,0 +1,341 @@
...
+    final List<ArtifactSigner> signerList =
+        Lists.newArrayList(
+            DirectoryLoader.loadFromDisk(configsDirectory, FILE_EXTENSION, signerParser));
+
+    assertThat(signerList.size()).isOne();
+    assertThat(signerList.get(0).getIdentifier()).isEqualTo("0x" + PUBLIC_KEY1);
+    verify(signerParser).parse(pathEndsWith(filename));
+  }
+
+  // TODO(tmm) - MOVE THIS TO ImmuntableArtifactSignerProvider
```
jframe, line 99: Have these tests been moved?

rain-on: have now

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoader.java
rain-on: That's true, but this means you don't need to pass everything _into_ everything, which made somethings nicer.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoader.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/ImmutableArtifactSignerProvider.java
rain-on: done (renamed to Default...)

acceptance-tests/src/test/java/tech/pegasys/eth2signer/tests/publickeys/PublicKeysAcceptanceTest.java
```diff
@@ -114,4 +121,23 @@ public void azureKeysReturnAppropriatePublicKey() {
     final Response response = callApiPublicKeysWithoutOpenApiClientSideFilter(SECP256K1);
     validateApiResponse(response, containsInAnyOrder("0x" + PUBLIC_KEY_HEX_STRING));
   }
+
+  @Test
+  public void ensureSystemCanLoadAndRerpotTenThousandKeysWithinExistingTimeLimits() {
```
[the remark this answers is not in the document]
rain-on: done

core/src/test/java/tech/pegasys/eth2signer/core/multikey/DirectoryLoaderTest.java
rain-on: done

PR #160, opened 2020-08-16
Perform single pass over key files during load

rain-on:


core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -29,18 +31,26 @@
...
+    this.secpFactory = secpFactory;
   }
 
   @Override
   public ArtifactSigner parse(final Path metadataPath) {
     try {
       final SigningMetadata metaDataInfo =
           OBJECT_MAPPER.readValue(metadataPath.toFile(), SigningMetadata.class);
-      return metaDataInfo.createSigner(artifactSignerFactory);
+      if (metaDataInfo.getKeyType().equals(KeyType.BLS)) {
```
jframe, line 49: No need to use .equals for enum

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -29,18 +31,26 @@
...
+        return metaDataInfo.createSigner(blsFactory);
```
jframe, line 50: nit: Think this would be a bit clearer with a ternary

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/AzureSecretSigningMetadata.java
```diff
@@ -31,7 +32,9 @@ public AzureSecretSigningMetadata(
       @JsonProperty("clientSecret") final String clientSecret,
       @JsonProperty("tenantId") final String tenantId,
       @JsonProperty("vaultName") final String vaultName,
-      @JsonProperty("secretName") final String secretName) {
+      @JsonProperty("secretName") final String secretName,
+      @JsonProperty(value = "keyType") final KeyType keyType) {
```
jframe, line 36: to be consistent with other @JsonProperty annotations can use leave out the value = and just use @JsonProperty("keyType"). Same for the other SigningMetadata classes

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -126,21 +134,45 @@ public void run() {
...
+          new Secp256k1ArtifactSignerFactory(
+              hashicorpConnectionFactory, config.getKeyConfigPath(), azureFactory);
+
+      final Collection<ArtifactSigner> signers =
+          SignerLoader.load(
+              config.getKeyConfigPath(),
+              "yaml",
+              new YamlSignerParser(blsArtifactSignerFactory, secp256k1ArtifactSignerFactory));
+
+      final DefaultArtifactSignerProvider blsArtifactSigners =
```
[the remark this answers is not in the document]
rain-on: oh, nice.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/AzureSecretSigningMetadata.java
rain-on: for sure - oops.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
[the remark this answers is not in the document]
rain-on: It is nice - however, if there are no BLS signers - then the "signersByType.get(BLS)" returns Null, not an empty list (same for Secp)
By the time you validate/convert between null and an empty list, you roughly end up with the same amount of code as here....

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -127,22 +139,74 @@ public void run() {
...
+                        } else if (i instanceof FcSecpArtifactSigner) {
+                          return SignerTypes.FC_SECP;
+                        } else {
+                          return SignerTypes.FC_BLS;
+                        }
+                      }));
+
+      final DefaultArtifactSignerProvider blsSignerProvider =
+          DefaultArtifactSignerProvider.create(
+              signersByType.get(SignerTypes.BLS) != null
```
jframe, line 191: you can use signersByType.getOrDefault(SignerTypes.BLS, emptyList()) to simplify this a bit and avoid that ternary

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -127,22 +139,74 @@ public void run() {
...
+                  blsArtifactSignerFactory,
+                  ethSecpArtifactSignerFactory,
+                  fcSecpArtifactSignerFactory));
+
+      final Map<SignerTypes, List<ArtifactSigner>> signersByType =
+          signers
+              .parallelStream()
+              .collect(
+                  Collectors.groupingBy(
+                      i -> {
```
jframe, line 177: switch make this a bit simpler

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -127,22 +139,74 @@ public void run() {
...
+                        if (i instanceof BlsArtifactSigner) {
```
jframe, line 178: Should a ArtifactSigner simply have an associated SignerType? Then we could just do groupingBy(ArtifactSigner::getSignerType)

core/src/main/java/tech/pegasys/eth2signer/core/SignerTypes.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.eth2signer.core;
+
+public enum SignerTypes {
+  BLS,
+  ETH,
```
jframe, line 17: why not SECP? this is consistent with how the other enum values are named. I guess most correct would be ETH_SECP since the secp signer is Ethereum specific.

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -12,35 +12,52 @@
...
-  final ArtifactSignerFactory artifactSignerFactory;
 
-  public YamlSignerParser(final ArtifactSignerFactory artifactSignerFactory) {
-    this.artifactSignerFactory = artifactSignerFactory;
+  private final BlsArtifactSignerFactory blsFactory;
+  private final Secp256k1ArtifactSignerFactory ethSecpFactory;
+  private final Secp256k1ArtifactSignerFactory fcSecpFactory;
+
+  public YamlSignerParser(
+      final BlsArtifactSignerFactory blsFactory,
```
jframe, line 44: This doesn't seem quite right. I wouldn't expect the YamlSignerParser to have deal with determining which factory to use for a keyType. Thinking that perhaps the keyType should be a property of the ArtifactSignerFactory then could just pass in a List<ArtifactSignerFactory> and then do a transform of List<ArtifactSignerFactory>.filter(metaDataInfo.getKeyType()).

core/src/test/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParserTest.java
```diff
@@ -250,8 +251,8 @@ void azureSecretMetadataInfoReturnsMetadata() throws IOException {
     azureMetaDataMap.put("secretName", "TEST-KEY");
     YAML_OBJECT_MAPPER.writeValue(filename.toFile(), azureMetaDataMap);
 
-    final ArtifactSigner result = signerParser.parse(filename);
-    assertThat(result).isEqualTo(artifactSigner);
+    final List<ArtifactSigner> result = signerParser.parse(filename);
+    assertThat(result).containsOnly(artifactSigner);
     verify(artifactSignerFactory).create(hasCorrectAzureMetadataArguments());
   }
```
jframe, line 258: Can a test be added for a Parser that produces multiple ArtifactSigners

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
rain-on: so many API I don't even think about ... but yes, nicer.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
rain-on: Yeah, I was thinking the same ... but equally, it only fixes this one bit of yuck, which is pretty well controlled (and only invoked once).

core/src/main/java/tech/pegasys/eth2signer/core/SignerTypes.java
rain-on: Done - but what is BLS? Eth_BLS?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
rain-on: yeah that's nicer.

core/src/main/java/tech/pegasys/eth2signer/core/signing/LoadedSigners.java
```diff
@@ -0,0 +1,111 @@
...
+                Collectors.groupingBy(
+                    i -> {
+                      if (i instanceof BlsArtifactSigner) {
+                        return SignerTypes.BLS;
+                      } else if (i instanceof EthSecpArtifactSigner) {
+                        return SignerTypes.ETH_SECP;
+                      } else if (i instanceof FcSecpArtifactSigner) {
+                        return SignerTypes.FC_SECP;
+                      } else {
+                        return SignerTypes.FC_BLS;
```
usmansaleem, line 91: ideally we should throw an IllegalStateException if we encountered an invalid (or potentially new signer type).

core/src/test/java/tech/pegasys/eth2signer/core/multikey/BlsArtifactSignerFactoryTest.java
```diff
@@ -94,7 +94,7 @@ void createsArtifactSignerFromKeyStoreUsingAbsolutePaths() {
   void nonExistentKeyStoreThrowsError() {
     final Path nonExistingKeystoreFile = configDir.resolve("someNonExistingKeystore");
     final FileKeyStoreMetadata fileKeyStoreMetadata =
-        new FileKeyStoreMetadata(nonExistingKeystoreFile, passwordFile);
+        new FileKeyStoreMetadata(nonExistingKeystoreFile, passwordFile, null);
```
usmansaleem, line 97: null seems to bad here, why can't we use Type.BLS to explicitly state we are dealing with BLS type?

core/src/main/java/tech/pegasys/eth2signer/core/signing/LoadedSigners.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  public DefaultArtifactSignerProvider getEthSignerProvider() {
+    return getSignerProvider(SignerTypes.ETH_SECP);
+  }
+
+  public DefaultArtifactSignerProvider getFcSecpSignerProvider() {
+    return getSignerProvider(SignerTypes.FC_SECP);
+  }
+
```
usmansaleem, line 107: should there be a method for `getFcBlsSignerProvider` as well?

usmansaleem (approved):
LGTM - minor suggestions

core/src/main/java/tech/pegasys/eth2signer/core/signing/LoadedSigners.java
rain-on: done

core/src/test/java/tech/pegasys/eth2signer/core/multikey/BlsArtifactSignerFactoryTest.java
rain-on: done

core/src/main/java/tech/pegasys/eth2signer/core/signing/LoadedSigners.java
rain-on: yes - you're right, this should exist now.

core/src/main/java/tech/pegasys/eth2signer/core/Runner.java
```diff
@@ -126,37 +124,30 @@ public void run() {
...
-          new KeyIdentifiers(new UnsupportedArtifactSignerProvider(), fcSecpSignerProvider);
+          new KeyIdentifiers(fcBlsSignerProvider, fcSecpSignerProvider);
 
       final OpenAPI3RouterFactory openApiRouterFactory =
           createOpenApiRouterFactory(vertx, ethKeyIdentifiers, blsSigner, secpSigner);
       registerHttpHostAllowListHandler(openApiRouterFactory);
       final Router router = openApiRouterFactory.getRouter();
 
-      // register non-openapi routes ...
+      // register non-
```
jframe, line 150: nit: looks like some of the comment got dropped?

core/src/main/java/tech/pegasys/eth2signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -12,35 +12,44 @@
...
-    this.artifactSignerFactory = artifactSignerFactory;
+  private final Collection<AbstractArtifactSignerFactory> signerFactories;
+
+  public YamlSignerParser(final Collection<AbstractArtifactSignerFactory> signerFactories) {
+    this.signerFactories = signerFactories;
   }
 
   @Override
-  public ArtifactSigner parse(final Path metadataPath) {
+  public List<ArtifactSigner> parse(final Path metadataPath) {
```
jframe, line 44: nit: is order important, will a collection do?

jframe (approved):
LGTM

PR #164, opened 2020-08-20
FileCoin Address decoding - Json annotated exception

usmansaleem:


jframe (approved):
LGTM
just need to fix the tests

Issue #112, opened 2020-07-01
FC sign API


arash009:
Add support for FC WalletSign RPC

The API details are defined here https://github.com/filecoin-project/lotus/blob/master/api/api_full.go#L168

This will require implementation of the FC address format https://filecoin-project.github.io/specs/#appendix__address

Example request:
`{ "jsonrpc": "2.0", "method": "Filecoin.WalletSign", "params": ["t3udmivibpkn7fgfc5heyz3znivubf5slvxvtiry3cwwqbvojlfikr56u34madf6ssvuwdo6btzpmm2is7rriq", "NDI="], "id": 3 }`

response:
`{"jsonrpc":"2.0","result":{"Type":2,"Data":"r8bzkJ4IrDmMseKcNbDLe3iMmTd0t5x1Dd7HUz5BhgMWi7xXjQ/Ey8quiBF8R+AWBO+5uOvOfKDcKpGJ70jh3wj3Q0/04NnxWrs8YrVgYNBsxOnZq3ODL1cZmIGl4/QA"},"id":3}`

arash009:
sign
signmessage

PR #166, opened 2020-08-20
Fc bls artifact signing

jframe:
Implement Filecoin BLS signing as a FcBLsArtifactSigner. This only implements signing, the FC signing RPC still needs to be implemented.

* Uses the existing Teku and Milagro BLS implementation
* It is a different signer as the DST used for FC is different specifically it uses BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_ and not BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_ that Eth2 does.

#112


core/src/main/java/tech/pegasys/eth2signer/core/signing/FcBlsArtifactSigner.java
```diff
@@ -34,6 +44,9 @@ public String getIdentifier() {
 
   @Override
   public ArtifactSignature sign(Bytes data) {
-    throw new UnsupportedOperationException("BLS Signing for FileCoin has not been implemented.");
+    final G2Point hashInGroup2 = new G2Point(hashToG2(data, DST));
```
rain-on, line 47: final on function param (final Bytes data) {

core/src/test/java/tech/pegasys/eth2signer/core/signing/FcBlsArtifactSignerTest.java
```diff
@@ -43,4 +45,17 @@ void identifierReturnsBlsAddress() {
         "t3sjhgtrk5fdio52k5lzanh7yy4mj4rqbiowd6odddzprrxejgbjbl2irr3gmpbf7epigf45oy7asljj3v3lva";
     assertThat(identifier).isEqualTo(expectedBlsAddress);
   }
+
+  @ParameterizedTest
+  @CsvSource({
+    "'',qqsXe9+kDAGOh8uKmI6+oGN90ydJ0jRELvxvf2VwAd+g3WRKT/bi65RlhiGTrzzCFP4gY0BiAMe3+ghInVFLs1CXBcYGol+RiPoFUew7UZ6ON1zq5Tjhs5F2vWFj1qt8",
+    "NDI=,p/LeJS8KPzuOco2qpqjhvJxFrA5qj++LpeVoActWkDRQTLhuGsnwAvY1kluQ6PntAqISjvR/RU0kzgkR38F8Hm5NuNPsu9/BEiU+IQheNeD7OoG8THq4OuZMbISg7uR/"
+  })
+  void signsData(final String message, final String expectedSignature) {
```
rain-on, line 54: any better name for the test? messagesComputedSignatureMatchesReferenceSignature

core/src/main/java/tech/pegasys/eth2signer/core/signing/FcBlsArtifactSigner.java
jframe: done

PR #172, opened 2020-08-26
Fix base32 decode hack for lowercase values

### LINKED ISSUE #170: FC base32 decode should handle lowercase

jframe:
This removes hack we had for handling lowercase base32 decoding in the Filecoin address by using guava base32 encoder which supports lowercase characters.

Fix #170

### DIFF
--- a/core/src/test/java/tech/pegasys/eth2signer/core/signing/FilecoinAddressTest.java
+++ b/core/src/test/java/tech/pegasys/eth2signer/core/signing/FilecoinAddressTest.java
@@ -56,7 +56,7 @@ class FilecoinAddressTest {
 
   @Test
   void addressWithInvalidChecksumThrowsError() {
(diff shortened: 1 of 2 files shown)


core/src/test/java/tech/pegasys/eth2signer/core/signing/FilecoinAddressTest.java
```diff
@@ -56,7 +56,7 @@ void referenceTestsForActorAddresses(final String address, final String payload)
 
   @Test
   void addressWithInvalidChecksumThrowsError() {
-    assertThatThrownBy(() -> FilecoinAddress.decode("f17uoq6tp427uzv7fztkbsnn64iwotfrristwprz"))
+    assertThatThrownBy(() -> FilecoinAddress.decode("f17uoq6tp427uzv7fztkbsnn64iwotfrristwpr"))
```
jframe, line 59: Had to change this value as this was actually an invalid length for base32 and we don't have any padding in filecoin addresses.

usmansaleem (approved):
LGTM

PR #177, opened 2020-08-27
FC Verify without loaded keys

jframe:
Change FC verify RPC so that keys don't need to be loaded for FC verify to work.

### DIFF
--- a/core/src/main/java/tech/pegasys/eth2signer/core/service/jsonrpc/FcJsonRpc.java
+++ b/core/src/main/java/tech/pegasys/eth2signer/core/service/jsonrpc/FcJsonRpc.java
@@ -12,6 +12,8 @@
  */
 package tech.pegasys.eth2signer.core.service.jsonrpc;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import tech.pegasys.eth2signer.core.signing.ArtifactSignature;
 import tech.pegasys.eth2signer.core.signing.ArtifactSigner;
 import tech.pegasys.eth2signer.core.signing.ArtifactSignerProvider;
@@ -20,8 +22,6 @@ import tech.pegasys.eth2signer.core.signing.SecpArtifactSignature;
 import tech.pegasys.eth2signer.core.signing.filecoin.FilecoinAddress;
 import tech.pegasys.eth2signer.core.signing.filecoin.FilecoinVerify;
 import tech.pegasys.eth2signer.core.signing.filecoin.exceptions.FilecoinSignerNotFoundException;
-import tech.pegasys.eth2signer.core.util.ByteUtils;
-import tech.pegasys.signers.secp256k1.api.Signature;
 import tech.pegasys.teku.bls.BLSSignature;
 
 import java.util.Optional;
@@ -33,8 +33,6 @@ import com.github.arteam.simplejsonrpc.core.annotation.JsonRpcService;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes;
-import org.apache.tuweni.bytes.Bytes32;
-import org.web3j.utils.Numeric;
 
 @JsonRpcService
 public class FcJsonRpc {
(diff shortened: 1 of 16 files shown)


core/src/main/java/tech/pegasys/eth2signer/core/service/jsonrpc/FcJsonRpc.java
```diff
@@ -107,34 +109,32 @@ public boolean filecoinWalletVerify(
...
+    final Bytes signature = Bytes.fromBase64String(filecoinSignature.getData());
 
-    if (signer.isPresent()) {
-      final Bytes bytesToVerify = Bytes.fromBase64String(dataToVerify);
-      final ArtifactSignature artifactSignature = createArtifactSignature(filecoinSignature);
-      return signer.get().verify(bytesToVerify, artifactSignature);
-    } else {
-      throw new FilecoinSignerNotFoundException();
+    switch (address.getProtocol()) {
+      case SECP256K1:
```
rain-on, line 116: do you need to verify that the address and the "signature type" align - i.e. both are BLS?

core/src/main/java/tech/pegasys/eth2signer/core/service/jsonrpc/FcJsonRpc.java
```diff
@@ -107,34 +109,32 @@ public boolean filecoinWalletVerify(
...
+        return FilecoinVerify.verify(
+            address, Bytes.fromBase64String(dataToVerify), createSecpArtifactSignature(signature));
```
rain-on, line 118: this function can PROBABLY take bytes - as the string decode will happen in the Json decoding module

core/src/main/java/tech/pegasys/eth2signer/core/service/jsonrpc/FcJsonRpc.java
```diff
@@ -107,34 +109,32 @@ public boolean filecoinWalletVerify(
...
-            new Signature(
-                Numeric.toBigInt(v.toArrayUnsafe()),
-                Numeric.toBigInt(r.toArrayUnsafe()),
-                Numeric.toBigInt(s.toArrayUnsafe())));
-      case BLS_VALUE:
-        return new BlsArtifactSignature(BLSSignature.fromBytes(signature));
-      default:
-        throw new IllegalArgumentException("Invalid Signature type");
-    }
+  private SecpArtifactSignature createSecpArtifactSignature(final Bytes signature) {
```
rain-on, line 129: suspect this belongs somewhere other than the JsonRpc handling class.
In GENERAL, I like the idea of keeping this class super thin, and simply being entry points rather than doing work ... haven't been great on that so far though

core/src/main/java/tech/pegasys/eth2signer/core/signing/filecoin/FilecoinVerify.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.web3j.crypto.ECDSASignature;
+import org.web3j.crypto.Sign;
+import org.web3j.utils.Numeric;
+
+public class FilecoinVerify {
+  private static final Logger LOG = LogManager.getLogger();
+  private static final Bytes DST =
```
rain-on, line 36: this feels like a BLS constant ... can/should it live somewhere centrally for filecoin/bls?

core/src/main/java/tech/pegasys/eth2signer/core/signing/filecoin/FilecoinVerify.java
```diff
@@ -0,0 +1,73 @@
...
+      final SecpArtifactSignature artifactSignature) {
+    final tech.pegasys.signers.secp256k1.api.Signature signatureData =
+        artifactSignature.getSignatureData();
+
+    final ECDSASignature initialSignature =
+        new ECDSASignature(signatureData.getR(), signatureData.getS());
+    final ECDSASignature canonicalSignature = initialSignature.toCanonicalised();
+
+    final int recId = signatureData.getV().intValue();
+    final byte[] digest = Blake2b.sum256(message).toArrayUnsafe();
```
rain-on, line 59: nit: wondering if we should do this up front - then call a separate "recover" function - that way, its easy to spilt data-management from signature management - the sub-function would have the same prototype as this function, but wouldn't blake2b the supplied message...

rain-on (approved):
Small comment, but fine to go in as is.

core/src/main/java/tech/pegasys/eth2signer/core/signing/filecoin/FilecoinVerify.java
jframe: done

PR #178, opened 2020-08-27
Move to using blocking handlers

rain-on:
The OpenApiRouter doesn't use blocking handlers, meaning HTTP receptions are handled in the eventloop, rather than in a worker thread.

This change moves all handlers to be "blocking" and thus execute on worker threads.


usmansaleem (approved):
LGTM

jframe (approved):

PR #184, opened 2020-08-31
FC BLS Performance tests

jframe:
Basic performance test for Filecoin signing with BLS

* Uses gatling for the performance test
* Runs Eth2Signer as process where the test is run. Ideally we will run Eth2Signer on already setup instance with know hardware configuration in the future
* Creates 1000 random BLS keys
* Signs FC requests using a random local unencrypted BLS key with random data
* Simulation of 10 requests/sec for 60s followed by a burst of 100 requests/sec over 5s


rain-on (approved):
LGTM - stick with the scala

PR #186, opened 2020-09-02
Catch more info for a failed sign message comparative test

rain-on:
Lotus Auth token can now be set via env var
such that comparative tests can be run against
a live, real, Lotus instance (rather than a devnet variant).

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/lotus/FilecoinJsonRequests.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/lotus/FilecoinJsonRequests.java
@@ -116,8 +116,8 @@ public class FilecoinJsonRequests {
     post.setEntity(new StringEntity(request, Charsets.UTF_8));
     post.setHeader(HttpHeaders.CONTENT_TYPE, MediaType.JSON_UTF_8.toString());
     final String authToken = System.getenv("ETH2SIGNER_BEARER_TOKEN");
-    if(authToken != null) {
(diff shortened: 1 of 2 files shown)


acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/lotus/FilecoinJsonRequests.java
```diff
@@ -115,6 +115,10 @@ public static String executeRawJsonRpcRequest(final String url, final String req
     final HttpPost post = new HttpPost(url);
     post.setEntity(new StringEntity(request, Charsets.UTF_8));
     post.setHeader(HttpHeaders.CONTENT_TYPE, MediaType.JSON_UTF_8.toString());
+    final String authToken = System.getenv("ETH2SIGNER_BEARER_TOKEN");
```
usmansaleem, line 118: can we move it up as `private static final Optional<String> authToken = Optional.ofNullable(System.getenv("ETH2SIGNER_BEARER_TOKEN"));`. This will allow any one reading this test class to read it at the beginning instead of buried down here?

usmansaleem (approved):
LGTM ... minor suggestion around moving as constant.

acceptance-tests/src/test/java/tech/pegasys/eth2signer/dsl/lotus/FilecoinJsonRequests.java
rain-on: done

usmansaleem (approved):
LGTM

PR #197, opened 2020-09-07
Api renames

usmansaleem:
# PR Description

- Changed OpenAPI paths to seperate ETH1 and ETH2 operations.
`/api/v1/eth2/sign/{identifier}`
`/api/v1/eth1/sign/{identifier}`
`/api/v1/eth2/publicKeys`
`/api/v1/eth1/publicKeys`
`/upcheck`

- Following paths are removed
`/signer/sign/{identifier}`
`/signer/publicKeys/{keyType}`

- Removed OpenAPI equivalent JSON-RPC operations from `/rpc/v1`.  (Filecoin JSON-RPC operations are still available at `/rpc/v1/filecoin`)
- Update metrics
- Update relevant test cases.


core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -147,20 +140,16 @@ public void run() {
       final FileCoinArtifactSignerProvider fcArtifactSignerProvider =
           new FileCoinArtifactSignerProvider(fcBlsSignerProvider, fcSecpSignerProvider);
 
+      // create router from OpenApi spec - web3signer.yaml
```
rain-on, line 143: not sure if this is for now - but I'd love to see us split the configuration/buildup into Eth1, Eth2 and Filecoin "partitions" - thus the secp and the bls identifiers would never need to be shipped to a function together - they are fundamentally different, individual entities.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -179,51 +168,82 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
...
-    openAPI3RouterFactory.addFailureHandlerByOperationId(UPCHECK_OPERATION_ID, errorHandler);
-
-    // public key handler
-    openAPI3RouterFactory.addHandlerByOperationId(
-        GET_PUBLIC_KEYS_OPERATION_ID,
-        new BlockingHandlerDecorator(new GetPublicKeysHandler(keyIdentifiers), false));
-    openAPI3RouterFactory.addFailureHandlerByOperationId(
-        GET_PUBLIC_KEYS_OPERATION_ID, errorHandler);
+    // public key list handler
+    addEth1PublicKeysListHandler(openAPI3RouterFactory, secpIdentifiers, errorHandler);
```
rain-on, line 181: _personally_ prefer to think about this stuff as eth1 and eth2, rather than signing and public_keys APIs ... but that may be a personal preference in how to split the problem space.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -179,51 +168,82 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
...
+      final SignerForIdentifier<BlsArtifactSignature> blsSigner,
+      final MetricsSystem metricsSystem,
+      final LogErrorHandler errorHandler) {
     openAPI3RouterFactory.addHandlerByOperationId(
-        SIGN_FOR_IDENTIFIER_OPERATION_ID,
+        ETH2_SIGN.name(),
         new BlockingHandlerDecorator(
-            new SignForIdentifierHandler(blsSigner, metricsSystem, "bls"), true));
+            new SignForIdentifierHandler(
+                blsSigner, new OpenApiOperationsMetrics(metricsSystem, BLS)),
```
rain-on, line 203: what is OpenApiOperationsMetrics? And how does it work - FEELS like this might be another change for another PR?

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -179,51 +168,82 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
...
+            false));
+    openAPI3RouterFactory.addFailureHandlerByOperationId(ETH2_SIGN.name(), errorHandler);
+  }
+
+  private void addEth1SignHandler(
```
rain-on, line 208: this feels really similar to the Eth2 function - can it be parameterised and duplication removed?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/UpcheckHandler.java
```diff
@@ -15,16 +15,12 @@
...
 import io.vertx.core.Handler;
 import io.vertx.ext.web.RoutingContext;
 
 public class UpcheckHandler implements Handler<RoutingContext> {
-  final Upcheck upcheck = new Upcheck();
-
   @Override
   public void handle(final RoutingContext routingContext) {
-    routingContext.response().putHeader(CONTENT_TYPE, TEXT_PLAIN_UTF_8).end(upcheck.status());
+    routingContext.response().putHeader(CONTENT_TYPE, TEXT_PLAIN_UTF_8).end("OK");
```
rain-on, line 24: So do we need the upcheck class anymore? Or has it been removed?

core/src/main/java/tech/pegasys/web3signer/core/service/http/metrics/OpenApiOperationsMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+import org.hyperledger.besu.plugin.services.metrics.OperationTimer;
+
+public class OpenApiOperationsMetrics {
+  private static Counter signersLoadedCounter;
+
+  private final Counter malformedRequestCounter;
+  private final OperationTimer signingTimer;
+  private final Counter missingSignerCounter;
+
+  public OpenApiOperationsMetrics(final MetricsSystem metricsSystem, final KeyType keyType) {
```
rain-on, line 29: are these actually "OpenApiOperationsMetrics" - or maybe "RouteMetrics"? Not sure ... or are they really SignOperationMetrics?

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/FcJsonRpc.java
```diff
@@ -102,7 +101,7 @@ public boolean filecoinWalletHas(@JsonRpcParam("address") final String address)
   public FilecoinSignedMessage filecoinSignMessage(
       @JsonRpcParam("identifier") final String identifier,
       @JsonRpcParam("message") final FilecoinMessage message) {
-    metrics.incwWalletSignMessageRequestCounter();
+    metrics.incWalletSignMessageRequestCounter();
```
rain-on, line 104: thanks :) for the typo fix.

core/src/main/resources/openapi/web3signer.yaml
```diff
@@ -57,23 +57,79 @@ paths:
...
           in: 'path'
           required: true
-          description: 'Type of public keys to return'
+          description: 'Key for which data to sign'
           schema:
-            type: string
-            enum:
-              - BLS
-              - SECP256K1
+            type: 'string'
```
rain-on, line 73: might need to ask JFrame to take a look at this bit,  but looks ok to me.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
usmansaleem: I simply wrapped scattered signing/http metrics into this class. Makes the runner class and SignForIdentifierHandler a bit simpler with regards to creating the matrices.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/UpcheckHandler.java
usmansaleem: The Upcheck class has been removed.

core/src/main/java/tech/pegasys/web3signer/core/service/http/metrics/OpenApiOperationsMetrics.java
usmansaleem: I think all of them are Sign operation metrics except for 1 which you had categorised under "http". Suggestions for class name??

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -179,51 +168,82 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
...
-        });
-    openAPI3RouterFactory.addFailureHandlerByOperationId(
-        SIGN_FOR_IDENTIFIER_OPERATION_ID, errorHandler);
+        ETH1_LIST.name(),
+        new BlockingHandlerDecorator(new PublicKeysListHandler(secpIdentifiers), false));
+    openAPI3RouterFactory.addFailureHandlerByOperationId(ETH1_LIST.name(), errorHandler);
+  }
 
-    return openAPI3RouterFactory;
+  private void addEth2PublicKeysListHandler(
```
jframe, line 232: similar to the above comment, looks like some dupe here, these publicKeysListHandlers are exactly the same except for some  naming

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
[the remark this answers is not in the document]
usmansaleem: done. HttpApiMetrics makes more sense.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
usmansaleem: refactored out to common logic.

jframe (approved):

PR #199, opened 2020-09-07
Use a fully populated json body

rain-on:
Currently the Sign handler assumes that the body of the sign requests contains only the data to be signed (the public key is specified in the path).

In order for slashing to be effectively, additional data is required (such has type fo artifact being signed, source/target epoch etc).

This change brings in a fully-formed signing body such that slashing can be enforced.


rain-on:
Superseded by events (JFrame's various PRs).

PR #200, opened 2020-09-08
Add no-op slashing protection

rain-on:
This includes structural code, but does not offer slashing protection.


commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerCommand.java
```diff
@@ -172,6 +172,14 @@
       arity = "1")
   private final FilecoinNetwork filecoinNetwork = FilecoinNetwork.TESTNET;
 
+  @Option(
+      names = {"--slashing-protection-enabled"},
+      description =
```
usmansaleem, line 177: it's a good idea to use `paramLabel=<BOOL>`. Otherwise picocli prints variable name.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -115,6 +117,14 @@ public void run() {
             config.getMetricsHostAllowList());
 
     final MetricsSystem metricsSystem = metricsEndpoint.getMetricsSystem();
+
+    final SlashingProtection slashingProtection;
```
usmansaleem, line 121: may I suggest Optional<SlashingProtection> instead of null? or may be a createNoOpSlashingProtection?

usmansaleem: nevermind ... I see that you have eventually used Optional. So its OK for now.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SignForIdentifierHandler.java
```diff
@@ -59,11 +67,21 @@ public void handle(final RoutingContext routingContext) {
...
           .ifPresentOrElse(
-              signature ->
+              signature -> {
+                if (slashingProtection.isPresent()) {
+                  if (slashingProtection.get().maySignAttestation(null, null, null)) {
+                    routingContext
+                        .response()
+                        .putHeader(CONTENT_TYPE, TEXT_PLAIN_UTF_8)
+                        .end(signature);
+                  }
```
usmansaleem, line 77: Is it a good idea to put an else here with a TODO?. This is the spot which should raise appropriate http error code (as per the spec).

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerCommand.java
```diff
@@ -172,6 +172,14 @@
...
+          "Set to true if all signing operations should be validated against historic data, prior to producing signature"
```
jframe, line 178: Think we should make it clear that this only applies to Eth2

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
jframe: Rather we didn't have to pass null around

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.web3signer.slashingprotection;
+
+public class SlashingProtectionFactory {
+
+  public static SlashingProtection createSlashingProtection() {
+    return new NoOpSlashingProtection();
```
usmansaleem, line 18: ahhh ...

usmansaleem (approved):
looks good. minor suggestion to put paramLabel.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SignForIdentifierHandler.java
```diff
@@ -59,11 +67,21 @@ public void handle(final RoutingContext routingContext) {
       signerForIdentifier
           .sign(normaliseIdentifier(identifier), data)
           .ifPresentOrElse(
-              signature ->
+              signature -> {
+                if (slashingProtection.isPresent()) {
+                  if (slashingProtection.get().maySignAttestation(null, null, null)) {
```
jframe, line 72: how do we check that block signing is allowed?

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerCommand.java
rain-on: done

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SignForIdentifierHandler.java
```diff
@@ -59,11 +67,21 @@ public void handle(final RoutingContext routingContext) {
...
+                    routingContext
+                        .response()
+                        .putHeader(CONTENT_TYPE, TEXT_PLAIN_UTF_8)
+                        .end(signature);
```
jframe, line 76: This block is the same as below can we remove this duplication somehow?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SignForIdentifierHandler.java
rain-on: we also have an API for that, but can;t use it - as we don't know if its a block or attestation yet

jframe: yeah good point

jframe (approved):

PR #201, opened 2020-09-09
Eth2 slashing protection API

jframe:
Adds additional fields to the Eth2 signing request to support slashing

#192

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfiguration.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfiguration.java
@@ -32,6 +32,7 @@ public class SignerConfiguration {
   private Optional<TlsOptions> serverTlsOptions;
   private Optional<TlsCertificateDefinition> overriddenCaTrustStore;
   private final int metricsPort;
+  private final boolean slashingProtectionEnabled;
 
   public SignerConfiguration(
       final String hostname,
@@ -42,7 +43,8 @@ public class SignerConfiguration {
       final List<String> metricsHostAllowList,
       final boolean metricsEnabled,
       final Optional<TlsOptions> serverTlsOptions,
-      final Optional<TlsCertificateDefinition> overriddenCaTrustStore) {
+      final Optional<TlsCertificateDefinition> overriddenCaTrustStore,
+      final boolean slashingProtectionEnabled) {
     this.hostname = hostname;
     this.httpRpcPort = httpRpcPort;
     this.httpHostAllowList = httpHostAllowList;
@@ -52,6 +54,7 @@ public class SignerConfiguration {
     this.metricsEnabled = metricsEnabled;
     this.serverTlsOptions = serverTlsOptions;
     this.overriddenCaTrustStore = overriddenCaTrustStore;
+    this.slashingProtectionEnabled = slashingProtectionEnabled;
   }
 
   public String hostname() {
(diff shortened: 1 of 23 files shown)


core/src/main/java/tech/pegasys/web3signer/core/service/JacksonSerializers.java
```diff
@@ -0,0 +1,133 @@
...
+  public static class Base64Serialiser extends JsonSerializer<Bytes> {
+    @Override
+    public void serialize(
+        final Bytes value, final JsonGenerator gen, final SerializerProvider serializers)
+        throws IOException {
+      gen.writeString(value.toBase64String());
+    }
+  }
+
+  public static class HexDeserialiser extends JsonDeserializer<Bytes> {
```
usmansaleem, line 52: won't it conflict with Base64Deserialiser? both of them are converting Bytes?

core/src/main/java/tech/pegasys/web3signer/core/service/http/SigningJsonRpcModule.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.units.bigints.UInt64;
+
+public class SigningJsonRpcModule extends SimpleModule {
+
+  public SigningJsonRpcModule() {
+    super("SigningJsonRpcModule");
+    addDeserializer(Bytes.class, new HexDeserialiser());
+    addSerializer(Bytes.class, new HexSerialiser());
+    addDeserializer(UInt64.class, new StringUInt64Deserializer());
```
usmansaleem, line 30: oh got it. Only FC will use Base64Deserializer.

core/src/main/java/tech/pegasys/web3signer/core/service/JacksonSerializers.java
jframe: Only if you use both at the same time. This class doesn't register any of the deserializers it's merely a factory, that is up to the Jackson module.

usmansaleem (approved):
Looks good to me.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfiguration.java
```diff
@@ -32,6 +32,7 @@
   private Optional<TlsOptions> serverTlsOptions;
   private Optional<TlsCertificateDefinition> overriddenCaTrustStore;
   private final int metricsPort;
+  private final boolean slashingProtectionEnabled;
```
rain-on, line 35: you can probably just put in a URL - if its not null, then slashing is enabled, and you can set the url at the same time - I think I have this in one of my PRs?

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/Web3SignerRunner.java
```diff
@@ -120,6 +120,8 @@ public void shutdown() {
       params.add("--data-path");
       params.add(dataPath.toAbsolutePath().toString());
     }
+    params.add("--slashing-protection-enabled");
```
rain-on, line 123: again - take a look at one of my PRs, think this is done with a URL

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfiguration.java
jframe: removing it, don't need to disable slashing protection as still using noop slashing protection. will just use the url when we merge in real slashing protection

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/Web3SignerRunner.java
jframe: removed it

PR #204, opened 2020-09-10
Move loading into Runners

rain-on:
Moving the filecoin network from the base command intot he filecoin subcommand necessitated file loading to be moved into the specific "Runners".

This has the knock on effect of ONLY loading relevant files (which is actually a required behaviour).


core/src/main/java/tech/pegasys/web3signer/core/FilecoinRunner.java
```diff
@@ -31,17 +47,48 @@
...
-  public FilecoinRunner(final Config config) {
+  public FilecoinRunner(final Config config, final FilecoinNetwork network) {
     super(config);
+    this.network = network;
+  }
+
+  @Override
+  public ArtifactSignerProvider loadSigners(
+      final Config config, final Vertx vertx, final MetricsSystem metricsSystem) {
+    final AzureKeyVaultSignerFactory azureFactory = new AzureKeyVaultSignerFactory();
```
jframe, line 60: nit: would be nice nice to avoid so much dupe with the other runners. also probably not a big deal either.

jframe (approved):
LGTM

PR #208, opened 2020-09-11
Remove unnecessary slashing validation

jframe:
Change slashing protection validation

* So this only happens if slashing protection is enabled
* Type parameter is no longer required if slashing protection isn't enabled
* Default slashing protection enabled to false until slashing protection is finished


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
```diff
@@ -71,8 +69,12 @@ public void handle(final RoutingContext routingContext) {
           .ifPresentOrElse(
               signature -> {
                 if (slashingProtection.isPresent()) {
-                  if (maySign(identifier, eth2SigningRequestBody)) {
-                    respondWithSignature(routingContext, signature);
+                  try {
+                    if (maySign(identifier, eth2SigningRequestBody)) {
+                      respondWithSignature(routingContext, signature);
+                    }
+                  } catch (IllegalArgumentException e) {
```
usmansaleem, line 76: nit: final

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
```diff
@@ -71,8 +69,12 @@ public void handle(final RoutingContext routingContext) {
           .ifPresentOrElse(
               signature -> {
                 if (slashingProtection.isPresent()) {
-                  if (maySign(identifier, eth2SigningRequestBody)) {
-                    respondWithSignature(routingContext, signature);
+                  try {
+                    if (maySign(identifier, eth2SigningRequestBody)) {
+                      respondWithSignature(routingContext, signature);
+                    }
```
usmansaleem, line 75: should there be an "else" here? responding with appropriate `routingContext.fail(code)` ??

usmansaleem (approved):
LGTM - Minor comments.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
jframe: done

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
jframe: done

PR #213, opened 2020-09-14
Register validators in slashing database

jframe:
Registers validator publics in the database.

* Create postgresql schema for slashing protection data
* On start inserts any missing validator public keys into validator table. This is done so we can use numeric id the validators.id for index in the slashing data tables
* Uses Hikari for connection pooling


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+      } catch (SQLException re) {
+        LOG.error("Rollback of validator registration failed", re);
+      }
+      throw new IllegalStateException("Failed registering validators", e);
+    }
+  }
+
+  private List<Bytes> retrieveMissingValidators(
+      final Connection connection, final List<Bytes> publicKeys) throws SQLException {
+    final String inSql = String.join(",", Collections.nCopies(publicKeys.size(), "?"));
```
rain-on, line 65: does the preparedstatement help here? I.e you know how may items to insert .. can you not just do it? Or is that a recipe for insecurity.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -41,13 +45,34 @@
...
+
+  @Option(
+      names = {"--slashing-db-username"},
+      description = "The username to use when connecting to the slashing database")
+  private String slashingDbUser;
+
+  @Option(
+      names = {"--slashing-db-password"},
+      description = "The password to use when connecting to the slashing database")
+  private String slashingDbPassword;
```
usmansaleem, line 63: is it ok to specify password on cli? or should we instead specify a password file.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbConnection.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class DbConnection {
+
+  public static Supplier<Connection> createConnectionSupplier(
+      final String jdbcUrl, final String username, final String password) {
+    final DataSource datasource = createDataSource(jdbcUrl, username, password);
+    return () -> {
+      try {
+        return datasource.getConnection();
+      } catch (SQLException e) {
```
usmansaleem, line 30: nit: final

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public ValidatorsDao(final Supplier<Connection> connectionSupplier) {
+    this.connectionSupplier = connectionSupplier;
+  }
+
+  public void registerValidators(final List<Bytes> validators) {
+    final Connection connection = connectionSupplier.get();
+    try {
+      connection.setAutoCommit(false);
+      final List<Bytes> missingValidators = retrieveMissingValidators(connection, validators);
```
rain-on, line 43: missingvalidators implies that the validator is in our list but not in db? [can it be named more clearly?]
Assumingly, the other way doesn't matter (i.e. in table, but not in validators?)

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+
+      final PreparedStatement insertStatement = connection.prepareStatement(INSERT_VALIDATOR);
+      for (Bytes missingValidator : missingValidators) {
+        insertStatement.setBytes(1, missingValidator.toArrayUnsafe());
+        insertStatement.addBatch();
+      }
+      insertStatement.executeBatch();
+      connection.commit();
+    } catch (SQLException e) {
```
usmansaleem, line 52: nit: final

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+      "SELECT id, public_key FROM validators WHERE public_key IN (%s)";
+  private static final String INSERT_VALIDATOR = "INSERT INTO validators (public_key) VALUES (?)";
+  private final Supplier<Connection> connectionSupplier;
+
+  public ValidatorsDao(final Supplier<Connection> connectionSupplier) {
+    this.connectionSupplier = connectionSupplier;
+  }
+
+  public void registerValidators(final List<Bytes> validators) {
+    final Connection connection = connectionSupplier.get();
```
usmansaleem, line 40: shouldn't connection be wrapped in try/resource block?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+    try {
+      connection.setAutoCommit(false);
+      final List<Bytes> missingValidators = retrieveMissingValidators(connection, validators);
+
+      final PreparedStatement insertStatement = connection.prepareStatement(INSERT_VALIDATOR);
```
usmansaleem, line 45: insertStatement should also be wrapped in try/resource block.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+      for (Bytes missingValidator : missingValidators) {
+        insertStatement.setBytes(1, missingValidator.toArrayUnsafe());
+        insertStatement.addBatch();
+      }
+      insertStatement.executeBatch();
+      connection.commit();
+    } catch (SQLException e) {
+      LOG.error("Failed registering validators. Check slashing database is correctly setup.", e);
+      try {
+        connection.rollback();
```
rain-on, line 55: why bother rolling back?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+      } catch (SQLException re) {
```
usmansaleem, line 56: nit: final

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+        LOG.error("Rollback of validator registration failed", re);
+      }
+      throw new IllegalStateException("Failed registering validators", e);
+    }
+  }
+
+  private List<Bytes> retrieveMissingValidators(
+      final Connection connection, final List<Bytes> publicKeys) throws SQLException {
+    final String inSql = String.join(",", Collections.nCopies(publicKeys.size(), "?"));
+    final PreparedStatement selectStatement =
```
usmansaleem, line 66: try/resource block

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,82 @@
...
+        connection.prepareStatement(String.format(SELECT_VALIDATOR, inSql));
+    for (int i = 0; i < publicKeys.size(); i++) {
+      selectStatement.setBytes(i + 1, publicKeys.get(i).toArrayUnsafe());
+    }
+
+    final ResultSet resultSet = selectStatement.executeQuery();
```
usmansaleem, line 72: try/resource block

usmansaleem:
@jframe where are we loading postgresql jdbc driver?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
usmansaleem: if you don't rollback, its a recipe to bring database to its knees. The transaction remains open unless a commit or rollback is performed.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -41,13 +45,34 @@
       arity = "1")
   private boolean slashingProtectionEnabled = false;
 
+  @Option(
```
usmansaleem, line 48: may I also recommend a cli option for JDBC Driver class? Which can be statically load using Class.forName

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
jframe: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
jframe: can't in this case as I need to call rollback or commit and need to the connection accessible from the catch block

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -79,6 +87,13 @@ protected ArtifactSignerProvider loadSigners(
             "yaml",
             new YamlSignerParser(List.of(artifactSignerFactory)));
 
+    final List<Bytes> validators =
```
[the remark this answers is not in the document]
jframe: don't want to do a loop as that would be slower and want this all to happen in the context of a transaction

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
usmansaleem: I dont think (I might be wrong though) we are using config file/env variable in web3signer (the change was in ethsigner).

jframe: There's a YamlConfigFileDefaultProvider and EnvironmentVariableDefaultProvider so it should be working for web3signer.

jframe:
> @jframe where are we loading postgresql jdbc driver?

Hikari uses the jdbc url to locate the jdbc driver on the classpath

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfigurationBuilder.java
```diff
@@ -33,6 +33,9 @@
   private List<String> httpHostAllowList = emptyList();
   private TlsOptions serverTlsOptions;
   private TlsCertificateDefinition overriddenCaTrustStore;
+  private String slashingProtectionDbUrl = "";
```
rain-on, line 36: do we have checks which ensure if a username is specified, then all other fields must be too?

build.gradle
```diff
@@ -416,7 +416,7 @@ distributions {
     contents {
       from("./LICENSE") { into "." }
       from("build/reports/license/license-dependency.html") { into "." }
-      from("./migrations") { into "./migrations" }
+      from("./slashing-protection/src/main/resources/migrations") { into "./migrations" }
```
usmansaleem, line 419: I recall one of this section ends up in the final jar .... and another section is required for "installdist". As installdist uses a different mechanism to place these bits.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -41,13 +36,31 @@
...
+      names = {"--slashing-protection-db-url"},
+      description = "The jdbc url to use to connect to the slashing protection database",
+      paramLabel = "<storage label>",
```
usmansaleem, line 42: storage label ??? you meant `paramLabel = "<jdbc url>"` ??

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -46,14 +49,41 @@
...
+      final Config config,
+      final boolean slashingProtectionEnabled,
+      final String slashingProtectionDbUrl,
+      final String slashingProtectionDbUser,
+      final String slashingProtectionDbPassword) {
     super(config);
-    this.slashingProtection = slashingProtection;
+    this.slashingProtection =
+        createSlashingProtection(
+            slashingProtectionEnabled,
```
rain-on, line 68: can someone disable slashing protection, but still supply all fields? (yes, this one is my bugbear)

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -41,13 +36,31 @@
       arity = "1")
   private boolean slashingProtectionEnabled = false;
```
rain-on, line 37: think slashing needs to be enabled by default - do we want to raise a ticket to change this over?

gradle/versions.gradle
```diff
@@ -96,5 +96,17 @@ dependencyManagement {
...
+    dependency 'org.postgresql:postgresql:42.2.16.jre7'
+
+    dependencySet(group: 'org.jdbi', version: '3.14.4') {
+      entry 'jdbi3-core'
+      entry 'jdbi3-sqlobject'
+      entry 'jdbi3-testing'
+    }
+
+    dependency 'com.opentable.components:otj-pg-embedded:0.11.3'
+    dependency 'org.flywaydb:flyway-core:6.1.1'
```
rain-on, line 110: what are you using flyway for?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jdbi.v3.sqlobject.transaction.Transaction;
+
+public interface ValidatorsDao {
+
+  @Transaction
+  default void registerMissingValidators(final List<Bytes> validators) {
+    final List<Validator> registeredValidators = retrieveValidators(validators);
+    final List<Bytes> validatorsMissingFromDb = new ArrayList<>(validators);
+    registeredValidators.forEach(v -> validatorsMissingFromDb.remove(v.getPublicKey()));
+    registerValidators(validatorsMissingFromDb);
```
usmansaleem, line 32: if we are using latest version of PostgreSQL (BUT it won't be compatible with other database), we can potentially use `INSERT INTO .... ON CONFLICT DO NOTHING`. which would avoid two calls from database.

usmansaleem (approved):
LGTM - minor comments

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
jframe: yes I think we do. but after it's implemented. so think raise another to enable it by default once we are done with all the pieces.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
jframe: yes I did. done

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
jframe: yes they could. not sure what the answer is here. we could display a warning if any of these values are set and slashing protection isn't enabled?

gradle/versions.gradle
jframe: for setting up the database tables in the dao unit tests

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/ValidatorsDao.java
jframe: we can't use that with an auto-incrementing id though. but perhaps we can increment the id ourselves as long as we have the appropriate transaction isolation is enabled.

build.gradle
jframe: Do think we need to do this as installDist is intended for developers.

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
ajsutton: It's extremely useful to be able to have a config file with lots of things specified and then just change an `enabled: true` to `enabled: false` rather than having to delete any other relevant options.  Similarly it's very common to have a long CLI command and then just go back and edit one flag - very annoying and user hostile when it then errors and demands you delete flags that could have easily been ignored.

jframe: changed to only validate the slashing db url if slashing is enabled. this is required if it is enabled to so we can't ignore this option otherwise web3signer will just fail to startup. if slashing isn't enabled we just ignore the options.

PR #220, opened 2020-09-21
Slashing protection for blocks

jframe:
Adds Eth2 slashing protection for block proposals


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -36,12 +48,40 @@ public boolean maySignAttestation(
...
+            return false;
+          } else {
+            final long id = validatorId.get();
+            final Optional<SignedBlock> existingBlock =
+                signedBlocksDao.findExistingBlock(h, id, blockSlot);
+
+            // same slot and signing_root is allowed for broadcasting previously signed block
+            // otherwise if slot and different signing_root then this is a double block proposal
+            final boolean isValid =
+                existingBlock.map(block -> block.getSigningRoot().equals(signingRoot)).orElse(true);
```
usmansaleem, line 67: orElse true or false??

usmansaleem: so if existing block is empty, it is valid!

usmansaleem (approved):
LGTM

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
jframe: Should be orElse true. What it's saying is if we haven't signed this block before then it's valid to sign it. We are only trying to avoid signing the same block for the slot and public key more than once.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -36,12 +48,40 @@ public boolean maySignAttestation(
...
+          if (validatorId.isEmpty()) {
+            return false;
+          } else {
+            final long id = validatorId.get();
+            final Optional<SignedBlock> existingBlock =
+                signedBlocksDao.findExistingBlock(h, id, blockSlot);
+
+            // same slot and signing_root is allowed for broadcasting previously signed block
+            // otherwise if slot and different signing_root then this is a double block proposal
+            final boolean isValid =
```
rain-on, line 66: is it worth changing this to be really clear language _in code_:
final Optional<SignedBlock> priorSigningForSlot = ...;
if(priorSigningForSlot.isPresent() {
   return priorSigningForSlot.signingRoot().equals();
}
return true;

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlock.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public long getValidatorId() {
+    return validatorId;
+  }
+
+  public void setValidatorId(final long validatorId) {
+    this.validatorId = validatorId;
+  }
+
+  public UInt64 getSlot() {
```
rain-on, line 40: why are there setters? Is that required for jdbi?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlocksDao.java
```diff
@@ -0,0 +1,44 @@
...
+    return handle
+        .createQuery(
+            "SELECT validator_id, slot, signing_root FROM signed_blocks WHERE validator_id = ? AND slot = ?")
+        .bind(0, validatorId)
+        .bind(1, slot)
+        .mapToBean(SignedBlock.class)
+        .findFirst();
+  }
+
+  public void insertBlockProposal(
```
rain-on, line 34: could this not take a SignedBlock?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlocksDao.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.units.bigints.UInt64;
+import org.jdbi.v3.core.Handle;
+
+public class SignedBlocksDao {
+
+  public Optional<SignedBlock> findExistingBlock(
```
rain-on, line 23: is there any value in this simply returning the SIgningRoot, rather than a whole SignedBlock - or is there no performance gains or otherwise by reducing it?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -36,12 +48,40 @@ public boolean maySignAttestation(
 
   @Override
   public boolean maySignBlock(
-      final String publicKey, final Bytes signingRoot, final UInt64 blockSlot) {
-    return true;
+      final Bytes publicKey, final Bytes signingRoot, final UInt64 blockSlot) {
+    return jdbi.inTransaction(
```
rain-on, line 52: confirming this is atomic - i.e. another signing can't "interlace" this transaction?
Does a 'big' transaction like this affect other threads accessing the attestation table?
I.e. do we need to mimise the lock size?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlock.java
jframe: The setters are needed for jbdi unfortunately

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
jframe: It is atomic across transactions. Using serializable transaction isolation level https://www.postgresql.org/docs/12/transaction-iso.html#XACT-SERIALIZABLE in postgresql. This won't affect other tables other than the tables in the transaction. We should minimise the lock size and so have take the retrieve validators out of this transaction.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlocksDao.java
jframe: Think the performance gains are insignificant if any at all. From an API point I prefer the SignedBlocksDao returning a SignedBlock object.

PR #222, opened 2020-09-22
YubiHSM2 Integration

usmansaleem:
YubiHSM2 integration. 


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -50,7 +55,7 @@
 
   public void createUnencryptedYamlFileAt(
       final Path metadataFilePath, final String privateKey, final KeyType keyType) {
-    final Map<String, String> signingMetadata = new HashMap<>();
+    final Map<String, Serializable> signingMetadata = new HashMap<>();
```
rain-on, line 58: nit Not sure how this helped - seems to be a tagging interface?

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -179,6 +184,49 @@ public void createAzureKeyYamlFileAt(
...
   }
 
+  public void createYubiHsmYamlFileAt(
+      final Path metadataFilePath,
+      final Path destinationYubiShellSimulator,
+      final KeyType keyType) {
+    final int opaqueObjId = keyType == KeyType.BLS ? 1 : 2;
+
+    final Map<String, Serializable> yaml =
+        Map.of(
```
rain-on, line 194: Map.of is quick, but hard to read - can we change it to be 2 per line, or just use the map.put?

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -179,6 +184,49 @@ public void createAzureKeyYamlFileAt(
...
+            "type",
+            "yubihsm2",
+            "yubiShellBinaryPath",
+            destinationYubiShellSimulator.toString(),
```
rain-on, line 198: not sure I love the executable being in the config file - this means you can't reuse this file-shape if we changed to using the yubihsm as an embedded library - maybe this should come in via Env Var (with default value?)

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -179,6 +184,49 @@ public void createAzureKeyYamlFileAt(
...
+            "password",
+            "opaqueObjId",
+            opaqueObjId,
+            "keyType",
+            keyType.name());
+
+    createYamlFile(metadataFilePath, yaml);
+  }
+
+  public static Path copyYubiHsmSimulator(final Path destDir) throws IOException {
```
rain-on, line 213: without going into it - this feels quite "hacky" i.e. copying files, injecting shebang, chmod ... is there any other way of doing this?

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -97,7 +102,7 @@ public void createKeyStoreYamlFileAt(
     final Path passwordFile = metadataFilePath.getParent().resolve(passwordFilename);
     createPasswordFile(passwordFile, password);
 
-    final Map<String, String> signingMetadata = new HashMap<>();
+    final Map<String, Serializable> signingMetadata = new HashMap<>();
```
rain-on, line 105: this can stay Map<String, String> (as String is derived from Serializable) (and elsewhere too)

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -17,13 +17,15 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.web3j.crypto.Sign.publicKeyFromPrivate;
 import static org.web3j.crypto.Sign.signedMessageToKey;
+import static tech.pegasys.web3signer.dsl.utils.MetadataFileHelpers.copyYubiHsmSimulator;
```
rain-on, line 20: do we need to add tests for Filecoin Signing with YubiHsm?

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/YubiHsm2SigningMetadata.java
```diff
@@ -0,0 +1,115 @@
...
+  private final Short opaqueObjId;
+
+  // following fields are optional and will be populated if defined in configuration file
+  private Optional<OutputFormat> outformat = Optional.empty();
+  private Optional<String> caCertPath = Optional.empty();
+  private Optional<String> proxyUrl = Optional.empty();
+
+  @JsonCreator
+  public YubiHsm2SigningMetadata(
+      @JsonProperty("yubiShellBinaryPath") final String yubiShellBinaryPath,
```
rain-on, line 41: I know we have this problem elsewhere - but @JsonProperty defaults "required" to FALSE ... which is probably wrong - so thinking we need to add required=true to each @JsonProperty (but not KeytType)

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
usmansaleem: agreed.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
usmansaleem: We can assume that the binary `yubihsm-shell` is available on global PATH, and can be overridden from environment variable `WEB3SIGNER_YUBISHELL_PATH`

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
usmansaleem: The other way is to either have a shell file (which can work on both mac and linux) or a native go binary for mac and linux.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
usmansaleem: I think we should.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -179,6 +183,42 @@ public void createAzureKeyYamlFileAt(
...
+    yaml.put("opaqueObjId", String.valueOf(opaqueObjId));
+    yaml.put("keyType", keyType.name());
+
+    createYamlFile(metadataFilePath, yaml);
+  }
+
+  public static Path copyYubiHsmSimulator(final Path destDir) throws IOException {
+    final Path sourceYubiShellSimulator =
+        Path.of(Resources.getResource("YubiShellSimulator").getPath());
+    final Path destinationYubiShellSimulator = destDir.resolve("YubiShellSimulator");
```
jframe, line 208: Could we just run the YubiShellSimulator in a thread so we have to don't have mess with copying files etc.

acceptance-tests/src/test/resources/YubiShellSimulator
```diff
@@ -0,0 +1,91 @@
...
+        if (!argsMap.containsKey("--connector")) {
+            System.err.println("--connctor is missing");
+            System.exit(-1);
+        }
+        if (!argsMap.containsKey("--authkey")) {
+            System.err.println("--authkey is missing");
+            System.exit(-1);
+        }
+        if (!argsMap.containsKey("--object-id")) {
+            System.err.println("--bject-id is missing");
```
jframe, line 47: typo: should be "--object-id is missing"

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/YubiHsm2SigningMetadata.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  @SuppressWarnings("UnunsedMethod")
+  @JsonSetter("proxyUrl")
+  public void setProxyUrl(final String proxyUrl) {
+    this.proxyUrl = Optional.ofNullable(proxyUrl);
+  }
+
+  @JsonSetter("outformat")
+  public void setOutformat(final OutputFormat outformat) {
```
jframe, line 75: outputFormat for config property would be clearer I think

usmansaleem: renamed to `outputFormat`.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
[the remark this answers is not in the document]
usmansaleem: moved it to acceptance test base.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
usmansaleem: we can't. The Signer's YubiHSM class needs to invoke it as an external process. I can create a small native binary instead?

acceptance-tests/src/test/resources/YubiShellSimulator
usmansaleem: fixed.

jframe (approved):
LGTM

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -184,49 +179,20 @@ public void createAzureKeyYamlFileAt(
...
-            "http://localhost:12345",
-            "authKey",
-            1,
-            "password",
-            "password",
-            "opaqueObjId",
-            opaqueObjId,
-            "keyType",
-            keyType.name());
+    final Map<String, String> yaml = new HashMap<>();
```
rain-on, line 185: nit: happy for this to stay as serializable if that makes life easier

acceptance-tests/src/test/resources/YubiShellSimulator.java
```diff
@@ -0,0 +1,90 @@
+/*
```
rain-on, line 1: is this a duplicate from somewhere? Or is this the only one?
If its the only one, that's awesome, but if we duplicated it, might be nice to work out how to work around that

rain-on (approved):

acceptance-tests/src/test/resources/YubiShellSimulator.java
usmansaleem: Not duplicated ... its very dumbed-down simulator that will only work for Web23Signer ATs

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
usmansaleem: Its all good .... the YAML format doesn't use quotes for strings/numbers (so we are good).

PR #232, opened 2020-09-28
Multiple artifact signers from metadata

usmansaleem:
-- Modify framework to allow multiple artifact-signers to be created from single metadata file
-- Will help in avoiding roundtrip to remote key store/vault services (such as upcoming FSecure Interlock HSM)


usmansaleem:
Implemented alternate approach to cache session with same credentials.

PR #235, opened 2020-09-29
create test which attempts to replicate codefi issue

rain-on:


rain-on:
Determined issue was linked to postgres version

PR #237, opened 2020-10-01
Fix transaction contention caused by serializable transaction isolation level

jframe:
This fixes an issue where failures were occurring when on updating the slashing protection db.

* Changes the transaction isolation to read committed
* To allow for safe updates a database lock is used against the validator id so that transactions for the same validator id are effectively done serially
* Add transaction retry, needed for the validator registration which still using serializable transaction isolation

### DIFF
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbConnection.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbConnection.java
@@ -23,6 +23,7 @@ import com.zaxxer.hikari.HikariDataSource;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.argument.Arguments;
 import org.jdbi.v3.core.mapper.ColumnMappers;
+import org.jdbi.v3.core.transaction.SerializableTransactionRunner;
 
 public class DbConnection {
 
@@ -41,6 +42,7 @@ public class DbConnection {
     jdbi.getConfig(ColumnMappers.class)
         .register(new BytesColumnMapper())
         .register(new UInt64ColumnMapper());
+    jdbi.setTransactionHandler(new SerializableTransactionRunner());
   }
 
   private static DataSource createDataSource(
@@ -49,7 +51,6 @@ public class DbConnection {
     dataSource.setJdbcUrl(jdbcUrl);
     dataSource.setUsername(username);
     dataSource.setPassword(password);
-    dataSource.setTransactionIsolation("TRANSACTION_SERIALIZABLE");
     return dataSource;
   }
 }
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
@@ -218,7 +218,7 @@ public class DbSlashingProtection implements SlashingProtection {
 
   private void lockForValidator(
       final Handle handle, final LockType lockType, final long validatorId) {
-    final String lockTypePrefix = lockType == LockType.BLOCK ? "0" : "1";
+    final String lockTypePrefix = Integer.toString(lockType.ordinal());
     final String lockId = lockTypePrefix + validatorId;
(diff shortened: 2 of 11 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -185,4 +215,11 @@ private long validatorId(final Bytes publicKey) {
     }
     return validatorId;
   }
+
+  private void lockForValidator(
+      final Handle handle, final LockType lockType, final long validatorId) {
+    final String lockTypePrefix = lockType == LockType.BLOCK ? "0" : "1";
```
usmansaleem, line 221: nit: you can potentially expose additional method in LockType enum, lockTypePrefix which can return appropriate prefix.

usmansaleem (approved):
LGTM - minor nit-pick suggestions.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbConnection.java
```diff
@@ -23,6 +23,7 @@
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.argument.Arguments;
 import org.jdbi.v3.core.mapper.ColumnMappers;
+import org.jdbi.v3.core.transaction.SerializableTransactionRunner;
```
rain-on, line 26: can this be removed?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
rain-on: maybe just used locktype.ordinal()?

slashing-protection/src/main/resources/migrations/postgresql/V1__initial.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE signed_blocks (
     slot NUMERIC(20) NOT NULL,
     signing_root BYTEA NOT NULL,
     FOREIGN KEY(validator_id) REFERENCES validators(id),
-    UNIQUE (validator_id, slot)
+    UNIQUE (validator_id, slot, signing_root)
```
rain-on, line 11: does this protect anything anymore?

jframe: no, not now we have the lock

slashing-protection/src/main/resources/migrations/postgresql/V1__initial.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE signed_blocks (
     slot NUMERIC(20) NOT NULL,
     signing_root BYTEA NOT NULL,
     FOREIGN KEY(validator_id) REFERENCES validators(id),
-    UNIQUE (validator_id, slot, signing_root)
+    UNIQUE (validator_id, slot)
```
rain-on: the constraint is nice ... but it's mitigated by the locking, isn't it?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -218,7 +218,7 @@ private long validatorId(final Bytes publicKey) {
 
   private void lockForValidator(
       final Handle handle, final LockType lockType, final long validatorId) {
-    final String lockTypePrefix = lockType == LockType.BLOCK ? "0" : "1";
+    final String lockTypePrefix = Integer.toString(lockType.ordinal());
```
rain-on, line 221: The problem is that the lockId is NOT guaranteed to fit into a long anymore.

rain-on (approved):

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
jframe: validator_id doesn't need to be long so I'll change to be an int then I can use the two int arg version of the lock to avoid any issue with size of the value.

PR #239, opened 2020-10-04
Cleaned up logging

rain-on:


acceptance-tests/src/test/resources/log4j2.xml
```diff
@@ -2,13 +2,30 @@
...
   <Loggers>
+    <Logger name="tech.pegasys.web3signer.SubProcessLog" level="INFO" additivity="false">
+      <AppenderRef ref="SubProcessConsole" />
+    </Logger>
+    <Logger name="com.zaxxer.hikari" level="${dependency.log.level}" additivity="false"/>
+    <Logger name="org.flywaydb.core" level="${dependency.log.level}" additivity="false"/>
+    <Logger name="org.apache.http" level="${dependency.log.level}" additivity="false"/>
+    <Logger name="io.restassured.internal" level="${dependency.log.level}" additivity="false"/>
+    <Logger name="io.netty" level="${dependency.log.level}" additivity="false"/>
+    <Logger name="io.swagger.v3.parser.OpenAPIV3Parser" level="${dependency.log.level}" additivity="false"/>
```
jframe, line 25: Do we just want set all swagger logs to warn? i.e. io.swagger as the name. And similarly for com.atlassian.oai?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
```diff
@@ -93,7 +94,7 @@ public void handle(final RoutingContext routingContext) {
                 }
               },
               () -> {
-                LOG.trace("Identifier not found {}", identifier);
+                LOG.warn("Identifier not found {}", identifier);
```
jframe, line 97: Don't think we need this log at all as when no signer was found we log an error in DefaultArtifactSignerProvider, log at warn here and then finally log in the LogErrorHandler so all up we get 3 logs entries.

e.g.
```
2020-10-06 15:22:55.650+10:00 | vert.x-worker-thread-5 | ERROR | DefaultArtifactSignerProvider | No signer was loaded matching identifier '0x900884a719c418272c1c2b038f3dd5629a4bf2465d30963d181f454565f4236c3441aa49049c9fedac57c381fb067a2a'
2020-10-06 15:22:55.650+10:00 | vert.x-worker-thread-5 | WARN  | Eth2SignForIdentifierHandler | Identifier not found 0x900884a719c418272c1c2b038f3dd5629a4bf2465d30963d181f454565f4236c3441aa49049c9fedac57c381fb067a2a
2020-10-06 15:22:55.650+10:00 | vert.x-eventloop-thread-4 | ERROR | LogErrorHandler | Failed request: http://localhost:9000/api/v1/eth2/sign/0x900884a719c418272c1c2b038f3dd5629a4bf2465d30963d181f454565f4236c3441aa49049c9fedac57c381fb067a2a
```

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/Web3SignerProcessRunner.java
```diff
@@ -115,7 +115,7 @@ private void printOutput() {
         new BufferedReader(new InputStreamReader(process.getInputStream(), UTF_8))) {
       String line = in.readLine();
       while (line != null) {
-        PROCESS_LOG.info("Web3SignerProc: {}", line);
+        PROCESS_LOG.info(line);
```
jframe, line 118: Is this confusing without the prefix? How do we know which log lines are from web3signer?

acceptance-tests/src/test/resources/log4j2.xml
usmansaleem: 👍

acceptance-tests/src/test/resources/log4j2.xml
```diff
@@ -2,13 +2,30 @@
...
     <Property name="root.log.level">DEBUG</Property>
+    <Property name="dependency.log.level">WARN</Property>
   </Properties>
 
   <Appenders>
     <Console name="Console" target="SYSTEM_OUT">
-      <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSSZZZ} | %t | %-5level | %c{1} | %msg%n" />    </Console>
+      <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSSZZZ} | %t | %-5level | %c{1} | %msg%n" />
+    </Console>
+    <Console name="SubProcessConsole" tareget="SYSTEM_OUT">
```
usmansaleem, line 12: mispell .... `target`

usmansaleem:
The log4j files in test resources can be appended with `-test` i.e. `log4j2-test.xml`. This allows these files to take precedence over standard files if they are on classpath.

acceptance-tests/src/test/resources/log4j2.xml
usmansaleem: I think all of the "dependent" libraries logs should be warn ... unless we want to debug those libraries themselves.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
usmansaleem: 👍

acceptance-tests/src/test/resources/log4j2.xml
rain-on: yeah, that's fine.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/Eth2SignForIdentifierHandler.java
rain-on: done

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/Web3SignerProcessRunner.java
rain-on: They still have "subprocess" as the first entry in the line due to the name of the logger.

acceptance-tests/src/test/resources/log4j2.xml
rain-on: good catch :)

usmansaleem (approved):
minor comments, looks good otherwise.

usmansaleem (approved):
LGTM

jframe (approved):

PR #241, opened 2020-10-07
Config file handling for subcommands

jframe:
Support subcommands command line options in config files. The key name is prefixed with the subcommand name e.g. eth2.slashing-protection-enabled: true

#223


usmansaleem (approved):
LGTM

Issue #254, opened 2020-10-12
Build errors on specific Java versions


bgravenorst:
`./gradlew build` fails when using JDK12 or JDK15.

When running `./gradlew build` using JDK12 the build fails with `errorprone` errors. With JDK 15 it fails with `jacoco` library errors.

This means the following Web3Signer instructions will not work for JDK12 and JDK15:
https://docs.web3signer.consensys.net/en/latest/HowTo/Get-Started/Build-From-Source/

PR #255, opened 2020-10-12
[#254] Upgrade gradle, errorprone and jacoco versions

usmansaleem:
- Gradle version to 6.6.1
- error prone to 2.4.0
- Jacoco to 0.8.6

Allows successful compilation from JDK11 to JDK 15.


gradlew.bat
```diff
@@ -84,6 +84,7 @@ set CMD_LINE_ARGS=%*
 
 set CLASSPATH=%APP_HOME%\gradle\wrapper\gradle-wrapper.jar
 
+
```
rain-on, line 87: is the extra white space necessary? IMHO it could be removed, but not too worried.

rain-on (approved):

gradlew.bat
usmansaleem: auto generated by gradle wrapper

PR #257, opened 2020-10-12
F-Secure Interlock for Armory II keystore integration

usmansaleem:
- [x] Configuration
-  [x] Signer implementation
-  [x] Acceptance test cases


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -193,6 +193,23 @@ public void createYubiHsmYamlFileAt(final Path metadataFilePath, final KeyType k
     createYamlFile(metadataFilePath, yaml);
   }
 
+  public void createInterlockYamlFileAt(
```
rain-on, line 196: Interlock _can_ be run simply as an app on a PC ... so thinking we should refer to this HSM style as Interlock, rather than an Armory HSM 👍

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
```diff
@@ -124,6 +125,18 @@ public void ableToSignUsingYubiHsm() {
     signAndVerifySignature(yubiHsmShellEnvMap());
   }
 
+  @Test
+  @Disabled("Requires access to Interlock on Armory II")
```
rain-on, line 129: can we prove this with interlock running as an app? 
edit: apperas to only have pre-builts for arm ... so maybe not

But would be better to not have this test disabled.

core/src/main/java/tech/pegasys/web3signer/core/Eth1Runner.java
```diff
@@ -81,16 +80,10 @@ protected Router populateRouter(final Context context) {
...
-        new HashicorpConnectionFactory(vertx);
 
     final Secp256k1ArtifactSignerFactory ethSecpArtifactSignerFactory =
         new Secp256k1ArtifactSignerFactory(
-            hashicorpConnectionFactory,
-            config.getKeyConfigPath(),
-            azureFactory,
-            EthSecpArtifactSigner::new,
-            true);
+            vertx, config.getKeyConfigPath(), azureFactory, EthSecpArtifactSigner::new, true);
```
rain-on, line 86: _prefer_ to see us passing in HashicorpFactory, and also the Interlock factory - as it means the lower layer doesn't need to be Vertx specific - but I can appreciate there maybe reasons it doesn't work so well.

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -156,15 +155,10 @@ private ArtifactSignerProvider loadSigners(
...
-    final HashicorpConnectionFactory hashicorpConnectionFactory =
-        new HashicorpConnectionFactory(vertx);
 
     final AbstractArtifactSignerFactory artifactSignerFactory =
         new BlsArtifactSignerFactory(
-            config.getKeyConfigPath(),
-            metricsSystem,
-            hashicorpConnectionFactory,
-            BlsArtifactSigner::new);
+            vertx, config.getKeyConfigPath(), metricsSystem, BlsArtifactSigner::new);
```
rain-on, line 161: comment as above.

core/src/main/java/tech/pegasys/web3signer/core/FilecoinRunner.java
```diff
@@ -100,19 +99,17 @@ private Router registerFilecoinJsonRpcRoute(
   private ArtifactSignerProvider loadSigners(
       final Config config, final Vertx vertx, final MetricsSystem metricsSystem) {
     final AzureKeyVaultSignerFactory azureFactory = new AzureKeyVaultSignerFactory();
-    final HashicorpConnectionFactory hashicorpConnectionFactory =
-        new HashicorpConnectionFactory(vertx);
 
     final AbstractArtifactSignerFactory blsArtifactSignerFactory =
         new BlsArtifactSignerFactory(
+            vertx,
```
rain-on, line 105: comment as above

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/Secp256k1ArtifactSignerFactory.java
```diff
@@ -38,12 +38,12 @@
   private final boolean needToHash;
 
   public Secp256k1ArtifactSignerFactory(
-      final HashicorpConnectionFactory connectionFactory,
+      final Vertx vertx,
```
rain-on, line 41: as before - prefer to not see Vertx get passed all the way down here

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
jframe: Could we use wiremock or similar here so we don't have to disable the test?

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/BlsArtifactSignerFactory.java
```diff
@@ -98,6 +98,16 @@ public ArtifactSigner create(final YubiHsm2SigningMetadata yubiHsm2SigningMetada
     }
   }
 
+  @Override
+  public ArtifactSigner create(final InterlockSigningMetadata interlockSigningMetadata) {
+    try (TimingContext ignored = privateKeyRetrievalTimer.labels("interlock").startTimer()) {
+
```
jframe, line 104: nit: seems like unnecessary extra line here

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AbstractArtifactSignerFactory.java
```diff
@@ -106,6 +111,19 @@ protected Bytes extractBytesFromVault(final YubiHsm2SigningMetadata metadata) {
     }
   }
 
+  protected Bytes extractBytesFromInterlock(final InterlockSigningMetadata metadata) {
+    final InterlockSessionFactoryImpl interlockSessionFactory =
+        InterlockSessionFactoryProvider.newInstance(vertx, metadata.getKnownServersFile());
+    try (final InterlockSession session =
+        interlockSessionFactory.newSession(
```
jframe, line 118: Think I remember that there can only be a single session existing at once, so this probably fail when we load multiple keys. As the SignerLoader processes the loading of keys in parallel.
Also isn't establishing a session expensive? So we only want to do this once for the all the interlock keys we load.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/ManualInterlockSigningAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+@Disabled("Requires access to Interlock on USB Armory II")
+public class ManualInterlockSigningAcceptanceTest extends SigningAcceptanceTestBase {
+  private static final MetadataFileHelpers metadataFileHelpers = new MetadataFileHelpers();
+
+  // following keys are expected at /bls/key1.txt...key11.txt in USB Armory
+  private final List<BLSSecretKey> blsSecretKeys =
+      List.of(
```
rain-on, line 43: can we make a list of hex strings, then only have the BLSSecretKey.fromBytes(Bytes32.fromHexString(val)) just written once inside a loop?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/ManualInterlockSigningAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+              Bytes32.fromHexString(
+                  "0bdeebbad8f9b240192635c42f40f2d02ee524c5a3fe8cda53fb4897b08c66fe")),
+          BLSSecretKey.fromBytes(
+              Bytes32.fromHexString(
+                  "5e8d5667ce78982a07242739ab03dc63c91e830c80a5b6adca777e3f216a405d")));
+
+  @BeforeEach
+  void init() {
+    final Path knownServersFile = testDirectory.resolve("interlockKnownServer.txt");
+    for (int i = 1; i <= 11; i++) {
```
rain-on, line 81: what is the magic 11? Is that the number of keys we have? Can we remove the magic number and put in what it represents?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/ManualInterlockSigningAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+      final Path configFile = testDirectory.resolve("interlock" + i + ".yaml");
+      metadataFileHelpers.createInterlockYamlFileAt(
+          configFile, knownServersFile, Path.of("/bls/key" + i + ".txt"), KeyType.BLS);
```
rain-on, line 84: Can we really use "Path" to represent the path in interlock? I.e. a Path _could_ have "\" depending on the platform - should it just be a string?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/ManualInterlockSigningAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // secp key
+    final Path configFile = testDirectory.resolve("interlock_secp.yaml");
+    metadataFileHelpers.createInterlockYamlFileAt(
+        configFile, knownServersFile, Path.of("/secp/key1.txt"), KeyType.SECP256K1);
```
rain-on, line 90: Path.of really should be Path.of("secp", "key1.txt") to be platform agnostic. (the forward slash won't work on windows)

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/ManualInterlockSigningAcceptanceTest.java
rain-on: Also - this is overkill for an AT - the AT just needs to prove that we can load a single key from the interlock (ok, maybe 2) and sign with them - no need to do 11.
IntegrationTests etc. should prove the underlying capability - the AT is just required to prove we've wired it all together correct at the application...

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/ManualInterlockSigningAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+    blsSecretKeys.forEach(
+        blsSecretKey -> {
+          final Response response;
+          try {
+            response = signer.eth2Sign(blsSecretKey.toPublicKey().toString(), request);
+          } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+          }
+
+          final Bytes signature = verifyAndGetSignatureResponse(response);
```
rain-on, line 109: I ALMOST don't even need to see data being signed - having the keys turn up in the publicKeys API might be enough for me ...

core/src/main/java/tech/pegasys/web3signer/core/Eth1Runner.java
```diff
@@ -89,6 +89,7 @@ private ArtifactSignerProvider loadSigners(final Config config, final Vertx vert
             hashicorpConnectionFactory,
             config.getKeyConfigPath(),
             azureFactory,
+            vertx,
```
rain-on, line 92: this should probably be abstracted as a yubikeyHsmConnectionFactory, rather than passing down the raw vertx - similar t how we handle Hashicorp.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
```diff
@@ -54,6 +55,8 @@
           .collect(Collectors.toSet());
     } catch (final IOException e) {
       LOG.error("Unable to access the supplied key directory", e);
+    } finally {
+      InterlockKeyProvider.INSTANCE.closeAllSessions();
```
rain-on, line 59: do not love the singleton, can we pass an InterlockKeyProvider into the AbstractArtifactSignerFactory at startup (instead of vertx)? Then there's only one of them.

core/src/test/java/tech/pegasys/web3signer/core/multikey/metadata/InterlockSigningMetadataParsingTest.java
```diff
@@ -0,0 +1,87 @@
...
+    assertThat(signingMetadata).isInstanceOf(InterlockSigningMetadata.class);
+
+    final InterlockSigningMetadata metadata = (InterlockSigningMetadata) signingMetadata;
+    assertThat(metadata.getInterlockUrl()).isEqualTo(URI.create("https://10.0.0.1"));
+    assertThat(metadata.getKnownServersFile())
+        .isEqualTo(Path.of("./interlockKnownServersFile.txt"));
+    assertThat(metadata.getVolume()).isEqualTo("armory");
+    assertThat(metadata.getPassword()).isEqualTo("usbarmory");
+    assertThat(metadata.getKeyPath()).isEqualTo(Path.of("/bls/key1.txt"));
+    assertThat(metadata.getKeyType()).isEqualTo(KeyType.BLS);
```
rain-on, line 51: where do defaults come from? Do we need to test them somewhere?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/InterlockKeysAcceptanceTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+@Disabled("Requires physical access to Interlock on USB Armory II")
+public class InterlockKeysAcceptanceTest extends KeyIdentifiersAcceptanceTestBase {
+  private static final MetadataFileHelpers metadataFileHelpers = new MetadataFileHelpers();
+
+  // following keys are expected to be pre-loaded in USB Armory/Interlock under path
+  // /bls/key<1..n>.txt
+  private static final List<String> PRE_LOADED_KEYS =
```
rain-on, line 36: assuming these are the private keys?

usmansaleem: yes they are, will update variable name to represent such.

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -172,6 +175,9 @@ private ArtifactSignerProvider loadSigners(
             "yaml",
             new YamlSignerParser(List.of(artifactSignerFactory))));
 
+    // required as we are manually maintaining them in cache
+    interlockKeyProvider.closeAllSessions();
```
rain-on, line 179: put inside a finally?

core/src/main/java/tech/pegasys/web3signer/core/FilecoinRunner.java
```diff
@@ -124,6 +128,9 @@ private ArtifactSignerProvider loadSigners(
             "yaml",
             new YamlSignerParser(List.of(blsArtifactSignerFactory, secpArtifactSignerFactory)));
 
+    // required as we are manually maintaining them in cache
+    interlockKeyProvider.closeAllSessions();
```
rain-on, line 132: put inside finally

rain-on (approved):
just the finally I think - but otherwise, looks really good

rain-on (approved):

PR #260, opened 2020-10-15
Repair log levels and how they are selected

rain-on:


commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -91,7 +91,7 @@
       paramLabel = "<LOG VERBOSITY LEVEL>",
       description =
           "Logging verbosity levels: OFF, FATAL, WARN, INFO, DEBUG, TRACE, ALL (default: INFO)")
-  private final Level logLevel = Level.INFO;
+  private final Level logLevel = null;
```
usmansaleem, line 94: are we not defaulting to INFO anymore?

rain-on: The defaulting is managed in the log4j2.xml file in under app

jframe (approved):
LGTM

PR #265, opened 2020-10-22
Move jackson serialisers into common:

rain-on:


common/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+      'Specification-Version': project.version,
+      'Implementation-Title': archiveBaseName,
+      'Implementation-Version': calculateVersion()
+      )
+  }
+}
+
+dependencies {
+
+  implementation 'com.google.guava:guava'
```
jframe, line 19: There are some dependencies here that I don't think we need for the Jackson serialisers guava, commons-io, commons-lang, web3j, jackson-cbor, jackson-yaml but perhaps these are transitive dependencies we need.

common/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+  runtimeOnly 'org.bouncycastle:bcpkix-jdk15on'
+
+  runtimeOnly 'tech.pegasys:jblst'
+
+  testImplementation 'org.junit.jupiter:junit-jupiter-api'
+  testImplementation 'org.junit.jupiter:junit-jupiter-params'
+  testImplementation 'org.assertj:assertj-core'
+  testImplementation 'org.mockito:mockito-junit-jupiter'
+
+  testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
```
jframe, line 42: Since there are no tests, shouldn't need these either.
Probably should write some tests for the serialisers though I guess.

rain-on: didn't exist before ... are we happy for them to be tested at the "layer up" where they are used in modules?

common/build.gradle
rain-on: cleaned up.

jframe (approved):
LGTM

PR #268, opened 2020-10-23
remove fork_info as required param for deposit data signing requests

jframe:
The signing request for deposit data doesn't require the forkInfo to compute the domain. This removes this from the request so that dummy data doesn't need to be provided.


usmansaleem (approved):
LGTM

PR #271, opened 2020-10-26
Eth2 slashing protection signing low watermark

jframe:
Adds a low watermark for signing blocks and attestations which below we will not sign. This is so we don't sign when we don't have data for the slot or epoch.

Required as part of implementing the slashing database interchange format.

### DIFF
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
@@ -105,6 +105,26 @@ public class DbSlashingProtection implements SlashingProtection {
       final UInt64 sourceEpoch,
       final UInt64 targetEpoch,
       final int validatorId) {
+    final Optional<UInt64> minimumSourceEpoch =
+        signedAttestationsDao.minimumSourceEpoch(h, validatorId);
+    if (minimumSourceEpoch.filter(minEpoch -> sourceEpoch.compareTo(minEpoch) < 0).isPresent()) {
+      LOG.warn(
+          "Attestation source epoch {} is below minimum existing attestation source epoch {}",
+          sourceEpoch,
+          minimumSourceEpoch);
+      return false;
+    }
+
+    final Optional<UInt64> minimumTargetEpoch =
+        signedAttestationsDao.minimumTargetEpoch(h, validatorId);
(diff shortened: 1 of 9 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -126,6 +126,26 @@ private boolean checkAndInsertAttestation(
...
+      LOG.warn(
+          "Attestation source epoch {} is below minimum existing attestation source epoch {}",
+          sourceEpoch,
+          minimumSourceEpoch.get());
+      return false;
+    }
+
+    final Optional<UInt64> minimumTargetEpoch =
+        signedAttestationsDao.minimumTargetEpoch(h, validatorId);
+    if (minimumTargetEpoch.filter(minEpoch -> targetEpoch.compareTo(minEpoch) <= 0).isPresent()) {
```
usmansaleem, line 141: reconfirming ... `targetEpoch <= minEpoch`

usmansaleem (approved):
LGTM

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
jframe: Yes, for sourceEpoch it's < minEpoch but for targetEpoch it's <= minEpoch. Not what I expected either but that's what in the spec listed here https://hackmd.io/@sproul/Bk0Y0qdGD#Semantics

rain-on: In my head I saw:
if(minimumTargetEpoch.map(minEpoch -> targetEpoch.compareTo(minEpoch)).orElse(false)) {
...
}
But it is literally semantics

rain-on: and tbh - a function something like "epochLessThanThreshold(threshold, value)" would make this a BIT easier to read ... mainly because I hate the compareTo function.

rain-on (approved):
comments are more for flavour 👍

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
jframe: changed to use map

PR #279, opened 2020-10-28
cleanup openapi debug logging

jframe:
This reduces the openAPI validation logging to info level. As when debug logging is enabled for com.networknt.schema there are large amounts of logging per request ~100 lines which isn't helpful for debugging.

If the debug level for com.networknt.schema is required this can be overridden with a custom log4j file.


usmansaleem (approved):
LGTM

rain-on (approved):

PR #284, opened 2020-10-30
Update signer library to 1.0.13

usmansaleem:
-- Allows to read BLS keystores (EIP-2335) without UUID and Path fields


jframe (approved):
LGTM

PR #285, opened 2020-10-30
Update Filecoin RPC to be compatible with Lotus RemoteWallet

jframe:
Update the Filecoin RPC so that it works when called using the Lotus RemoteWallet as a signer.

* Update the path to be /rpc/v0 instead of /rpc/v1/filecoin as this is hard-coded in Lotus
* Add support for the MsgMeta field in the Wallet.Sign
* Verify that the Wallet.Sign is valid by computing the CID(MsgMeta.extra) == data
* Remove the unused Wallet.SignMessage RPC


usmansaleem (approved):
LGTM

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/FcCidEncoder.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.tuweni.bytes.Bytes;
+
+public class FcCidEncoder {
+
+  private static final BigInteger FC_HASHING_ALGO_CODE = BigInteger.valueOf(0xb220);
+  private static final byte CID_VERSION = (byte) 1;
+  private static final byte DagCBOR_CODEC_ID = (byte) 113;
+
+  public Bytes createCid(final Bytes cborEncodedBytes) {
```
rain-on, line 28: do we care if its cborencodedbytes - or is simply data?

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/FcCidEncoder.java
```diff
@@ -0,0 +1,49 @@
...
+import tech.pegasys.web3signer.core.util.Blake2b;
+import tech.pegasys.web3signer.core.util.ByteUtils;
+
+import java.math.BigInteger;
+
+import org.apache.tuweni.bytes.Bytes;
+
+public class FcCidEncoder {
+
+  private static final BigInteger FC_HASHING_ALGO_CODE = BigInteger.valueOf(0xb220);
```
rain-on, line 24: a comment that these are hard-coded magic numbers from lotus?

rain-on (approved):

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/FcCidEncoder.java
jframe: Replacing all these constants and using the Java CID library. That should make all of this clearer.

PR #288, opened 2020-10-31
Fix eth2 slashing deposit openAPI schema had wrong scope level of "type" element

### LINKED ISSUE #287: [OpenAPI Spec][Eth2] DepositSigning object contains unused element and scoping typo
The openapi spec for eth2 signing at `core/src/main/resources/openapi/web3signer-eth2.yaml` seems to have the discriminator element `type` at the `deposit` element level instead of being at the `DepositSigning` level.
The signingRoot while having acceptance for deposit doesn't seems to be used as the input is directly used as a `DepositMessage` to compute the signingRoot based on `pubkey/withdr[...]/amount`.
Current:
```
    DepositSigning:
      type: object
      properties:
        deposit:
          type: object
          properties:
            type:
              type: "string"
            signingRoot:
              type: "string"
            pubkey:
              type: "string"
            withdrawal_credentials:
              type: "string"
            amount:
              type: "string"
      required:
        - deposit
```
Proposed changes:
```
    DepositSigning:
      type: object
      properties:
        type:
          type: "string"
        deposit:
          type: object
          properties:
            pubkey:
              type: "string"
            withdrawal_credentials:
              type: "string"
            amount:
              type: "string"
      required:
        - type
        - deposit
```

Magicking:
This fix #287

### DIFF
--- a/core/src/main/resources/openapi/web3signer-eth2.yaml
+++ b/core/src/main/resources/openapi/web3signer-eth2.yaml
@@ -169,20 +169,23 @@ components:
     DepositSigning:
       type: object
       properties:
+        type:
+          type: "string"
         deposit:
           type: object
           properties:
-            type:
-              type: "string"
-            signingRoot:
-              type: "string"
             pubkey:
               type: "string"
             withdrawal_credentials:
               type: "string"
             amount:
               type: "string"
+            signingRoot:
+              description: 'signing root for optional verification if field present'
+              type: "string"
+
       required:
+        - type
         - deposit
     RandaoRevealSigning:
       allOf:


core/src/main/resources/openapi/web3signer-eth2.yaml
```diff
@@ -169,20 +169,19 @@ components:
     DepositSigning:
       type: object
       properties:
+        type:
```
jframe, line 172: Thanks for fixing this, the type was at the wrong level.

core/src/main/resources/openapi/web3signer-eth2.yaml
```diff
@@ -169,20 +169,19 @@ components:
...
+          type: "string"
         deposit:
           type: object
           properties:
-            type:
-              type: "string"
-            signingRoot:
```
jframe, line 177: We were keeping signingRoot as a field for now on all signing request types for compatibility with the spec so this field is still allowed. And whilst we don't use the value directly we compute the signing root from the included data and if the signingRoot field is present we will validate that is matches computed signingRoot value.

Magicking: Indeed, I misread the implementation! Thanks for the clarification, I replaced that with a description reflecting your comment

jframe (approved):
LGTM

Issue #280, opened 2020-10-28
Integrate updated Yubihsm (via Signers) with Web3Signer


usmansaleem:
Integration updated YubiHSM with Web3Signer.

**Sample meta file**

Connector url can either be http or yhusb

http connector:
`http://localhost:12345`

usb connector:
`yhusb://` (will pick first YubiHSM automatically)
`yhusb://serial=111222333` (Using serial number)

additionalInitConfig: _The additional init config can be obtained from https://developers.yubico.com/YubiHSM2/Component_Reference/PKCS_11/. They have to be specified as a space separated string._

~~~
type: yubihsm
pkcs11ModulePath: /path/to/yubihsm-sdk/lib/pkcs11/yubihsm_pkcs11.so
connectorUrl: yhusb://
additionalInitConfig: debug libdebug timeout=5
authId: 1
password: test
opaqueDataId: 1
keyType: BLS
~~~

usmansaleem:
Notes:
Test data creation

`
  ./yubihsm-shell --connector=yhusb:// --authkey=1 --password=password \
   --action=put-opaque \
   --object-id=1 --domains=1 \
   --algorithm=opaque-data --capabilities=none \
   --informat=hex --in=000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f

   ./yubihsm-shell --connector=yhusb:// --authkey=1 --password=password \
   --action=put-opaque \
   --object-id=2 --domains=1 \
   --algorithm=opaque-data --capabilities=none \
   --informat=hex --in=73d51abbd89cb8196f0efb6892f94d68fccc2c35f0b84609e5f12c55dd85aba8

   ./yubihsm-shell --connector=yhusb:// --authkey=1 --password=password \
   --action=put-opaque \
   --object-id=3 --domains=1 \
   --algorithm=opaque-data --capabilities=none \
   --informat=hex --in=39722cbbf8b91a4b9045c5e6175f1001eac32f7fcd5eccda5c6e62fc4e638508

   ./yubihsm-shell --connector=yhusb:// --authkey=1 --password=password \
   --action=put-opaque \
   --object-id=4 --domains=1 \
   --algorithm=opaque-data --capabilities=none \
   --informat=hex --in=4c9326bb9805fa8f85882c12eae724cef0c62e118427f5948aefa5c428c43c93

   ./yubihsm-shell --connector=yhusb:// --authkey=1 --password=password \
   --action=put-opaque \
[...]

PR #289, opened 2020-11-02
YubiHSM (via Signers) library integration

usmansaleem:
Integrate YubiHSM (Pkcs11 module) to load the keys from.

- [x] YubiHSM signing meta data file
- [x] Cache module/session to avoid duplicate loading
- [x] (Manual) Acceptance Test to show keys are loaded successfully. 

See Issue #280


core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private Pkcs11Session getPkcs11Session(final YubiHsmSigningMetadata yubiHsmSigningMetadata) {
+    final Pkcs11Module pkcs11Module = getPkcs11Module(yubiHsmSigningMetadata);
+
+    final YubiHsmIdentifier identifier =
+        new YubiHsmIdentifier(
+            yubiHsmSigningMetadata.getAuthId(), yubiHsmSigningMetadata.getPassword());
+
+    return pkcs11SessionMap.computeIfAbsent(
```
jframe, line 44: Can a yubihsm have multiple accounts? If so then this won't prevent multiple sessions to the Yubihsm. Multiple devices can also be connected I think? So this would also fail in this case if the same auth details were used.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
```diff
@@ -0,0 +1,69 @@
...
+        identifier, i -> pkcs11Module.createSession(i.convertToPkcs11Pin()));
+  }
+
+  private Pkcs11Module getPkcs11Module(final YubiHsmSigningMetadata yubiHsmSigningMetadata) {
+    return pkcs11ModuleMap.computeIfAbsent(
+        yubiHsmSigningMetadata.getPkcs11ModulePath(),
+        s ->
```
jframe, line 51: nit: what s stand for? this represents the module path I think?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/YubiHsmKeysAcceptanceTest.java
```diff
@@ -0,0 +1,85 @@
...
+  @BeforeEach
+  void init() {
+
+    // create meta configuration files
+    for (short opaqueDataId = (short) 1;
+        opaqueDataId <= PRE_LOADED_PRIVATE_KEYS.size();
+        opaqueDataId++) {
+      final Path configFile = testDirectory.resolve("yubihsm_" + opaqueDataId + ".yaml");
+      METADATA_FILE_HELPERS.createYubihsmYamlFileAt(
+          configFile, MODULE_PATH, CONNECTOR_URL, AUTH_ID, PASSWORD, opaqueDataId, KeyType.BLS);
```
jframe, line 70: We should also test secp keys work with yubihsm

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/BlsArtifactSignerFactory.java
```diff
@@ -99,6 +101,15 @@ public ArtifactSigner create(final InterlockSigningMetadata interlockSigningMeta
     }
   }
 
+  @Override
+  public ArtifactSigner create(final YubiHsmSigningMetadata yubiHsmSigningMetadata) {
+    try (TimingContext ignored = privateKeyRetrievalTimer.labels("interlock").startTimer()) {
```
jframe, line 106: The metric label should be yubihsm not interlock

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/YubiHsmSigningMetadata.java
```diff
@@ -0,0 +1,78 @@
...
+    this.authId = authId;
+    this.password = password;
+    this.opaqueDataId = opaqueDataId;
+  }
+
+  public String getPkcs11ModulePath() {
+    return pkcs11ModulePath;
+  }
+
+  public String getInitConfig() {
```
jframe, line 56: Would rather this logic was in the YubiHsmOpaqueDataProvider so this could be kept a POJO

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
usmansaleem: This is a bit tricky. Only one instance of PKCS11 instance is loaded by Sun provider. Which means that one PKCS11 module can only be used with one YubiHSM device. If another YubiHSM device has to be used, then previous module has to be finalized and then re-loaded again with new initConfig. May be we have to document it as a limitation?

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
usmansaleem: updated.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/YubiHsmKeysAcceptanceTest.java
usmansaleem: added

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/BlsArtifactSignerFactory.java
usmansaleem: 👍  ... fixed.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/YubiHsmSigningMetadata.java
usmansaleem: agreed, updated as suggested.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
jframe: Think document it as a limitation

jframe (approved):

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/YubiHsmKeysAcceptanceTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.tuweni.bytes.Bytes32;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.web3j.crypto.Credentials;
+
+@Disabled("Requires physical access to YubiHSM device and SDK installed")
+public class YubiHsmKeysAcceptanceTest extends KeyIdentifiersAcceptanceTestBase {
+  // update following path with exact location of yubi pkcs11 module
+  private static final String MODULE_PATH =
```
rain-on, line 37: should this just an env var? I.e. don't leave behind any static strings?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/YubiHsmKeysAcceptanceTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    final Response response = signer.callApiPublicKeys(KeyType.SECP256K1);
+
+    final String[] expectedPublicKeys =
+        PRE_LOADED_SECP_PRIVATE_KEYS.stream().map(this::getPublicKey).toArray(String[]::new);
+    validateApiResponse(response, containsInAnyOrder(expectedPublicKeys));
+  }
+
+  private void createSecpConfigurationFiles() {
+    for (short opaqueDataId = 20;
```
rain-on, line 123: whats the magic 20?

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.tuweni.bytes.Bytes;
+
+// Note: PKCS11 Sun provider has a limitation that one loaded module can only connect to one YubiHSM
+// device
+// In order to connect to multiple YubiHSM devices (in parallel), different module
+// installations/path must be used.
+public class YubiHsmOpaqueDataProvider implements AutoCloseable {
+  // maintains a cache of YubiHSM PKCS11 module and session to avoid module/session duplication
+  private final Map<String, Pkcs11Module> pkcs11ModuleMap = new ConcurrentHashMap<>();
+  private final Map<String, Pkcs11Session> pkcs11SessionMap = new ConcurrentHashMap<>();
```
rain-on, line 35: wondering about the string key-type - its appears arbitrary - is it worth having an object which represents the key - then rely on Java hashcode to ensure equality etc?

rain-on: maybe even just use the metadata object? Alternatively, have a "set" rather than a map?

core/src/test/java/tech/pegasys/web3signer/core/multikey/metadata/YubiHsmSigningMetadataFileParsingTest.java
```diff
@@ -0,0 +1,74 @@
...
+    final SigningMetadata signingMetadata =
+        OBJECT_MAPPER.readValue(testFile, SigningMetadata.class);
+
+    assertThat(signingMetadata).isInstanceOf(YubiHsmSigningMetadata.class);
+
+    final YubiHsmSigningMetadata metadata = (YubiHsmSigningMetadata) signingMetadata;
+    assertThat(metadata.getPkcs11ModulePath()).isEqualTo("/path/to/yubihsm_pkcs11.so");
+    assertThat(metadata.getConnectorUrl()).isEqualTo("http://localhost:12345");
+    assertThat(metadata.getAdditionalInitConfig()).isEqualTo("debug libdebug");
+    assertThat(metadata.getAuthId()).isEqualTo((short) 1);
```
rain-on, line 61: love to not have the "short" cast  - can we use shorts.assertThat? (and elsewhere)

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/YubiHsmKeysAcceptanceTest.java
usmansaleem: Hmm ... env variable would work as well.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/YubiHsmKeysAcceptanceTest.java
usmansaleem: expected starting OpaqueObjId for SECP keys. I'll create a constant.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/yubihsm/YubiHsmOpaqueDataProvider.java
usmansaleem: The session String 'identifier' is a combination of `modulePath + connectorUrl + AuthId`. We can create a separate value object to represent the same (in fact I had one earlier, then removed it as it was easily constructed using a method within this class). It has to be a map because we have to look up already created sessions.

core/src/test/java/tech/pegasys/web3signer/core/multikey/metadata/YubiHsmSigningMetadataFileParsingTest.java
usmansaleem: hmm, let me have a look.

usmansaleem: I've moved the expected value as short constants. In Java integer literals has to be cast as short.

`private static final short EXPECTED_AUTH_ID = (short) 1;`

PR #291, opened 2020-11-05
Move Slashing rules to separate class

rain-on:


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedAttestationsDao.java
```diff
@@ -12,22 +12,44 @@
...
-      final Handle handle, final int validatorId, final UInt64 targetEpoch) {
+  public List<SignedAttestation> findAttestationsForEpochWithDifferentSigningRoot(
+      final Handle handle,
+      final int validatorId,
+      final UInt64 targetEpoch,
+      final Bytes signingRoot) {
+    return handle
+        .createQuery(
+            "SELECT validator_id, source_epoch, target_epoch, signing_root "
+                + "FROM signed_attestations WHERE (validator_id = ? AND target_epoch = ?) AND (signing_root <> ? OR signing_root IS NULL)")
```
usmansaleem, line 33: you may also make the signing root condition simpler with `...AND signing_root NOT IN (?)`. Hopefully these should pick both null and different signing roots.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/BlockValidator.java
```diff
@@ -0,0 +1,75 @@
...
+      final SignedBlocksDao signedBlocksDao) {
+    this.handle = handle;
+    this.signingRoot = signingRoot;
+    this.blockSlot = blockSlot;
+    this.validatorId = validatorId;
+    this.signedBlocksDao = signedBlocksDao;
+  }
+
+  public boolean directlyConflictsWithExistingEntry() {
+    return !signedBlocksDao
```
usmansaleem, line 50: nit: avoid ! and use isPresent ?

usmansaleem (approved):
LGTM

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/BlockValidator.java
rain-on: Alas - the sigendBlockDao function returns a list, not an optional - so isPresent isn't going to work.
Two options:
1. Rename function to be the positive case - eg avoidsConflict - but atm, all of these functions return true, if slashing should be prevented I think ...
2. change it to say "return list.size() != 0"

usmansaleem: oh didn't realize that it is returning a list. In that case, leave it as it is. It was just a nit-pick anyways.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
```diff
@@ -0,0 +1,147 @@
...
+          sourceEpoch,
+          targetEpoch,
+          publicKey);
+      return true;
+    }
+    return false;
+  }
+
+  public boolean isSurroundedByExistingAttestation() {
+    // check that no previous vote is surrounding the attestation
```
jframe, line 131: nit: this comment isn't needed now this is split out into it's own function

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
```diff
@@ -0,0 +1,147 @@
...
+          "Attestation target epoch {} is below minimum existing attestation target epoch {}",
+          targetEpoch,
+          minimumTargetEpoch.get());
+      return true;
+    }
+    return false;
+  }
+
+  public boolean surroundsExistingAttestation() {
+    // check that no previous vote is surrounded by attestation
```
jframe, line 113: nit: this comment isn't needed now this is split out into it's own function

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  public void insertToDatabase() {
+    final SignedAttestation signedAttestation =
+        new SignedAttestation(validatorId, sourceEpoch, targetEpoch, signingRoot);
+    signedAttestationsDao.insertAttestation(handle, signedAttestation);
+  }
+
+  public boolean directlyConflictsWithExistingEntry() {
+    return !signedAttestationsDao
```
jframe, line 80: Think we used to log a warn message in this case. It would be nice to know if we have received a attestation with a different signing root.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/BlockValidator.java
```diff
@@ -0,0 +1,75 @@
...
+        .findBlockForSlotWithDifferentSigningRoot(handle, validatorId, blockSlot, signingRoot)
+        .isEmpty();
```
jframe, line 52: Be good to log something if the there is a conflict

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtectionTest.java
```diff
@@ -77,34 +78,43 @@ public void setup() {
   }
 
   @Test
-  public void blockCanSignWhenNoMatchForPublicKey() {
-    when(signedBlocksDao.findExistingBlock(any(), anyInt(), any())).thenReturn(Optional.empty());
+  public void blockCanSignWhenNoMatchForValidator() {
```
jframe, line 81: These existing tests should probably moved into BlockValidatorTest and AttestationValidatorTest. Although we do need some tests showing that the order of rules is adhered too so do need something in the DbSlashingProtectionTest UTs.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlocksDao.java
```diff
@@ -12,21 +12,36 @@
...
 
+import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.units.bigints.UInt64;
 import org.jdbi.v3.core.Handle;
 
 public class SignedBlocksDao {
 
-  public Optional<SignedBlock> findExistingBlock(
-      final Handle handle, final int validatorId, final UInt64 slot) {
+  public List<SignedBlock> findBlockForSlotWithDifferentSigningRoot(
```
jframe, line 25: Looks like we don't need a list do just need to know if there is a block that exists? So could add a limit 1 and return Optional<Block>

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedAttestationsDao.java
```diff
@@ -12,22 +12,44 @@
...
 
+import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.units.bigints.UInt64;
 import org.jdbi.v3.core.Handle;
 
 public class SignedAttestationsDao {
 
-  public Optional<SignedAttestation> findExistingAttestation(
-      final Handle handle, final int validatorId, final UInt64 targetEpoch) {
+  public List<SignedAttestation> findAttestationsForEpochWithDifferentSigningRoot(
```
jframe, line 25: Similar to the BlocksDao looks like don't actually need all the entries so could do a LIMIT 1 and return Optional<SignedAttestation> using findFirst

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -108,104 +107,38 @@ public boolean maySignAttestation(
...
-        new SignedAttestation(validatorId, sourceEpoch, targetEpoch, signingRoot);
-    signedAttestationsDao.insertAttestation(h, signedAttestation);
-    return true;
+          if (attestationValidator.directlyConflictsWithExistingEntry()
+              || attestationValidator.isSurroundedByExistingAttestation()
+              || attestationValidator.surroundsExistingAttestation()) {
+            return false;
+          } else if (attestationValidator.existsInDatabase()) {
+            return true;
+          } else if (attestationValidator.hasSourceOlderThanWatermark()
```
jframe, line 135: Should the watermark checks come last? Before they were being done between the existing & conflict checks and the surround and surrounding checks.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/AttestationValidator.java
```diff
@@ -0,0 +1,158 @@
...
+
+  private final Handle h;
+  private final Bytes publicKey;
+  private final Bytes signingRoot;
+  private final UInt64 sourceEpoch;
+  private final UInt64 targetEpoch;
+  private final int validatorId;
+  private final SignedAttestationsDao signedAttestationsDao;
+
+  public static enum MATCHES_PRIOR {
```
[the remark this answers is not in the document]
rain-on: this class has been removed.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
rain-on: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
rain-on: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/BlockValidator.java
rain-on: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
rain-on: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedBlocksDao.java
rain-on: Yeah, I was wondering, but then decided that using a list is more generic, so potentially has other uses - no point in limiting the flexibility of this class.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/SignedAttestationsDao.java
rain-on: see above.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
rain-on: Yeah, they need to, as, even if a low-watermark exists, if I match an existing then I can sign it.
Effectively, in the past we did the "do I exist" as a precheck - and didn't care if others also existed.
What we've really done, is separate the concept of "do others exist" and "do I exist".

Issue #281, opened 2020-10-29
Eth2 signing should return json response


jframe:
Eth2 signing should return a json object response instead of plain text hexadecimal string we are returning at the moment. Teku will also need to be updated to support this potentially supporting both using content-type response.

The response should be matching was is EIP-3303 response:
< HTTP/1.1 200 OK
< content-type: application/json
< content-length: 210
< date: Wed, 30 Sep 2020 02:16:02 GMT
<
{"signature":"0xb5d0c01cef3b028e2c5f357c2d4b886f8e374d09dd660cd7dd14680d4f956778808b4d3b2ab743e890fc1a77ae62c3c90d613561b23c6adaeb5b0e288832304fddc08c7415080be73e556e8862a1b4d0f6aa8084e34a901544d5bb6aeed3a612"}

PR #292, opened 2020-11-05
JSON and Text as acceptable media type for eth2 sign operation

usmansaleem:
#281


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SigningAcceptanceTestBase.java
```diff
@@ -37,7 +37,18 @@ protected void setupSigner(final String mode, final Map<String, String> env) {
...
-    response.then().contentType(ContentType.TEXT).statusCode(200);
-    return Bytes.fromHexString(response.body().print());
+    return verifyAndGetSignatureResponse(response, ContentType.TEXT);
+  }
+
+  protected Bytes verifyAndGetSignatureResponse(
+      final Response response, final ContentType contentType) {
+    response.then().contentType(contentType).statusCode(200);
+    final String signature;
+    if (contentType == ContentType.JSON) {
```
jframe, line 47: Doesn't look like this is actually use for the json content type. Would be good to have an AT that shows we return json. The ableToSignWithoutSigningRootField only checks the contentType is set correctly, it doesn't check the response.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
```diff
@@ -147,8 +148,12 @@ public void failsIfSigningRootDoesNotMatchSigningData() throws JsonProcessingExc
     assertThat(response.getStatusCode()).isEqualTo(500);
   }
 
-  @Test
-  public void ableToSignWithoutSigningRootField() throws JsonProcessingException {
+  @ParameterizedTest
+  @EnumSource(
```
jframe, line 152: Would also be good to have this on a test that sends through the entire set of data, say the signDataWithKeyLoadedFromUnencryptedFile AT

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
```diff
@@ -213,7 +216,19 @@ private UInt64 toUInt64(final tech.pegasys.teku.infrastructure.unsigned.UInt64 u
...
+    final String acceptableContentType = routingContext.getAcceptableContentType();
+    final String contentType;
+    final String body;
+
+    LOG.trace("Acceptable Content Type {}", acceptableContentType);
+    if (Objects.equal("application/json", acceptableContentType)) {
+      contentType = JSON_UTF_8;
+      body = new JsonObject().put("signature", signature).encode();
+    } else {
+      contentType = TEXT_PLAIN_UTF_8;
```
jframe, line 228: Does this default to text/plain? If I send a request to Accept: */* will this return the plain text response?

usmansaleem: For `Accept` header with values `*/*` and `application/json` the `routingContext.getAcceptableContentType();`  returns `application/json`. If Accept header is any thing other than */*, application/json and text/<?>, the openapi router factory kicks in and returns a 403 I think (invalid media type).

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SigningAcceptanceTestBase.java
```diff
@@ -37,7 +37,18 @@ protected void setupSigner(final String mode, final Map<String, String> env) {
...
-    response.then().contentType(ContentType.TEXT).statusCode(200);
-    return Bytes.fromHexString(response.body().print());
+    return verifyAndGetSignatureResponse(response, ContentType.TEXT);
+  }
+
+  protected Bytes verifyAndGetSignatureResponse(
+      final Response response, final ContentType expectedContentType) {
+    response.then().contentType(expectedContentType).statusCode(200);
+    final String signature;
+    if (expectedContentType == ContentType.JSON) {
```
jframe, line 47: nit: a ternary would probably read nicer

jframe (approved):

PR #294, opened 2020-11-06
Update migrations to reflect changes

rain-on:


jframe:
These migrations fail for me with an existing DB from the 0.2 release with the error "ERROR: Validate failed: Migrations have failed validation"

Also don't think having these migrations are a good idea or needed. The DB did change on the master builds, but slashing protection was a hidden option and disabled so don't think that is an issue.

slashing-protection/src/main/resources/migrations/postgresql/V1__initial.sql
```diff
@@ -6,15 +6,15 @@ CREATE TABLE validators (
 CREATE TABLE signed_blocks (
     validator_id INTEGER NOT NULL,
     slot NUMERIC(20) NOT NULL,
-    signing_root BYTEA,
+    signing_root BYTEA NOT NULL,
```
jframe, line 9: We shouldn't change what V1 was as there will be existing databases using this schema

slashing-protection/src/main/resources/migrations/postgresql/V2__removeNullConstraint.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE signed_blocks
```
jframe, line 1: Similarly should just get rid of this, these changes are part of the V1 that was in the 0.2 release so think it's fine that they stay that way.

slashing-protection/src/main/resources/migrations/postgresql/schema
```diff
@@ -0,0 +1,177 @@
+--
```
jframe, line 1: Can we remove this. As it is I don't think it adds value. It contains a lot unnecessary changes such as the flyway database and ownership permissions.

slashing-protection/src/main/resources/migrations/postgresql/V1__initial.sql
rain-on: deal

slashing-protection/src/main/resources/migrations/postgresql/V2__removeNullConstraint.sql
rain-on: deal

slashing-protection/src/main/resources/migrations/postgresql/schema
rain-on: gone

jframe (approved):

slashing-protection/src/main/resources/migrations/postgresql/V2__removeUniqueConstraints.sql
```diff
@@ -0,0 +1,19 @@
...
+  WHERE conrelid = 'signed_blocks'::regclass
+  AND contype = 'u');
+
+  _signed_attestation_conn text := (
+  SELECT conname
+  FROM pg_constraint
+  WHERE conrelid = 'signed_attestations'::regclass
+  AND contype = 'u');
+BEGIN
+  EXECUTE 'ALTER TABLE signed_blocks DROP CONSTRAINT ' || _signed_block_conn;
```
jframe, line 16: Does removing these constraints affect performance at all? Essentially this was acting a primary key before

PR #296, opened 2020-11-10
Update Eth2 aggregate and proof

jframe:
Updates AggregateAndProof schema to match what Teku is sending on the external signing request.

### DIFF
--- a/core/src/main/resources/openapi/web3signer-eth2.yaml
+++ b/core/src/main/resources/openapi/web3signer-eth2.yaml
@@ -398,10 +398,10 @@ components:
     AggregateAndProof:
       type: object
       properties:
-        index:
+        aggregator_index:
           type: string
           format: uint64
-        attestation:
+        aggregate:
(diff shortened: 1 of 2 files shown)


core/src/main/resources/openapi/web3signer-eth2.yaml
```diff
@@ -398,10 +398,10 @@ components:
     AggregateAndProof:
       type: object
       properties:
-        index:
+        aggregator_index:
           type: string
           format: uint64
-        attestation:
+        aggregate:
```
rain-on, line 404: changed name, but not the "ref" content - is this expected? (i.e. still have attestation in the $ref

rain-on (approved):

core/src/main/resources/openapi/web3signer-eth2.yaml
jframe: Yes that's expected. Well I think it is based on the changes in Teku

PR #301, opened 2020-11-13
Use low watermark in Slashing logic

rain-on:
Rather than relying on the lowest-block/attestation to determine how far back Web3signer can safely sign, there is now a set of metadata which is set on first signing.

In future, this data will also be updated as part of the import logic.

### DIFF
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
@@ -69,8 +69,17 @@ public class DbSlashingProtection implements SlashingProtection {
       final Jdbi jdbi,
       final ValidatorsDao validatorsDao,
       final SignedBlocksDao signedBlocksDao,
-      final SignedAttestationsDao signedAttestationsDao) {
-    this(jdbi, validatorsDao, signedBlocksDao, signedAttestationsDao, new HashMap<>());
+      final SignedAttestationsDao signedAttestationsDao,
+      final MetadataDao metadataDao,
+      final LowWatermarkDao lowWatermarkDao) {
+    this(
+        jdbi,
+        validatorsDao,
+        signedBlocksDao,
+        signedAttestationsDao,
+        metadataDao,
+        lowWatermarkDao,
+        new HashMap<>());
   }
 
   public DbSlashingProtection(
@@ -145,6 +154,7 @@ public class DbSlashingProtection implements SlashingProtection {
                   validatorId,
                   signedAttestationsDao,
                   lowWatermarkDao);
+
           final GenesisValidatorRootValidator gvrValidator =
               new GenesisValidatorRootValidator(handle, metadataDao);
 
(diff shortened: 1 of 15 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -68,8 +70,16 @@ public DbSlashingProtection(
...
+      final MetadataDao metadataDao,
+      final LowWatermarkDao lowWatermarkDao) {
+    this(
+        jdbi,
+        validatorsDao,
+        signedBlocksDao,
+        signedAttestationsDao,
+        metadataDao,
+        lowWatermarkDao,
+        new HashMap<>());
```
usmansaleem, line 82: if the map is not meant to be modified, you can pass `Collections.emptyMap` instead.

usmansaleem (approved):
LGTM - one minor comment around new HashMap

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
```diff
@@ -65,16 +74,21 @@ public boolean sourceGreaterThanTargetEpoch() {
...
   }
 
-  public void insertToDatabase() {
+  public void persist() {
     final SignedAttestation signedAttestation =
         new SignedAttestation(validatorId, sourceEpoch, targetEpoch, signingRoot);
     signedAttestationsDao.insertAttestation(handle, signedAttestation);
+
+    // update the watermark if is otherwise blank
+    if (watermark.isEmpty() || (watermark.get().getSourceEpoch() == null)) {
```
jframe, line 89: Possibly also check that targetEpoch is null. That should be set in pairs but the db allows for them to be set independently.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/BlockValidator.java
```diff
@@ -32,18 +34,24 @@
...
-      final SignedBlocksDao signedBlocksDao) {
+      final SignedBlocksDao signedBlocksDao,
+      final LowWatermarkDao lowWatermarkDao) {
     this.handle = handle;
     this.signingRoot = signingRoot;
     this.blockSlot = blockSlot;
     this.validatorId = validatorId;
     this.signedBlocksDao = signedBlocksDao;
+    this.lowWatermarkDao = lowWatermarkDao;
+    watermark = lowWatermarkDao.findLowWatermarkForValidator(handle, validatorId);
```
jframe, line 54: Don't like this happening in the constructor. perhaps use a memoize so it can be done when needed or just pass it in?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
rain-on: it does get changed, so can't use an emptyMap.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/AttestationValidator.java
rain-on: it looks like we can create a check constraint in the db ... but not sure how

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/BlockValidator.java
rain-on: changed to memoizing

jframe (approved):

PR #303, opened 2020-11-13
Slashing rules enforced - return status code 412

usmansaleem:
Return status code 412 instead of 403 as 403 is often interpreted by reverse proxy (such as nginx) as authorization failure.


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/slashing/SlashingAcceptanceTest.java
```diff
@@ -39,6 +39,7 @@
   public static final String DB_USERNAME = "postgres";
   public static final String DB_PASSWORD = "postgres";
   protected final BLSKeyPair keyPair = BLSKeyPair.random(0);
+  static final int SLASHING_PROTECTION_ENFORCED = 412;
```
jframe, line 42: can this be private?

jframe, line 42: Bit unsure if 412 is the correct status code here. It intended to be used with the if condition headers. Are we going to mask errors if this is used like we could have if the 403 was used?

jframe (approved):

PR #304, opened 2020-11-16
Fix Eth2 slashing protections metrics category not working on CLI

jframe:
This allows the ETH2_SLASHING_PROTECTION to be specified on the command line.

* Support for ETH2_SLASHING_PROTECTION as a value for the metrics-category CLI option
* Adds acceptance test for the Eth2 slashing protection
* Update the description for the Eth prevented signings metric


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfigurationBuilder.java
```diff
@@ -70,6 +71,11 @@ public SignerConfigurationBuilder withMetricsHostAllowList(final List<String> al
     return this;
   }
 
+  public SignerConfigurationBuilder withMetricsCategories(final List<String> metricsCategories) {
```
rain-on, line 74: is it worth making this an ellipsis - then do the conversion to a list in here (Saves a List.of everytime we use it)

rain-on (approved):

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfigurationBuilder.java
jframe: done

usmansaleem (approved):
LGTM

PR #318, opened 2020-11-24
Fix metadata table conflict inserting GVR config

jframe:
The genesis validators root is insert on first received request. If multiple requests are received at the same time on an empty database some of those requests could fail due to a constraint ensuring that only one GVR should be in the metadata table.

This change will retry the check and insert logic for the GVR if we get an insert error as it is quite likely that just retrying will fix the error.


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidator.java
```diff
@@ -15,33 +15,70 @@
...
+            metadataDao.insertGenesisValidatorsRoot(handle, genesisValidatorsRoot);
+          }
+          return isValidGvr;
+        });
+  }
+
+  private <V> V retry(final Callable<V> callable) {
+    try {
+      return callable.call();
+    } catch (Exception e) {
```
usmansaleem, line 65: nit: final

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbTransactionRetryer.java
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+    throw new IllegalStateException(
+        "Transaction max retries " + maxRetries + " reached. Not retrying transaction");
+  }
+
+  private void retrySleep() {
+    try {
+      final int jitter = random.nextInt(maxJitterMs);
+      Thread.sleep(retryMs + jitter);
```
usmansaleem, line 62: ouch ...

usmansaleem: may I suggest something like this instead? https://github.com/jhalterman/failsafe

usmansaleem (approved):
LGTM - suggestion around using Failsafe API which provides a better DSL to deal with retry and/or circuitbreaker pattern.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbTransactionRetryer.java
jframe: nice suggestion. done

PR #323, opened 2020-11-24
update release check

### LINKED ISSUE #322: Update openapi config script for new version system
since calculateVersion() changed and is used in core gradle to update openapi spec, I think it may have an impact on [checking if the spec is a release or not.](https://github.com/ConsenSys/web3signer/blob/master/.openapidoc/config.js#L75)
Updating .openapidoc/config.js is required

Also remove `0.2.0+30-g0d8024f` from the stable in versions.js as it was considered as a release.

NicolasMassart:
make the check match versioning strategy changes.

fixes #322

### DIFF
--- a/.openapidoc/config.js
+++ b/.openapidoc/config.js
@@ -73,8 +73,9 @@ function calculateSpecVersion(specFile) {
 }
 
 function isReleaseVersion(specVersion) {
-  // our main project's gradle's build calculateVersion puts -dev- for snapshot version
-  return !specVersion.includes("-dev-");
+  // our main project's gradle's build calculateVersion adds "+<new commits since stable>-<hash>"
+  // after the version for dev builds
+  return !specVersion.includes("+");
 }
 
 function destinationPath(specFile, suffix) {


.openapidoc/config.js
```diff
@@ -73,8 +73,9 @@ function calculateSpecVersion(specFile) {
 }
 
 function isReleaseVersion(specVersion) {
-  // our main project's gradle's build calculateVersion puts -dev- for snapshot version
-  return !specVersion.includes("-dev-");
+  // our main project's gradle's build calculateVersion adds "+<new commits since stable>-<hash>"
+  // after the version for dev builds
+  return !specVersion.includes("+");
```
usmansaleem, line 78: suggestion: `return !specVersion.includes("+") && !specVersion.includes("-dev-");`

This will allow to manage both both versioning while switch happens.

usmansaleem: nvm ... the main PR is already in.

usmansaleem (approved):

PR #324, opened 2020-11-24
remove files considered as release because of version changes

NicolasMassart:
See PR #323 for the fix on the generation side

THIS PR IS A DRAFT TO MAKE SURE IT'S NOT MERGED BEFORE #323


usmansaleem (approved):
LGTM ... might need to be updated.

PR #325, opened 2020-11-24
deposit signing AT

jframe:
Acceptance tests for the eth2 deposit signing. This uses data generated using for the eth2 deposit tool for the medalla and pyrmont chains.

Also fixed the openAPI spec for the genesis_fork_version


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2DepositSigningAcceptanceTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
+
+public class Eth2DepositSigningAcceptanceTest extends SigningAcceptanceTestBase {
+
+  private static final MetadataFileHelpers metadataFileHelpers = new MetadataFileHelpers();
+
+  @TestFactory
+  List<DynamicTest> signDepositData() {
+    loadKeystore("keystore-1");
```
rain-on, line 50: I assume this needs to be a keystore as you're aiming for an expected signature? Can you not just have a hard-coded priv-key here instead of needing resources?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2DepositSigningAcceptanceTest.java
```diff
@@ -0,0 +1,111 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.io.Resources;
+import io.restassured.response.Response;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
+
+public class Eth2DepositSigningAcceptanceTest extends SigningAcceptanceTestBase {
```
rain-on, line 44: do we have reasonable integration tests for this? (or is this easier? This is definitely more runtime expensive)

jframe: This is easier. The problem with integration tests is setting the Eth2SignForIdentifierHandler and constructing a RoutingContext. I did originally try that, but it much simpler to just use an AT in the end

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2DepositSigningAcceptanceTest.java
jframe: done

rain-on (approved):

PR #327, opened 2020-11-25
Hide scheme and domain from REST API doc samples panel

NicolasMassart:
Hide scheme and domain from REST API doc samples panel as we don't have a live Web3signer service running behind these doc samples to be tried by users.

The same change was made for Teku in ConsenSys/teku#3289


usmansaleem (approved):
LGTM

PR #329, opened 2020-11-25
Prevent duplicate formatversion in export

rain-on:


usmansaleem (approved):
LGTM. Will it has any impact on other clients importing this file?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/model/Metadata.java
```diff
@@ -31,6 +32,7 @@ public Metadata(
     this.genesisValidatorsRoot = genesisValidatorsRoot;
   }
 
+  @JsonIgnore
```
jframe, line 35: Looks like the getFormatVersionAsString is probably unnecessary and this this could be getter instead.

jframe: In that getFormatVersionAsString is only used in the tests. The value used in production code is coming from this function

rain-on: get version as string is what is used by the exporter - and it is required as the standard states that the version is a STRING not a numeric value :(

rain-on:
The problem is that we output the format version twice - once as a string "interchange_format_version" and again, as an integer field called format_version.
We ONLY want the string version.

jframe (approved):

PR #330, opened 2020-11-26
basic eth2 performance test

jframe:
Basic Eth2 signing test with slashing protection database. This send signing requests for blocks at 100/s for 60s. This is just to give an indication of our performance. We will need to expand on these tests to check different scenarios.

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
@@ -154,7 +154,7 @@ public class Signer extends FilecoinJsonRpcEndpoint {
   }
 
   public List<String> listPublicKeys(final KeyType keyType) {
-    return given().baseUri(getUrl()).get(publicKeysPath(keyType)).as(new TypeRef<>() {});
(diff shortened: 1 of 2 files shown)


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
```diff
@@ -152,6 +153,10 @@ public Response callApiPublicKeys(final KeyType keyType) {
     return given().baseUri(getUrl()).get(publicKeysPath(keyType));
   }
 
+  public List<String> listPublicKeys(final KeyType keyType) {
+    return given().baseUri(getUrl()).get(publicKeysPath(keyType)).as(new TypeRef<>() {});
```
rain-on, line 157: any reason this doesn't call into callApiPublicKeys?

rain-on (approved):
one minor.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
jframe: no I like that. done

PR #333, opened 2020-11-27
Azure secrets managed identity mode - cli options and configuration files

usmansaleem:
Provide support for using Azure managed identity when loading keys from Azure vault (via cli options and configuration files).


commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -89,22 +91,32 @@ private void validateArgs() {
...
 
-      if (azureKeyVaultParameters.getClientSecret() == null) {
-        missingAzureFields.add("--azure-client-secret");
+      if (azureKeyVaultParameters.getKeyVaultName() == null) {
+        missingAzureFields.add("--azure-vault-name");
       }
 
-      if (azureKeyVaultParameters.getClientlId() == null) {
-        missingAzureFields.add("--azure-client-id");
+      if (azureKeyVaultParameters.getAuthenticationMode() == null) {
```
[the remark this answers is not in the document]
usmansaleem: not in this case because we default to CLIENT_SECRET at tech.pegasys.web3signer.commandline.PicoCliAzureKeyVaultParameters#authenticationMode

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -89,22 +91,32 @@ private void validateArgs() {
...
+          missingAzureFields.add("--azure-client-id");
+        }
+
+        if (azureKeyVaultParameters.getTenantId() == null) {
+          missingAzureFields.add("--azure-tenant-id");
+        }
       }
-      if (missingAzureFields.size() != 0) {
+
+      //  no extra validation required for "managed identity". It can optionally use client-id.
```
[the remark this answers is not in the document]
usmansaleem: yes ... if user-defined-managed identity is setup, we need to specify clientId. If system-assigned managed identity is used, then clientId is not required to be set. i.e. User Defined Managed Identity has a clientId that needs to be specified.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AzureSecretSigningMetadata.java
```diff
@@ -25,21 +30,27 @@
...
 
   @JsonCreator
   public AzureSecretSigningMetadata(
       @JsonProperty("clientId") final String clientId,
       @JsonProperty("clientSecret") final String clientSecret,
       @JsonProperty("tenantId") final String tenantId,
-      @JsonProperty("vaultName") final String vaultName,
-      @JsonProperty("secretName") final String secretName,
+      @JsonProperty(value = "vaultName", required = true) final String vaultName,
+      @JsonProperty(value = "secretName", required = true) final String secretName,
```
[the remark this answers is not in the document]
usmansaleem: they were always required, but we didn't set them to required earlier (so it would have failed with a NPE or similar exception upon constructing the signer). Now I added the required flag which are required by both auth modes. Rest of the options are custom validated .. for CLIENT_SECRET mode, we need all other options, for Managed_Identity, we optionally need clinetId only.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AbstractArtifactSignerFactory.java
```diff
@@ -53,12 +54,19 @@ protected AbstractArtifactSignerFactory(
...
+          AzureKeyVault.createUsingClientSecretCredentials(
+              metadata.getClientId(),
+              metadata.getClientSecret(),
+              metadata.getTenantId(),
+              metadata.getVaultName());
+    } else {
+      azureVault =
+          AzureKeyVault.createUsingManagedIdentity(
+              Optional.ofNullable(metadata.getClientId()), metadata.getVaultName());
+    }
```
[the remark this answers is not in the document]
usmansaleem: oh the code in Eth2Runner is the "azure cli-mode" and the options are coming in from mixin class. This one here is coming from metadata class.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AzureSecretSigningMetadata.java
```diff
@@ -62,8 +73,36 @@ public String getSecretName() {
...
+    return authenticationMode;
+  }
+
   @Override
   public ArtifactSigner createSigner(final ArtifactSignerFactory factory) {
     return factory.create(this);
   }
+
+  @Override
+  public void validate() throws SigningMetadataException {
```
rain-on, line 86: nit: I can't put my finger on it (and it could be nothing) - but having a validate method on this class, implies that this class knows what's valid or not - but doesn't know _how_ its being used, so not sure how this logic can exist ... but equally sure I've argued in the opposite direction before.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/SigningMetadata.java
```diff
@@ -41,4 +41,8 @@ protected SigningMetadata(final KeyType keyType) {
   public KeyType getKeyType() {
     return keyType;
   }
+
+  public void validate() throws SigningMetadataException {
```
rain-on, line 45: the fact that no one else needs the validate method, feels like its the wrong abstraction

usmansaleem: addressed. Refactored to a separate class.

commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliAzureKeyVaultParameters.java
```diff
@@ -34,7 +43,9 @@
 
   @Option(
       names = {"--azure-client-id"},
-      description = "The ID used to authenticate with Azure key vault",
+      description =
+          "The ID used by client secret or user-assigned managed identity authentication mode to access Azure key vault. "
+              + "Optional for system-assigned managed identity.",
```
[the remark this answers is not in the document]
usmansaleem: clientId is required when mode is CLIENT_SECRET. If can be left empty (or assign empty value, via config file for instance) for MANAGED_IDENTITY (System-Generated). It can be set to a User-Assigned Managed Identity (hence Optional for Managed Identity). Feel free to suggest alternate wordings.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -87,24 +89,32 @@ private void validateArgs() {
 
     if (azureKeyVaultParameters.isAzureKeyVaultEnabled()) {
 
-      List<String> missingAzureFields = Lists.newArrayList();
+      final List<String> missingAzureFields = Lists.newArrayList();
 
-      if (azureKeyVaultParameters.getClientSecret() == null) {
-        missingAzureFields.add("--azure-client-secret");
+      if (azureKeyVaultParameters.getKeyVaultName() == null) {
```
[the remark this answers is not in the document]
usmansaleem: moving it out of here would require quite a bit of refactoring, hence I would leave it as it is for now.

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -196,12 +198,23 @@ private ArtifactSignerProvider loadSigners(
...
 
   final Collection<ArtifactSigner> loadAzureSigners() {
-    final AzureKeyVault keyVault =
-        new AzureKeyVault(
-            azureKeyVaultParameters.getClientlId(),
-            azureKeyVaultParameters.getClientSecret(),
-            azureKeyVaultParameters.getTenantId(),
-            azureKeyVaultParameters.getKeyVaultName());
+    final AzureKeyVault keyVault;
+    if (azureKeyVaultParameters.getAuthenticationMode() == AzureAuthenticationMode.CLIENT_SECRET) {
```
rain-on, line 202: wondering if this is the wrong place to be doing this conditional construction - i.e. IMHO this class just wants to say "get me a vault - here's some config" - this class shouldn't necessarily understand the config.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AbstractArtifactSignerFactory.java
```diff
@@ -53,12 +55,21 @@ protected AbstractArtifactSignerFactory(
...
 
   protected Bytes extractBytesFromVault(final AzureSecretSigningMetadata metadata) {
-    final AzureKeyVault azureVault =
-        new AzureKeyVault(
-            metadata.getClientId(),
-            metadata.getClientSecret(),
-            metadata.getTenantId(),
-            metadata.getVaultName());
+    final AzureKeyVault azureVault;
+    if (metadata.getAuthenticationMode() == AzureAuthenticationMode.CLIENT_SECRET) {
```
rain-on, line 59: is this a duplicate of the Eth2Runner work? Probably should live in 1 place.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/SigningMetadataValidator.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+
+/** SigningMetadata custom validation which may not be performed by parsing libraries */
+public class SigningMetadataValidator {
+
+  public static void validate(final SigningMetadata metadata) throws SigningMetadataException {
```
rain-on, line 25: don't really like this class :/ The public "validate" function throws away information, which has to be re-created using the "instanceof" - and I'm wondering why we only validate 1 config type, and no others?

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -50,6 +51,9 @@ public YamlSignerParser(final Collection<AbstractArtifactSignerFactory> signerFa
       final SigningMetadata metaDataInfo =
           OBJECT_MAPPER.readValue(metadataPath.toFile(), SigningMetadata.class);
 
+      // Custom validation of metadata parameters which is not enforced by Jackson
+      SigningMetadataValidator.validate(metaDataInfo);
```
rain-on, line 55: maybe move the validation work into the "createSigner"? At least then the function knows what type is expected.

core/src/test/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParserTest.java
```diff
@@ -271,16 +275,118 @@ void azureSecretMetadataInfoReturnsMetadata() throws IOException {
...
     assertThat(result).containsOnly(artifactSigner);
-    verify(blsArtifactSignerFactory).create(hasCorrectAzureMetadataArguments());
+    verify(blsArtifactSignerFactory)
+        .create(hasCorrectAzureMetadataArguments(AzureAuthenticationMode.CLIENT_SECRET));
   }
 
-  private AzureSecretSigningMetadata hasCorrectAzureMetadataArguments() {
+  @ParameterizedTest
+  @EnumSource(AzureAuthenticationMode.class)
+  void azureSecretMetadataWithAuthenticationModeReturnsMetadata(
```
rain-on, line 284: not quite sure if this is testing our code, or Jackson.
Its got wide ranging coverage (validators, factories etc.) :/ not sure.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/SigningMetadataValidator.java
usmansaleem: let me try to use jackson annotations to perform the validation ... that will get rid of this class.

core/src/main/java/tech/pegasys/web3signer/core/config/AzureAuthenticationMode.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.web3signer.core.config;
+
+public enum AzureAuthenticationMode {
+  CLIENT_SECRET,
+  MANAGED_IDENTITY
```
jframe, line 17: Wondering if MANAGED_IDENTITY should actually be two modes. Seems a bit magical that system or client managed identity is used based on the presence of the client id.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AbstractArtifactSignerFactory.java
```diff
@@ -53,12 +55,21 @@ protected AbstractArtifactSignerFactory(
...
+      azureVault =
+          AzureKeyVault.createUsingClientSecretCredentials(
+              metadata.getClientId(),
+              metadata.getClientSecret(),
+              metadata.getTenantId(),
+              metadata.getVaultName());
+    } else {
+      final Optional<String> clientId =
```
jframe, line 67: this could be removed if the AzureSecretSigningMetadata clientId was optional field in the Config object

core/src/test/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParserTest.java
usmansaleem: It primarily testing YamlSignerParser's parse method ... with specific metadata ... and mocks to ensure parse method has called appropriate logic after metadata is parsed!

jframe (approved):

rain-on (approved):

PR #335, opened 2020-11-29
Prevent low watermark from decreasing

rain-on:


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/LowWatermarkDao.java
```diff
@@ -37,12 +37,13 @@ public void updateEpochWatermarksFor(
       final UInt64 targetEpoch) {
     handle
         .createUpdate(
-            "INSERT INTO low_watermarks (validator_id, source_epoch, target_epoch) VALUES (:validator_id, :source_epoch, :target_epoch) "
+            "INSERT INTO low_watermarks (validator_id, source_epoch, target_epoch) VALUES (:validator_id, :srcEpoch, :tgtEpoch) "
                 + "ON CONFLICT (validator_id) "
-                + "DO UPDATE set source_epoch = :source_epoch , target_epoch = :target_epoch")
+                + "DO UPDATE set source_epoch=:srcEpoch, target_epoch=:tgtEpoch WHERE (((low_watermarks.source_epoch <= :srcEpoch) and (low_watermarks.target_epoch <= :tgtEpoch)) OR "
```
jframe, line 42: Is it possible for just one of the source or target epochs to be null? In that case what should happen?

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/LowWatermarkDaoTest.java
```diff
@@ -158,6 +158,69 @@ public void ensureBothWatermarksMustBeNonNull() {
...
+    Optional<SigningWatermark> watermark;
+    insertValidator(Bytes.of(100), 1);
+
+    lowWatermarkDao.updateEpochWatermarksFor(handle, 1, UInt64.valueOf(2), UInt64.valueOf(3));
+    watermark = lowWatermarkDao.findLowWatermarkForValidator(handle, 1);
+    assertThat(watermark).isNotEmpty();
+    assertThat(watermark.get().getSourceEpoch()).isEqualTo(UInt64.valueOf(2));
+    assertThat(watermark.get().getTargetEpoch()).isEqualTo(UInt64.valueOf(3));
+
+    lowWatermarkDao.updateEpochWatermarksFor(handle, 1, UInt64.valueOf(1), UInt64.valueOf(2));
```
jframe, line 172: If the target was greater existing target watermark but the source was lower than the existing source watermark what should happen? Think from the current logic nothing would be updated?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/LowWatermarkDao.java
rain-on: can't do it - we have an xnor constraint on the table

jframe (approved):

PR #336, opened 2020-11-29
Warn if no key config files exist

jframe:
Log a warning message if there are no key config files and guard against registering the empty list of validator keys as this fails with JDBI.


usmansaleem (approved):
LGTM

PR #338, opened 2020-11-30
Disable swagger-ui by default

usmansaleem:

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/Runner.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/Runner.java
@@ -170,6 +170,10 @@ public abstract class Runner implements Runnable {
   }
 
   private void registerSwaggerUIRoute(final Router router) throws IOException {
+    if (!config.isSwaggerUIEnabled()) {
+      return;
+    }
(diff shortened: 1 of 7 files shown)


core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -170,6 +170,10 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
   }
 
   private void registerSwaggerUIRoute(final Router router) throws IOException {
+    if (!config.isSwaggerUIEnabled()) {
```
jframe, line 173: nit: would rather not even call registerSwaggerUIRoute if isSwaggerUIEnabled is false. i.e. don't call this method in the run is swagger ui isn't enabled.

jframe (approved):
LGTM

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
usmansaleem: done, moved it out of method.

PR #340, opened 2020-11-30
Do not resign below watermark even if matching signing_root

rain-on:


jframe (approved):
LGTM - except for the ATs failing

PR #343, opened 2020-12-01
Use custom ForkJoinPool when processing configuration files

usmansaleem:
-- This fixes issue with Azure configuration files which can be processed in a batch of 10 due to a bug in Azure libraries

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
@@ -13,6 +13,7 @@
 package tech.pegasys.web3signer.core.multikey;
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.emptySet;
 
 import tech.pegasys.web3signer.core.multikey.metadata.parser.SignerParser;
 import tech.pegasys.web3signer.core.signing.ArtifactSigner;
@@ -21,11 +22,9 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
```diff
@@ -84,4 +118,15 @@ private static void renderException(final Throwable t, final String filename) {
         ExceptionUtils.getRootCauseMessage(t));
     LOG.debug(ExceptionUtils.getStackTrace(t));
   }
+
+  private static int numberOfThreads() {
+    int defaultNumberOfThreads = Runtime.getRuntime().availableProcessors() - 1;
+    // we only want to use 10 threads at max
+    if (defaultNumberOfThreads >= 10) {
```
rain-on, line 125: 10 should be a constant somewhere rather than a magic number.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
```diff
@@ -41,33 +46,62 @@
...
     LOG.info("Loading signer configuration files from {}", configsDirectory);
     final AtomicLong configFilesHandled = new AtomicLong();
+
+    final List<Path> configFilesList = getConfigFilesList(configsDirectory);
+    LOG.info("Processing {} files from configuration directory", configFilesList.size());
+
+    // use custom ForkJoin with limited number of threads if default is greater than 10
+    Set<ArtifactSigner> artifactSigners = Collections.emptySet();
+    ForkJoinPool forkJoinPool = null;
+    try {
```
rain-on, line 56: can this be split into subfunctions? It feels a bit busy, and might be easier to digest if smaller chunks.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
```diff
@@ -41,33 +46,62 @@
...
+      forkJoinPool = new ForkJoinPool(numberOfThreads());
+      artifactSigners =
+          forkJoinPool
+              .submit(
+                  () ->
+                      configFilesList.stream()
+                          .parallel()
+                          .filter(path -> matchesFileExtension(fileExtension, path))
+                          .flatMap(
+                              signerConfigFile -> {
```
rain-on, line 66: everything in this lambda can probably be broken out into separate function

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
```diff
@@ -41,33 +46,62 @@
 
     LOG.info("Loading signer configuration files from {}", configsDirectory);
     final AtomicLong configFilesHandled = new AtomicLong();
+
+    final List<Path> configFilesList = getConfigFilesList(configsDirectory);
```
rain-on, line 50: :/ don't love getting the filenames from disk, then iterating over them - but can see that the log info is potentially useful.

rain-on (approved):
minor comments only - tbh looks better in the IDE than the browser.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
usmansaleem: refactored to a constant.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
usmansaleem: refactored to sub methods.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
usmansaleem: refactored to sub methods.

core/src/main/java/tech/pegasys/web3signer/core/multikey/SignerLoader.java
usmansaleem: this allows to parallel-process files ... as Files.list.stream does not return parallel stream.

PR #345, opened 2020-12-02
Check db matches expected version

rain-on:


slashing-protection/referencetests/src/test/resources/log4j2.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Configuration status="INFO">
   <Properties>
-    <Property name="root.log.level">DEBUG</Property>
-    <Property name="dependency.log.level">WARN</Property>
+    <Property name="root.log.level">TRACE</Property>
```
usmansaleem, line 4: are these temporary for testing?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
```diff
@@ -22,16 +23,38 @@
 
 public class SlashingProtectionFactory {
 
+  public static final int EXPECTED_DATABASE_VERSION = 6;
```
usmansaleem, line 26: so, this would mean anytime we update our database scripts, we have to change the version here?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
```diff
@@ -22,16 +23,38 @@
...
+    verifyVersion(jdbi);
+
     return createSlashingProtection(jdbi);
   }
 
+  private static void verifyVersion(final Jdbi jdbi) {
+    final DatabaseVersionDao databaseVersionDao = new DatabaseVersionDao();
+
+    try {
+      final Integer version = jdbi.withHandle(databaseVersionDao::findDatabaseVersion);
```
usmansaleem, line 49: I am assuming this would cover the scenario if it is running against a database where table doesn't exist yet.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDaoTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @After
+  public void cleanup() {
+    handle.close();
+  }
+
+  @Test
+  public void migratedDatabaseReturnsValue() {
+    final int version = databaseVersionDao.findDatabaseVersion(handle);
+    assertThat(version).isEqualTo(6);
```
usmansaleem, line 52: won't these tests fail in future when version is not 6 anymore? probably use the constant from SlashingProtectionFactory?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
```diff
@@ -22,16 +23,38 @@
...
+      if (version.compareTo(EXPECTED_DATABASE_VERSION) != 0) {
```
jframe, line 50: any reason not to just use equals? if (version != EXPECTED_DATABASE_VERSION)...

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
jframe, line 49: nit: int instead of Integer, I know will box but then we can use == instead of compareTo

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
```diff
@@ -22,16 +23,38 @@
...
+        throw new IllegalStateException(errorMsg);
+      }
+    } catch (final IllegalStateException e) {
```
jframe, line 53: Is this IllegalStateException is the exception we are throwing on line 51? Looks like it could be.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
```diff
@@ -22,16 +23,38 @@
 
 public class SlashingProtectionFactory {
 
+  public static final int EXPECTED_DATABASE_VERSION = 6;
+  private static final String errorMsg =
+      String.format(
+          "Database does not have expected version (%s), please run migrations and try again.",
```
jframe, line 29: Can we also include the current db version in the message

slashing-protection/src/main/resources/migrations/postgresql/V6__add_db_version.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE database_version (
+    id INTEGER PRIMARY KEY,
```
jframe, line 2: Is there even a need for a primary key? Or are planning to use to enforce we only have a single entry like with the metadata table.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDaoTest.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
```
jframe, line 54: A test to show we get the correct value if we have multiple versions could be good

slashing-protection/src/main/resources/migrations/postgresql/V6__add_db_version.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE database_version (
```
jframe, line 1: Could just add this into the metadata table? Maybe that table should of been generic string, string to store all this slashing protection metadata.

slashing-protection/referencetests/src/test/resources/log4j2.xml
rain-on: oops - yes.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
rain-on: yes, and the database script would need to perform an UPDATE on the version table

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
rain-on: there are unit tests which prove that:
1. table is missing
2. table is empty
Both result in an exception being thrown

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDaoTest.java
rain-on: THAT is probably a good idea.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
rain-on: prefer not to box, and use == 👍

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
rain-on: No - coz we don't know if it was there or not :( (might be empty table)

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
rain-on: ah - this could be bad, refactored.

slashing-protection/src/main/resources/migrations/postgresql/V6__add_db_version.sql
rain-on: yes, it makes future updates easier.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDaoTest.java
rain-on: No. But have added a test to ensure version number matches number of migrations (for future protection).

slashing-protection/src/main/resources/migrations/postgresql/V6__add_db_version.sql
rain-on: makes life really hard due to the non-null constraint on GVR in the metadata table :/

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDao.java
```diff
@@ -0,0 +1,29 @@
...
+package tech.pegasys.web3signer.slashingprotection.dao;
+
+import org.jdbi.v3.core.Handle;
+
+public class DatabaseVersionDao {
+
+  public Integer findDatabaseVersion(final Handle handle) {
+    try {
+      return handle
+          .createQuery("SELECT version from database_version WHERE id = 1")
```
usmansaleem, line 22: just for clarification ... so there will always be one record and version field would keep updating? Rather than inserting a new record every time we update and pick the highest (or latest) version.

slashing-protection/src/main/resources/migrations/postgresql/V6__add_db_version.sql
```diff
@@ -0,0 +1,6 @@
...
+    id INTEGER PRIMARY KEY,
+    version INTEGER NOT NULL
+);
+-- Start at version 6, should have previously existed (but now represents migration index).
+INSERT INTO database_version (id, version) VALUES (1, 6);
```
usmansaleem, line 6: does it make sense to keep track of when was the last/latest version updated? if we keep separate records for each update we can insert an auto-date (or am I thinking too much into it).

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDaoTest.java
usmansaleem: is this change still pending?

usmansaleem (approved):
LGTM - minor comments, can go without them.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionFactory.java
tzapu: what i found in our migrations is that at times, you want to add an index which is not a breaking change and can allow the service to work with a version less than the latest, while migrating to the latest in parallel. not necessarily the case here as there is not much data in the web3signer db, so you most likely won't be creating any indexes in parallel, but thought it's worth mentioning

PR #347, opened 2020-12-03
Fix the external link for documentation on OpenAPI documentation

cdivitotawela:
Url https://doc.web3signer.pegasys.tech is no longer valid and updated to correct url https://docs.web3signer.consensys.net


usmansaleem (approved):
LGTM

PR #350, opened 2020-12-09
Fix errors in openAPI spec

jframe:
Fixes some errors in the openapi spec.

* The Attestation had fork as property at top level but we don’t send it at that level. It is included as part of ForkInfo. See https://github.com/ConsenSys/teku/blob/master/validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/ExternalSigner.java#L94
* ProposerSlashing slashing fields had the wrong field names it should match these https://github.com/ConsenSys/teku/blob/master/data/serializer/src/main/java/tech/pegasys/teku/api/schema/ProposerSlashing.java#L21. This is included in the BeaconBlockBody for block signings data/serializer/src/main/java/tech/pegasys/teku/api/schema/ProposerSlashing.java:21

### DIFF
--- a/core/src/main/resources/openapi/web3signer-eth2.yaml
+++ b/core/src/main/resources/openapi/web3signer-eth2.yaml
@@ -163,8 +163,6 @@ components:
           properties:
             attestation:
               $ref: "#/components/schemas/AttestationData"
-            fork:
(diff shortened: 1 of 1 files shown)


usmansaleem (approved):
LGTM

PR #359, opened 2021-01-07
Update dashboard for sharing

jframe:
Update the Grafana dashboard for import. This is the result of using share and export for external sharing. This was needed uploading on grafana.com and presumably for other users who don't have exact   same datasources we have.


usmansaleem (approved):
LGTM

PR #373, opened 2021-03-03
Support building from shallow clones

ajsutton:
Fix version detection when building from a shallow clone.  Same as the change in teku: https://github.com/ConsenSys/teku/pull/3687


usmansaleem (approved):
LGTM

Issue #375, opened 2021-03-07
reload API call should not clear previous loaded keys


usmansaleem:
Currently reload API call clears previously loaded keys, this may result in 0 key loaded if there is a network error or if key files location is unavailable. Reload API should add any new keys and keep old keys loaded for signing.

PR #376, opened 2021-03-07
[#375] Fix reload API logic

usmansaleem:
-- avoid cleaning up existing loaded keys


jframe (approved):
LGTM

PR #378, opened 2021-03-09
Rename integration base class

jframe:
Rename integration base class so it can be used for other integration tests


usmansaleem (approved):
LGTM

PR #385, opened 2021-03-24
fix pruning deleting too much data

jframe:
Fix issue with pruning where the delete would derive the watermark incorrectly in the delete below watermark query. This resulted too much data being deleted. Changed the delete to use a subquery also fixing the issue.

Also added some more logging so the progress of pruning can be seen.


usmansaleem (approved):
LGTM

PR #400, opened 2021-05-26
Upgrade gradle and various build plugins

usmansaleem:
- Upgrade Gradle to 7.x. Allows to compile against JDK 16
- Upgrade various plugins
- Use new dependency license generation plugin with custom html renderer.


buildSrc/src/main/java/tech/pegasys/internal/license/GroupedLicensesHtmlRenderer.java
```diff
@@ -0,0 +1,449 @@
...
+    }
+    output = new File(config.outputDir, fileName);
+
+    final Map<ModuleData, Set<Pair<String, String>>> moduleLicenses = buildModuleLicenses(data);
+    final Map<String, List<ModuleData>> licenseGroup = buildLicenseGroup(moduleLicenses);
+
+    writeReport(licenseGroup, moduleLicenses);
+  }
+
+  private Map<ModuleData, Set<Pair<String, String>>> buildModuleLicenses(final ProjectData data) {
```
jframe, line 119: Can you break this up into some smaller methods? There is a lot happening in this method

buildSrc/src/main/java/tech/pegasys/internal/license/GroupedLicensesHtmlRenderer.java
```diff
@@ -0,0 +1,449 @@
...
+    Map<ModuleData, Set<Pair<String, String>>> moduleLicences = new TreeMap<>();
```
jframe, line 120: nit: final and there few others I saw too

buildSrc/src/main/java/tech/pegasys/internal/license/GroupedLicensesHtmlRenderer.java
```diff
@@ -0,0 +1,449 @@
...
+      throw new UncheckedIOException(e);
+    }
+    return multipleLicenseMap;
+  }
+
+  private void writeReport(
+      final Map<String, List<ModuleData>> licenseGroup,
+      final Map<ModuleData, Set<Pair<String, String>>> moduleLicenses) {
+    final StringBuilder html = new StringBuilder();
+    html.append("<!doctype html>\n");
```
jframe, line 311: Don't particularly like html being directly in this class. Would be too much overhead to use a simple renderer like moustache?

buildSrc/src/main/java/tech/pegasys/internal/license/GroupedLicensesHtmlRenderer.java
usmansaleem: moved into its own repo.

buildSrc/src/main/java/tech/pegasys/internal/license/GroupedLicensesHtmlRenderer.java
usmansaleem: moved into its own repo.

buildSrc/src/main/java/tech/pegasys/internal/license/GroupedLicensesHtmlRenderer.java
usmansaleem: moved into its own repo.

build.gradle
```diff
@@ -12,20 +12,31 @@
...
+import tech.pegasys.internal.license.reporter.GroupedLicenseHtmlRenderer
 
 import java.text.SimpleDateFormat
 
+buildscript {
+  repositories {
+    maven { url "https://artifacts.consensys.net/public/maven/maven/" }
+  }
+  dependencies {
+    classpath 'tech.pegasys.internal.license.reporter:license-reporter:develop'
```
jframe, line 27: Is there a release version for the license reporter we can use?

usmansaleem: good catch, 1.0.0!

jframe (approved):
LGTM

PR #406, opened 2021-06-10
--network CLI option and Teku libraries upgrade

usmansaleem:
Update Teku libraries. As a result, introduce `--network` cli option which should match same as Teku at runtime.


build.gradle
```diff
@@ -116,7 +116,7 @@ allprojects {
     maven { url "https://hyperledger.jfrog.io/artifactory/besu-maven" }
     maven { url "https://hyperledger-org.bintray.com/besu-repo/" }
     maven { url "https://artifacts.consensys.net/public/maven/maven/" }
```
jframe, line 118: Do we need this maven repo for anything?

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -40,21 +44,35 @@
 
   public static final String COMMAND_NAME = "eth2";
 
-  @Spec CommandSpec spec;
+  @Spec CommandSpec commandSpec;
+
+  @CommandLine.Option(
+      names = {"-n", "--network"},
+      paramLabel = "<NETWORK>",
+      description = "Represents which network to use.",
```
jframe, line 52: Mention what values it could, the enum values? And should also mention that this could be a filename

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -40,21 +44,35 @@
...
   @Override
   public Runner createRunner() {
-    return new Eth2Runner(config, slashingProtectionParameters, azureKeyVaultParameters);
+    return new Eth2Runner(config, slashingProtectionParameters, azureKeyVaultParameters, eth2Spec);
   }
 
   @Override
   protected void validateArgs() {
+    final String networkConfigName =
+        Eth2Network.fromStringLenient(network).map(Eth2Network::configName).orElse(network);
```
jframe, line 68: Won't this change use mainnet if an invalid network is provided? That seems a bit dangerous to me, rather see with fail.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
```diff
@@ -258,4 +267,20 @@ private boolean isJsonCompatibleHeader(final MIMEHeader mimeHeader) {
     return Objects.equals("application/json", mimeComponent)
         || Objects.equals("*/*", mimeComponent);
   }
+
+  private static Bytes32 compute_fork_data_root(
```
jframe, line 271: finals on function args

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
```diff
@@ -258,4 +267,20 @@ private boolean isJsonCompatibleHeader(final MIMEHeader mimeHeader) {
...
+      Bytes4 current_version, Bytes32 genesis_validators_root) {
+    return new ForkData(current_version, genesis_validators_root).hashTreeRoot();
+  }
+
+  private static Bytes32 compute_domain(
```
jframe, line 276: finals on function args

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
jframe, line 271: this doesn't need to be static anymore

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
jframe, line 276: this doesn't need to be static anymore

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
```diff
@@ -258,4 +267,20 @@ private boolean isJsonCompatibleHeader(final MIMEHeader mimeHeader) {
...
+      Bytes4 domain_type, Bytes4 fork_version, Bytes32 genesis_validators_root) {
+    final Bytes32 fork_data_root = compute_fork_data_root(fork_version, genesis_validators_root);
+    return compute_domain(domain_type, fork_data_root);
+  }
+
+  private static Bytes32 compute_domain(final Bytes4 domain_type, final Bytes32 fork_data_root) {
```
jframe, line 282: this doesn't need to be static anymore

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
```diff
@@ -188,29 +196,30 @@ private Bytes computeSigningRoot(final Eth2SigningRequestBody body) {
...
-        return SigningRootUtil.signingRootForRandaoReveal(
+        return signingRootUtil.signingRootForRandaoReveal(
             body.getRandaoReveal().getEpoch(), body.getForkInfo().asInternalForkInfo());
       case VOLUNTARY_EXIT:
         checkArgument(body.getVoluntaryExit() != null, "voluntaryExit must be specified");
-        return SigningRootUtil.signingRootForSignVoluntaryExit(
+        return signingRootUtil.signingRootForSignVoluntaryExit(
             body.getVoluntaryExit().asInternalVoluntaryExit(),
             body.getForkInfo().asInternalForkInfo());
       case DEPOSIT:
```
jframe, line 225: Is there an alternative to using the deprecated compute_signing_root method for computing the deposit signing root?

gradle/versions.gradle
```diff
@@ -75,15 +75,17 @@ dependencyManagement {
     dependency 'org.hyperledger.besu:plugin-api:21.1.3'
     dependency 'org.hyperledger.besu.internal:metrics-core:21.1.3'
 
-    dependencySet(group: 'tech.pegasys.teku.internal', version: '20.11.0') {
+    dependencySet(group: 'tech.pegasys.teku.internal', version: '21.6.0') {
       entry 'bls'
-      entry 'core'
-      entry 'datastructures'
+      entry('core') {
+        exclude 'teku:data'
```
jframe, line 81: Why does teku:data need to be excluded?

usmansaleem: This one is reported by core's pom however it is not published (as it is empty project - contains submodules though). The license report plugin was failing since it can't find the dependency. 

This should be fixed in the latest release of Teku. Let me remove it. (It was added while testing the earlier release 21.5.0).

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
usmansaleem: no, this will first attempt to use config name as defined in Eth2Network. If not available, then network will be used whatever user has specified (for example a URL or directory location). Teku uses `_` in few of the enums in Eth2Network. The fromStringLenient converts `-` to `_` to get the correct value. Copied verbatim from Teku.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
usmansaleem: This one is copied verbatim from Teku codebase as well. Let me update the description to mention URI or directory path. However, network config names (i.e. mainnet, pyrmont, prater etc.) are subject to change. https://docs.teku.consensys.net/en/latest/Reference/CLI/CLI-Syntax/#network

build.gradle
usmansaleem: signers libraries would come from `https://artifacts.consensys.net/public/maven/maven/`

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -40,21 +45,41 @@
 
   public static final String COMMAND_NAME = "eth2";
 
-  @Spec CommandSpec spec;
+  @Spec CommandSpec commandSpec;
+
+  @CommandLine.Option(
+      names = {"--network"},
+      paramLabel = "<NETWORK>",
+      description =
```
jframe, line 53: This can also be a yaml file path too can't it?

build.gradle
jframe: 👍

jframe (approved):
LGTM

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
usmansaleem: yes it can be. Updated description.

PR #419, opened 2021-08-17
Upgrade Teku libraries to 21.8.1.

usmansaleem:
-- Upgrade Teku libraries to 21.8.1. Added support for Altair upgrade on the Pyrmont testnet at epoch 61650.


jframe (approved):
LGTM

PR #420, opened 2021-08-19
Fix spelling of AggregateAndProofSigning in openapi spec file

usmansaleem:
Fix spelling of AggregateAndProofSigning in openapi spec file.


jframe (approved):
LGTM

PR #422, opened 2021-08-25
sign type block_v2 to support Altair and future forks

usmansaleem:
Block signing request for Altair fork uses different block than phase 0. Introduce new sign type BLOCK_V2 which will allow Altair and future forks block signing requests. Keep BLOCK for phase 0 backward compatibility.


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/SignerConfigurationBuilder.java
```diff
@@ -196,11 +197,16 @@ public SignerConfigurationBuilder withUseConfigFile(final boolean useConfigFile)
     return this;
   }
 
-  public SignerConfigurationBuilder withAltairForkEpoch(final long altairForkEpoch) {
+  public SignerConfigurationBuilder withAltairForkEpoch(final Optional<Long> altairForkEpoch) {
```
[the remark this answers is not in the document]
usmansaleem: I could've used `Optional.ofNullable()` .... but most of the time calling this method, we would use literal values like `0L` ... I wanted to force `Optional.of` .. or `Optional.empty` instead of passing `null`.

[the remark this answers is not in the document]
usmansaleem: It's ultimate usage is in DSL where we construct command line arguments to create a web3signer instance.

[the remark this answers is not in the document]
siladu: Doesn't the `Builder.with...` method imply it's an optional argument already? I would only expect to call this method if I had a non-null value to pass in.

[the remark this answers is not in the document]
usmansaleem: fair point :)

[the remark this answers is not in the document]
usmansaleem: suggestion incorporated!

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2AltairBlockSigningAcceptanceTest.java
```diff
@@ -77,7 +72,7 @@ void signAndVerifyBlockV2Signature(final SpecMilestone specMilestone) throws Exc
   @Test
   void signAndVerifyLegacyBlockSignature() throws Exception {
     final Eth2BlockSigningRequestUtil util = new Eth2BlockSigningRequestUtil(SpecMilestone.PHASE0);
-    setupSigner("eth2", null, "minimal", Optional.empty());
+    setupSigner("eth2", null, "minimal", null);
```
siladu, line 75: Might be overkill in this case, but could avoid passing the null into the Optional with another overloaded setupSigner method which doesn't call `withAltairForkEpoch`

usmansaleem: you are correct, its technically only for legacy block signing that we need to start web3signer without altair fork. Rest of signing types works with altair fork.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
```diff
@@ -59,7 +59,7 @@
   private static final BLSPublicKey publicKey = keyPair.getPublicKey();
 
   @ParameterizedTest
-  @EnumSource
+  @EnumSource(mode = EnumSource.Mode.EXCLUDE, names = "BLOCK_V2")
```
jframe, line 62: Why do we need to exclude BLOCK_V2? It would good to include the block v2 in these tests for completeness.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2AltairBlockSigningAcceptanceTest.java
jframe: Rather we didn't pass in null as well, even an optional altair block would be nicer. Might be nit picking though.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SigningRequestBody.java
```diff
@@ -44,6 +45,7 @@ public Eth2SigningRequestBody(
       @JsonProperty("signingRoot") final Bytes signingRoot,
       @JsonProperty("fork_info") final ForkInfo fork_info,
       @JsonProperty("block") final BeaconBlock block,
+      @JsonProperty("beacon_block") final BlockRequest blockRequest,
```
jframe, line 48: It is a bit confusing having block and beacon_block in the request. Not sure there is much we can do about now that Teku has implemented this though

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/json/BlockRequestDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class BlockRequestDeserializer extends JsonDeserializer<BlockRequest> {
+  private final ObjectMapper objectMapper;
+
+  public BlockRequestDeserializer(final ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  public BlockRequest deserialize(final JsonParser p, final DeserializationContext ctxt)
```
jframe, line 36: Perhaps a test? We do have AT though, so maybe that is enough

core/src/main/java/tech/pegasys/web3signer/core/service/http/SigningJsonProvider.java
```diff
@@ -0,0 +1,104 @@
...
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.apache.tuweni.units.bigints.UInt64;
+
+public class SigningJsonProvider {
```
jframe, line 54: The way we are using this doing `new SigningJsonProvider().getObjectMapper()` perhaps a factory would be would be better?

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/Eth2BlockSigningRequestUtil.java
```diff
@@ -0,0 +1,129 @@
...
+    tekuForkInfo = dataStructureUtil.randomForkInfo();
+    tekuFork = new Fork(tekuForkInfo.getFork());
+    forkInfo =
+        new tech.pegasys.web3signer.core.service.http.handlers.signing.eth2.ForkInfo(
+            tekuFork, tekuForkInfo.getGenesisValidatorsRoot());
+    beaconBlock = dataStructureUtil.randomBeaconBlock(0);
+    signingRoot = signingRootUtil.signingRootForSignBlock(beaconBlock, tekuForkInfo);
+  }
+
+  public Eth2SigningRequestBody createRandomBlockV2Request() {
```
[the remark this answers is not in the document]
usmansaleem: it is deterministic ... I think createCanned.... naming makes more sense.

jframe (approved):
LGTM

PR #426, opened 2021-09-07
Specify description and example for version in OpenAPI spec

### LINKED ISSUE #425: Specify `version` values for BLOCK_V2
Thanks for adding Altair support! Whilst implementing it for Lighthouse I had some trouble determining what the `version` value should be. Eventually I figured out it should be either:

- `PHASE0`
- `ALTAIR`

I tried modifying the docs myself to add the descriptions, but I couldn't figure out how to view the OpenAPI in my browser. I kept getting a `File not found /swagger-ui/web3signer.yaml` error.

usmansaleem:
- Specify description and example for version property of BlockRequestPhase0 and BlockRequestAltair

Fixes #425

### DIFF
--- a/core/src/main/resources/openapi/web3signer-eth2.yaml
+++ b/core/src/main/resources/openapi/web3signer-eth2.yaml
@@ -533,6 +533,8 @@ components:
       properties:
         version:
           type: string
+          description: 'version to identify block request type.'
+          example: PHASE0
         block:
           $ref: "#/components/schemas/BeaconBlock"
       required:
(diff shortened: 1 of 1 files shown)


core/src/main/resources/openapi/web3signer-eth2.yaml
```diff
@@ -533,6 +533,8 @@ components:
       properties:
         version:
           type: string
+          description: 'version to identify block request type.'
```
jframe, line 536: Perhaps define the version as an enum so all the possible values are defined somewhere?

jframe (approved):

PR #432, opened 2021-09-22
Upgrade signers to 1.0.19

siladu:
Includes vertx upgrade and azure key vault upgrade with breaking change to SECP256K1 curve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Next Version
+### Breaking Changes
+- Upgrade to signers 1.0.18 removes support for deprecated SECP256K1 curve.
+
 ## 21.8.1
 
 ### Features Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
 ## Next Version
+
+### Features Added
+- Upgrade to signers 1.0.19 allows empty password files to be read when creating a Signer.
+
 ### Breaking Changes
-- Upgrade to signers 1.0.18 removes support for deprecated SECP256K1 curve.
+- Upgrade to signers 1.0.19 removes support for deprecated SECP256K1 curve in Azure remote signing.
 
 ## 21.8.1
 
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
@@ -15,10 +15,10 @@ package tech.pegasys.web3signer.tests.signing;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Collections.singletonMap;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Next Version
+### Breaking Changes
+- Upgrade to signers 1.0.18 removes support for deprecated SECP256K1 curve.
```
usmansaleem, line 5: suggestion: Upgrade to signers 1.0.18 removes support for deprecated SECP256K1 curve in Azure remote signing.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -113,35 +97,39 @@ public void signDataWithKeyFromHashicorp(@TempDir Path keyConfigDirectory) {
...
     final Bytes signature = verifyAndGetSignatureResponse(response);
-    verifySignature(signature);
+    verifySignature(signature, publicKeyHex);
   }
 
-  void verifySignature(final Bytes signature) {
-    final BigInteger privateKey = new BigInteger(1, Bytes.fromHexString(PRIVATE_KEY).toArray());
-    final BigInteger expectedPublicKey = publicKeyFromPrivate(privateKey);
+  void verifySignature(final Bytes signature, final String publicKeyHex) {
+    ECPublicKey expectedPublicKey =
```
usmansaleem, line 125: nit: `final ECPublicKey expectedPublicKey ... ` (using final keyword)

gradle/versions.gradle
```diff
@@ -13,7 +13,7 @@
 
 dependencyManagement {
   dependencies {
-    dependency 'com.fasterxml.jackson.core:jackson-databind:2.11.2'
+    dependency 'com.fasterxml.jackson.core:jackson-databind:2.12.5'
     dependency 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.11.2'
```
usmansaleem, line 17: does `com.fasterxml.jackson.dataformat:jackson-dataformat-yaml` version required to be upgraded as well?

usmansaleem (approved):
LGTM - minor suggestions.

gradle/versions.gradle
jframe: Would be good to use a dependency set here too to keep the jackson versions in sync

jframe (approved):

gradle/versions.gradle
usmansaleem: @jframe  They are two different groups unfortunately.

siladu: AFAIK, the azure upgrade doesn't require dataformat-yaml to be upgraded (signers doesn't use it) Updating dataformat-yaml causes CommandlineParserTest failures which is starting to feel out of scope for this PR so I'm going to leave it as 2.11.2 unless there's any objections. I'd rather handle it in a 'upgrade web3signers deps' PR.

usmansaleem: 👍🏼 .. lets keep it as it is then.

usmansaleem (approved):
LGTM

PR #436, opened 2021-10-15
Prepare for release 21.10.0

siladu:
updated changelog with version


usmansaleem (commented):
I think we should add an entry in changelog:
- Upgrade Teku libraries to allow support for Altair fork on mainnet

Issue #439, opened 2021-10-30
Implementing a standard key manager API


joaquim-verges:
I'm looking to help out on [this proposal](https://hackmd.io/@dapplion/web3signer#) by @dapplion.

Working branch: https://github.com/joaquim-verges/web3signer/tree/keymanager

The branch will be broken down into individual PRs for each of the main tasks below.

## Tasks breakdown

* [x] Extract existing signing API yaml definitions into their own files - #447 
* [x] Add new key manager API yaml definition - #455 
* [x] Add command line flag to enable API - #448 
* [x] Conditionally register routes + handlers in the Eth2Runner
* [x] Implement List handler + tests - #456
  * [x] Json model parsing
  * [x] Return the pubkeys from the artifact signer provider
  * [x] (bonus) add plumbing required to pass derivation_data in the response
* [x] Implement Import handler + tests - #462 
   * [x] Json model parsing
   * [x] Save keystores + passwords into a yaml config file with the expected format 
   * [x] That new yaml config file should be read and loaded in Eth2Runner#createArtifactSignerProvider as long as its saved in yaml format in the same keystore path passed through CLI 
   * [x] Invoke reload, which should just be a call to artifactSignerProvider.load() 
   * [x] Return expected response, handle errors
   * [x] Write unit tests
* [x] Implement Delete handler + tests - #489
    * [x] Json model parsing
    * [x] Find and delete the yaml files corresponding to the pubkeys coming from the request
    * [x] export slashing data for removed/non_active keys
    * [x] Return expected response, handle errors
    * [ ] Write unit tests
* [ ] Implement Authentication
    * [ ] Add Vertx JWT library dependency
    * [ ] Generate JWT token if doesn't exist or read from file and load in memory
    * [ ] Persist JWT token if doesn't exist on first generation
    * [ ] Add security handler to the OpenAPI router to verify the correct auth token is passed and fail otherwise

## Context

The high level goal of this work is to increase client diversity by making it easy for individual stakers to spin up a minority client. 

Working backwards from that high level goal, a possible solution would be to: 

- Enable the management of minority clients via a platform like [DAppNode](https://dappnode.io/). Right now Prysm is the only client supported.
- To do this, the other clients (Teku, Lodestar, Lighthouse, Nimbus) would ideally have a mini frontend for their client that can be accessed via the DappNode dashboard. At a minimum, the frontend needs to allow the user to manage keystores.
[...]

PR #444, opened 2021-11-12
Split up OpenAPI spec into multiple files to prepare for new key manager api integration

joaquim-verges:
first preliminary step to start integrating the key manager API as discussed in #439 

This PR doesn't introduce any changes, just organizes the OpenAPI yaml file to make it easier and cleaner to add the new key manager endpoints, which will be in the follow up PR.


dapplion (approved):
:pray:

jframe:
Looks good, just want to check that the open api publishing still works.

joaquim-verges:
@jframe how do we check that the publishing works?

usmansaleem:
@joaquim-verges we have a gh-pages branch which ends up rendering OpenAPI doc at https://consensys.github.io/web3signer/ for each master/release commit. The circlecli (CI) executes `npm ci` on `https://github.com/ConsenSys/web3signer/tree/master/.openapidoc` which is a nodejs project which replaces version in the yaml and pushes them to gh-pages branch. 

We need to perform a quick sanity test whether the separated yaml files also gets published or not. Its been sometime since we have worked on this script (its kinda like "once it worked, leave it alone script"). Might need a bit of hack to figure out which files will be pushed without actually pushing remotely (while working on it in local dev mode).

jframe:
@joaquim-verges Was just testing the swagger-ui endpoint it's not able to display the openapi spec anymore in web3signer.
Screenshot: . 

Most likely because the referenced yaml files cannot be served. In the Runner we register a single file that we are expecting to contain the openapi doc https://github.com/ConsenSys/web3signer/blob/master/core/src/main/java/tech/pegasys/web3signer/core/Runner.java#L232

joaquim-verges:
@usmansaleem @jframe  thanks for checking/testing! That's what I was worried about, seems like not all tools support external `$ref` out of the box.

There is an option which would be to compile the split API spec into a single file, using something like https://github.com/mohsen1/multi-file-swagger-example

It does introduce an extra build step though, not sure how you guys feel about that.

The whole reason for doing this is purely for code organization (it does make the main API spec nice and tidy), but if it brings annoying problems it might not be worth it.

Let me know what you want to do here, either:

a. Just add the new endpoints to the main API spec (simplest solution, but will be a fairly large API spec with the 3 extra key manager endpoints + schemas)
b. Explore the single file compilation route / fix the tools in order to keep a clean API spec file

usmansaleem:
@joaquim-verges I am going to change the nodejs script that publishes to gh-pages .. so that it uses version in a directory instead of in filename. This will allow relative paths to work correctly. I'll also work on fixing the swagger-ui stuff. Your PR would need to wait till that work goes in. Should be in couple of days max.

joaquim-verges:
@usmansaleem that would be awesome, thank you! No rush, it doesn't block me. I can keep working on the key manager APIs in the meantime.

jframe:
@joaquim-verges It is seems tools don't seem to like multiple yaml files that much. Glad Vertx worked with the multiple files though.

We chatted a bit about an approach so I think we should be able to get something working though.

dapplion:
This repo hosts an OpenAPI specification in multiple files that are bundled and deployed successfully https://github.com/ethereum/beacon-APIs/blob/master/.github/workflows/release.yaml

usmansaleem:
https://github.com/ConsenSys/web3signer/pull/447 should address this PR (will close once merged).

usmansaleem:
Implemented by #447

PR #447, opened 2021-11-19
Openapi spec split

usmansaleem:
- Extract OpenAPI specs from classpath resources to disk
- Adjust relative ref paths to absolute paths for OpenApi3Router as it doesn't work well with relative paths
- Use relative paths for /swagger-ui by loading contents in memory
- Update .openapidoc nodejs project so that it publishes specs with version in directory name instead of filenames.

Merge #444 into this PR.


core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -224,22 +224,45 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
...
+            try {
+              file = com.google.common.io.Resources.toString(yaml, StandardCharsets.UTF_8);
+            } catch (final IOException e) {
+              throw new UncheckedIOException(e);
+            }
+
+            // adjust the ref (which works with OpenApi3Router) to relative that works with
+            // swagger-ui
+            final String adjustedFile =
+                StringUtils.replace(file, "/openapi/eth2/signing/schemas.yaml", "../schemas.yaml");
```
joaquim-verges, line 248: quite annoying to have to do this twice, both in the gradle task and here :( 

Once I add the new API endpoints I'll have another path to replace too on top of this one. Will be `/openapi/eth2/keymanager/schemas.yaml`. 

Would be great to make this replace logic generic so I don't have to touch this code once I add the new endpoints. Maybe with a regex? If starts with `/` -> replace everything before the last slash with `..` or something like that.

joaquim-verges (commented):
Thanks for doing this!

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
joaquim-verges: Also realized that this flattening of files into one folder means that all filenames must be unique right? Kind of a bummer. Anyway to preserve some of the folder hierarchy and still use relative paths? Like `../signing/schemas.yaml` and `../keymanager/schemas.yaml`

Would also make the string replace logic simpler that way. Would just be replace `/openapi/eth2` with `..` and done

usmansaleem: @joaquim-verges 

I am about to re-do this logic slightly ... apparently the Vertx's OpenApi3Router (which internally uses swagger-parser) is not fool-proof (i.e. fails to parse) when it deals with relative paths in `$ref` (such as sign.yaml referencing `../schema.yaml#....` where it sits two folder deep. It works if all files are on same level. On the other hand, tools like swagger-ui and redoc (which are javascript based) works fine with such relative paths.

I am adding a bit of logic to modify the relative `$ref` and `discriminator` entries to absolute paths for OpenApi3Router handling.

The end file structure that should work would look like:
~~~
openapi
├── eth1
│   └── web3signer.yaml
├── eth2
│   ├── signing
│   │   ├── paths
│   │   │   ├── public_keys.yaml
│   │   │   ├── reload.yaml
│   │   │   ├── sign.yaml
│   │   │   └── upcheck.yaml
│   │   └── schemas.yaml
│   └── web3signer.yaml
├── filecoin
│   └── web3signer.yaml
└── index.html
~~~

joaquim-verges: file structure looks good to me! Def had a hard time making relative paths work with Vertx, if there's a robust way to make it work that would be the cleanest for sure.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/SwaggerUIAcceptanceTest.java
```diff
@@ -78,7 +78,7 @@ void web3signerYamlEndPointRespondsWith200() {
     given()
         .baseUri(signer.getUrl())
         .when()
-        .get(SWAGGER_UI_ENDPOINT + "/web3signer.yaml")
+        .get(SWAGGER_UI_ENDPOINT + "/eth2/web3signer.yaml")
```
joaquim-verges, line 81: this is the failing test. Can't see exactly what's wrong with this though, at first glance seems like it should work 🤔

joaquim-verges (commented):
Huge progress! Much more robust solution. I'm still baffled that we have to do this relative/absolute paths gymnastic in the first place. Really wish the $ref standard was more standardized across the different OpenAPI tools. 

Getting super close though!

joaquim-verges:
btw @usmansaleem feel free to close my PR and do the API split in yours! Probably simpler to test and merge confidently.

usmansaleem:
> btw @usmansaleem feel free to close my PR and do the API split in yours! Probably simpler to test and merge confidently.

Thank you, @joaquim-verges , I'll proceed with split openapi specs in this PR 👍🏼

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/SwaggerUIAcceptanceTest.java
usmansaleem: Its Vertx 3.x handling of trailing / in route paths. I've added quick fix and a note in Runner class.

usmansaleem:
> Huge progress! Much more robust solution. I'm still baffled that we have to do this relative/absolute paths gymnastic in the first place. Really wish the $ref standard was more standardized across the different OpenAPI tools.
Its the swagger-parser-v3 which has issues with relative paths (specially when involved with multi-level folders), going to create a reproducer and file an issue with them.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/SwaggerUIAcceptanceTest.java
joaquim-verges: ah I see, nice!

joaquim-verges (approved):
lgtm! I'll let @jframe take the final look, but this should unblock the rest of the tasks in #439 

Thank you @usmansaleem !

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -113,7 +113,16 @@ public void run() {
       }
       incSignerLoadCount(metricsSystem, artifactSignerProvider.availableIdentifiers().size());
 
-      final OpenAPI3RouterFactory routerFactory = getOpenAPI3RouterFactory(vertx);
+      final OpenApiSpecsExtractor openApiSpecsExtractor =
+          new OpenApiSpecsExtractor.OpenApiSpecsExtractorBuilder()
+              .withFixRelativeRefPaths(true)
+              .build();
```
jframe, line 119: Also want withForceDeleteOnJvmExit(true)?

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -113,7 +113,16 @@ public void run() {
...
+      final Path openApiSpec =
+          openApiSpecsExtractor
+              .getSpecFilePathAtDestination(getOpenApiSpecResource())
+              .orElseThrow();
```
jframe, line 123: Be good throw with a nice error message

core/src/main/java/tech/pegasys/web3signer/core/util/OpenApiSpecsExtractor.java
```diff
@@ -0,0 +1,206 @@
...
+  private List<Path> extractToDestination() throws IOException {
+    final List<Path> extractedResources = new ArrayList<>();
+
+    try (final Stream<URL> webrootYamlFiles =
+        Resources.find("/" + OPENAPI_RESOURCES_ROOT + "**.*")) {
+      webrootYamlFiles.forEach(
+          openapiResource -> {
+            final String jarPath = openapiResource.getPath();
+            // the path is in file://.jar!<path> format. We need the non-jar path.
+            final String filePath =
```
jframe, line 155: This is going to fail if we run Web3Signer through our IDE. It was failing locally for me

core/src/main/java/tech/pegasys/web3signer/core/util/OpenApiSpecsExtractor.java
```diff
@@ -0,0 +1,206 @@
...
+                StringUtils.substringAfter(jarPath, "!/" + OPENAPI_RESOURCES_ROOT + "/");
+            final Path extractedResource = destinationDirectory.resolve(filePath);
+            copyContents(openapiResource, extractedResource);
+            extractedResources.add(extractedResource);
+          });
+    }
+    return Collections.unmodifiableList(extractedResources);
+  }
+
+  private static void copyContents(final URL openapiResource, final Path extractedResource) {
```
jframe, line 165: Could we avoid copying files? I think OpenApiV3Parser takes a string and we are also returning a string to Vertx.

jframe: This doesn't need to be static

core/src/test/java/tech/pegasys/web3signer/core/util/OpenApiSpecsExtractorTest.java
```diff
@@ -0,0 +1,42 @@
...
+  void openapiSpecsAreExtractedAndLoaded(final String spec) throws Exception {
+    final OpenApiSpecsExtractor openApiSpecsExtractor =
+        new OpenApiSpecsExtractor.OpenApiSpecsExtractorBuilder()
+            .withFixRelativeRefPaths(true)
+            .build();
+    final Optional<Path> specPath = openApiSpecsExtractor.getSpecFilePathAtDestination(spec);
+
+    // assert that OpenAPI3RouterFactory is able to load the extracted specs
+    assertThatCode(() -> Runner.getOpenAPI3RouterFactory(Vertx.vertx(), specPath.get().toString()))
+        .doesNotThrowAnyException();
```
jframe, line 40: Perhaps add some asserts that the paths are converted to relative paths

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -224,22 +234,62 @@ private void registerHttpHostAllowListHandler(final OpenAPI3RouterFactory openAp
...
+                    throw new UncheckedIOException(e);
+                  }
+                })
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+    // map /swagger-ui to index.html content
+    final String index_html =
+        swaggerUIContentMap.get(Path.of(SWAGGER_ENDPOINT).resolve("index.html"));
+    swaggerUIContentMap.put(Path.of(SWAGGER_ENDPOINT), index_html);
+    return swaggerUIContentMap;
```
jframe, line 292: This is probably large enough that we want to move this into it's own handler now

core/src/main/java/tech/pegasys/web3signer/core/util/OpenApiSpecsExtractor.java
usmansaleem: fixed.

core/src/main/java/tech/pegasys/web3signer/core/util/OpenApiSpecsExtractor.java
usmansaleem: we need to extract resource files from jar to local dist.

jframe (approved):

PR #455, opened 2021-11-25
Introduce new Key Manager endpoints

joaquim-verges:
Adds the new Ethereum Key manager endpoints ( as defined in https://github.com/ethereum/keymanager-APIs) in the OpenAPI spec.

Also adds a pass through (for now) security handler to the OpenAPI router factory (required by Vertx)

This is one of the tasks of the Key manager API integration discussed in #439


core/src/main/resources/openapi-specs/eth2/keymanager/paths/keystores.yaml
```diff
@@ -0,0 +1,180 @@
...
+                        - error: Any other status different to the above: decrypting error, I/O errors, etc.
+                      enum:
+                        - imported
+                        - duplicate
+                        - error
+                      example: imported
+                    message:
+                      type: string
+                      description: error message if status == error
+    "401":
```
jframe, line 104: This is missing the 400 bad response from the keymanager openapi spec.

core/src/main/resources/openapi-specs/eth2/keymanager/paths/keystores.yaml
```diff
@@ -0,0 +1,180 @@
...
+                        - not_active
+                        - not_found
+                        - error
+                      example: deleted
+                    message:
+                      type: string
+                      description: error message if status == error
+              slashing_protection:
+                $ref: "../schemas.yaml#/components/schemas/SlashingProtectionData"
+    "401":
```
jframe, line 175: Missing 400 bad request response

core/src/main/resources/openapi-specs/eth2/keymanager/schemas.yaml
```diff
@@ -0,0 +1,57 @@
...
+      example: '{"metadata":{"interchange_format_version":"5","genesis_validators_root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"},"data":[{"pubkey":"0x93247f2209abcacf57b75a51dafae777f9dd38bc7053d1af526f220a7489a6d3a2753e5f3e8b1cfe39b56f43611df74a","signed_blocks":[],"signed_attestations":[]}]}'
+
+    ErrorResponse:
+      type: object
+      required: [message]
+      properties:
+        message:
+          description: "Message describing error"
+          type: string
+          example: "Internal server error"
```
jframe, line 35: In the spec it has the values instead
```
          description: "Detailed error message"
          type: string
          example: "description of the error that occurred"
```

core/src/main/resources/openapi-specs/eth2/keymanager/schemas.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  responses:
+    Unauthorized:
```
jframe, line 38: Missing the BadRequest response
```
    BadRequest:
      description: "Bad request. Request was malformed and could not be processed"
      content:
        application/json:
          schema:
            $ref: "#/components/schemas/ErrorResponse"
```

core/src/main/resources/openapi-specs/eth2/keymanager/schemas.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+    Forbidden:
+      description: "Forbidden, a token is found but is invalid"
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorResponse"
+
+    InternalError:
+      description: "Beacon node internal error."
```
jframe, line 53: The description here is different from the spec. In the spec it has `      description: "Internal server error. The server encountered an unexpected error indicative of
                    a serious fault in the system, or a bug."`

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -123,6 +123,20 @@ public void run() {
       if (config.isAccessLogsEnabled()) {
         routerFactory.addGlobalHandler(LoggerHandler.create(LoggerFormat.DEFAULT));
       }
+
+      // security handler for keymanager endpoints
+      routerFactory.addSecurityHandler(
```
jframe, line 128: Can this be moved into the Eth2Runner as this only applies for Eth2

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -123,6 +123,20 @@ public void run() {
...
+          "bearerAuth",
+          context -> {
+            // TODO Auth token security logic
+            final boolean autorized = true;
```
jframe, line 132: nit: authorized instead of autorized

joaquim-verges: woops

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
joaquim-verges: done

core/src/main/resources/openapi-specs/eth2/keymanager/schemas.yaml
joaquim-verges: updated everything to the latest spec 👍

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -129,6 +129,19 @@ private void registerEth2Routes(
       final Optional<SlashingProtection> slashingProtection) {
     final ObjectMapper objectMapper = SigningObjectMapperFactory.createObjectMapper();
 
+    // security handler for keymanager endpoints
```
jframe, line 132: nit: Would be better if adding this was conditional on the key manager api being enabled

jframe (approved):

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
joaquim-verges: unfortunately Vertx requires this at parsing time, as soon as it sees an auth'd endpoint in the spec it goes look for the a security handler and crashes if none is found.

PR #457, opened 2021-12-01
Publish multiarch jdk17 docker images

usmansaleem:
- Docker images are now published with multi-arch support including Linux/amd64 and Linux/arm64
- The default docker image now uses JDK 17 instead of 11. The JDK 11 image is still available with the version suffix `-jdk11`
- The docker image now uses `web3signer` as user/group instead of `root` which may result in compatibility/permissions issues with existing directory mounts.


docker/tests/01/goss.yaml
```diff
@@ -2,8 +2,8 @@ file:
   /opt/web3signer/bin/web3signer:
     exists: true
     mode: "0755"
-    owner: root
-    group: root
+    owner: web3signer
+    group: web3signer
```
ajsutton, line 6: Bear in mind that changing from the default user being root to web3signer introduces backwards incompatibility when directories are mounted into the docker container as ownership rights won't match up.  You'll want to at least note that in the changelog and may need to give some advance warning about the change.

ajsutton (approved):
LGTM but beware the default user change.

jframe (approved):
LGTM

docker/tests/01/goss.yaml
usmansaleem: that is a good point, I have made a changelog entry.

jframe (approved):

PR #481, opened 2022-01-07
Updating slashing-protection url

fredriksvantes:
https://github.com/eth2-clients was renamed to https://github.com/eth-clients


jframe (approved):
Thanks for your contribution and fixing the interchange tests repo location.

Change looks good.

PR #482, opened 2022-01-10
Rename the eth cmd line option --enable-key-manager-api to --key-manager-api-enabled

jframe:
Rename the eth cmd line option --enable-key-manager-api to --key-manager-api-enabled to make it consistent with the other Web3Signer command options that have -enabled suffix.


siladu (approved):
LGTM

joaquim-verges:
Thank you @jframe !

prestonvanloon:
Please alias the old flag name. I am running from HEAD and was broken with this error. 

```
Error parsing parameters: Unknown option: '--enable-key-manager-api=true'
```

jframe:
Looks like the old flag went out in a release so adding an alias for it see #518.

PR #489, opened 2022-01-18
Implement Key Manager Delete Endpoint

joaquim-verges:
API spec: https://github.com/ethereum/keymanager-APIs/tree/master/flows#delete

Part of the keymanger API implementation: #439

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
@@ -14,6 +14,7 @@ package tech.pegasys.web3signer.core;
 
 import static tech.pegasys.web3signer.core.service.http.OpenApiOperationsId.ETH2_LIST;
 import static tech.pegasys.web3signer.core.service.http.OpenApiOperationsId.ETH2_SIGN;
+import static tech.pegasys.web3signer.core.service.http.OpenApiOperationsId.KEYMANAGER_DELETE;
 import static tech.pegasys.web3signer.core.service.http.OpenApiOperationsId.KEYMANAGER_IMPORT;
 import static tech.pegasys.web3signer.core.service.http.OpenApiOperationsId.KEYMANAGER_LIST;
 import static tech.pegasys.web3signer.core.service.http.OpenApiOperationsId.RELOAD;
@@ -38,6 +39,7 @@ import tech.pegasys.web3signer.core.multikey.metadata.parser.YamlSignerParser;
 import tech.pegasys.web3signer.core.multikey.metadata.yubihsm.YubiHsmOpaqueDataProvider;
 import tech.pegasys.web3signer.core.service.http.SigningObjectMapperFactory;
 import tech.pegasys.web3signer.core.service.http.handlers.LogErrorHandler;
+import tech.pegasys.web3signer.core.service.http.handlers.keymanager.delete.DeleteKeystoresHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.keymanager.imports.ImportKeystoresHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.keymanager.list.ListKeystoresHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.signing.SignerForIdentifier;
@@ -176,6 +178,13 @@ public class Eth2Runner extends Runner {
           KEYMANAGER_IMPORT.name(),
           new BlockingHandlerDecorator(
               new ImportKeystoresHandler(
+                  objectMapper, config.getKeyConfigPath(), slashingProtection, blsSignerProvider),
+              false));
+
+      routerFactory.addHandlerByOperationId(
+          KEYMANAGER_DELETE.name(),
+          new BlockingHandlerDecorator(
+              new DeleteKeystoresHandler(
                   objectMapper, config.getKeyConfigPath(), slashingProtection, blsSignerProvider),
               false));
     }
--- a/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
@@ -174,6 +174,7 @@ public class Eth2Runner extends Runner {
           KEYMANAGER_LIST.name(),
           new BlockingHandlerDecorator(
               new ListKeystoresHandler(blsSignerProvider, objectMapper), false));
+      routerFactory.addFailureHandlerByOperationId(KEYMANAGER_DELETE.name(), errorHandler);
 
       routerFactory.addHandlerByOperationId(
           KEYMANAGER_IMPORT.name(),
@@ -181,6 +182,7 @@ public class Eth2Runner extends Runner {
               new ImportKeystoresHandler(
                   objectMapper, config.getKeyConfigPath(), slashingProtection, blsSignerProvider),
               false));
+      routerFactory.addFailureHandlerByOperationId(KEYMANAGER_IMPORT.name(), errorHandler);
 
       routerFactory.addHandlerByOperationId(
           KEYMANAGER_DELETE.name(),
@@ -191,6 +193,7 @@ public class Eth2Runner extends Runner {
                   slashingProtection,
                   blsSignerProvider),
               false));
+      routerFactory.addFailureHandlerByOperationId(KEYMANAGER_DELETE.name(), errorHandler);
     }
   }
 
--- a/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
@@ -40,6 +40,7 @@ import tech.pegasys.web3signer.core.multikey.metadata.yubihsm.YubiHsmOpaqueDataP
 import tech.pegasys.web3signer.core.service.http.SigningObjectMapperFactory;
 import tech.pegasys.web3signer.core.service.http.handlers.LogErrorHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.keymanager.delete.DeleteKeystoresHandler;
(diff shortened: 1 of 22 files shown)


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+        // Remove active key from memory first, will stop any further signing with this key
+        signerProvider.removeSigner(pubkey).get();
+        // Then, delete the corresponding keystore file
+        if (keystoreConfigFile.isPresent()) {
+          Files.delete(keystoreConfigFile.get());
+        }
+        // finally, add result response
+        keysToExport.add(pubkey);
+        results.add(new DeleteKeystoreResult(DeleteKeystoreStatus.DELETED, ""));
```
jframe, line 133: If the keystore is read-only then we should be returning an error as it hasn't been deleted

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
```diff
@@ -0,0 +1,229 @@
...
+      } catch (Exception e) {
+        results.add(
```
jframe, line 135: Can a test be added for the case where deleting a key fails

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
```diff
@@ -0,0 +1,229 @@
...
+
+    // export slashing protection data for 'deleted' and 'not_active' keys
+    String slashingProtectionExport = null;
+    if (slashingProtection.isPresent()) {
+      try {
+        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        slashingProtection.get().exportWithFilter(outputStream, keysToExport);
+        slashingProtectionExport = outputStream.toString(StandardCharsets.UTF_8);
+      } catch (Exception e) {
+        LOG.debug("Failed to export slashing data", e);
```
jframe, line 149: Can test be added for the case that slashing protection data can't be exported and errors need to be returned for the keys

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
```diff
@@ -0,0 +1,229 @@
...
+          .setStatusCode(SUCCESS)
+          .end(
+              objectMapper.writeValueAsString(
+                  new DeleteKeystoresResponse(results, slashingProtectionExport)));
+    } catch (JsonProcessingException e) {
+      context.fail(SERVER_ERROR, e);
+    }
+  }
+
+  private Optional<Path> findKeystoreConfigFile(final String pubkey) throws IOException {
```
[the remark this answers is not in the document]
joaquim-verges: that would be nice - for now to clean things up I can extract this out into its own class

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
joaquim-verges: any good example on how to simulate an IO failure? not sure how to do that in tests

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
joaquim-verges: how do you simulate an export failure? any examples in other tests?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
```diff
@@ -0,0 +1,198 @@
...
+              new DeleteKeystoreResult(
+                  DeleteKeystoreStatus.ERROR, "Unable to delete readonly key: " + pubkey));
+          continue;
+        }
+
+        // Remove active key from memory first, will stop any further signing with this key
+        signerProvider.removeSigner(pubkey).get();
+        // Then, delete the corresponding keystore file
+        if (keystoreConfigFiles.isPresent()) {
+          for (final Path path : keystoreConfigFiles.get()) {
```
jframe, line 144: Seems like we don't need to the key config files we just need to be able to delete them. Perhaps change the KeystoreFileManager api to have a delete method instead of a findKeystoreConfigFiles method.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
```diff
@@ -0,0 +1,198 @@
...
+      final List<DeleteKeystoreResult> results, final List<String> keysToExport) {
+    // export slashing protection data for 'deleted' and 'not_active' keys
+    String slashingProtectionExport = null;
+    if (slashingProtection.isPresent()) {
+      try {
+        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        slashingProtection.get().exportWithFilter(outputStream, keysToExport);
+        slashingProtectionExport = outputStream.toString(StandardCharsets.UTF_8);
+      } catch (Exception e) {
+        LOG.error("Failed to export slashing data", e);
```
jframe, line 171: Can you add a test for failing to export the slashing protection data

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresHandler.java
joaquim-verges: done

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -178,6 +181,16 @@ private void registerEth2Routes(
...
+
+      routerFactory.addHandlerByOperationId(
+          KEYMANAGER_DELETE.name(),
+          new BlockingHandlerDecorator(
+              new DeleteKeystoresHandler(
+                  objectMapper,
+                  new KeystoreFileManager(config.getKeyConfigPath()),
+                  slashingProtection,
+                  blsSignerProvider),
+              false));
```
siladu, line 193: Does this need the errorHandler adding as failure handler? I think same applies to the list and import operations too.

siladu: Maybe a question for @jframe

joaquim-verges:
@jframe added some unit tests for all the checked exceptions - lmk if that works for you

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
jframe: Yes we should add that. That will add some additional logging if the request fails.

jframe:
> @jframe added some unit tests for all the checked exceptions - lmk if that works for you

Unit tests look good.

jframe (commented):
Looks your PR is failing due to spotless

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
joaquim-verges: done

joaquim-verges:
@jframe @siladu thanks for the reviews!

jframe (approved):

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresProcessor.java
```diff
@@ -0,0 +1,143 @@
...
+        if (signer.get() instanceof BlsArtifactSigner
+            && ((BlsArtifactSigner) signer.get()).isReadOnlyKey()) {
+          results.add(
+              new DeleteKeystoreResult(
+                  DeleteKeystoreStatus.ERROR, "Unable to delete readonly key: " + pubkey));
+          continue;
+        }
+
+        // Remove active key from memory first, will stop any further signing with this key
+        signerProvider.removeSigner(pubkey).get();
```
siladu, line 101: nit: If we want removeSigner to be synchronous (i.e. wait until it's definitely deleted) for all usages then is there any need for removeSigner to be implemented with an executerService?

siladu: I think the same could be said for addSigner

siladu (approved):
One nitpick about some redundant code but otherwise looks good. I like the splitting out of the processor and key manager classes, makes it easier to digest and think about test cases.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/DeleteKeystoresProcessor.java
joaquim-verges: the reason why I kept it using the executorService is because that way I can guarantee that the add/remove happens on a single thread (the executor used there is a `Executors.newSingleThreadExecutor()`)

siladu: ah, makes sense - thanks for the explanation!

PR #508, opened 2022-02-07
docker multi architecture signing

joshuafernandes:
Buildx doesn't allow for signing multi architecture images, so have removed it. In its stead the docker build job now runs on an ARM and AMD executor and builds, signs and pushes the image with the architecture in the tag as well. The final step is a manifest create and push which manually creates the manifest from the multiarch tags and then pushes that manifest out

Refer images here https://hub.docker.com/repository/docker/consensys/dctweb3signer

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,12 +58,18 @@ executors:
     machine:
       image: ubuntu-2004:202101-01
     resource_class: arm.medium
+    environment:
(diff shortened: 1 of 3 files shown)


Ezzahhh (approved):
LGTM

usmansaleem (approved):
Looks good to me.

PR #513, opened 2022-02-16
Cors

jframe:
Add CORS support to web3signer.

This adds a new CLI option --http-cors-origins that can be used to specify any domains that CORS should be enabled for. Defaults to none.


usmansaleem (approved):
LGTM

PR #514, opened 2022-02-21
Allow validators to be disabled

jframe:
This adds the ability for validators to be disabled by adding the checking logic and database changes needed to support this. It does not add anything to disable a validator, this will done in a later PR.

* Slashing protection now checks that the validator is enabled before proceeding to do any checks against the slashing protection database
* Adds new column in the validators table, enabled which indicates whether the validator is enabled or not
* Existing validators are treated as being enabled due to default value true on the new enabled column


slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDaoTest.java
```diff
@@ -81,8 +86,50 @@ public void storesUnregisteredValidatorsInDb(final Handle handle) {
...
+    assertThat(new ValidatorsDao().isEnabled(handle, 1)).isTrue();
+  }
+
+  @Test
+  public void isEnabledReturnsTrueForExistingValidator() {
+    final TestDatabaseInfo testDatabaseInfo = DatabaseUtil.createWithoutMigration();
+    final Jdbi jdbi = testDatabaseInfo.getJdbi();
+
+    try (Handle handle = jdbi.open()) {
+      final String versionBeforeEnableFlag = String.valueOf(VALIDATOR_ENABLE_FLAG_VERSION - 1);
```
usmansaleem, line 101: should we declare constant '9' as that is known to be not have validator enabled logic? Otherwise, when version is 11, the above would return 10.

usmansaleem: or VALIDATOR_ENABLE_FLAG_VERSION is meant to remain 10 and not to be incremented?

slashing-protection/src/testFixtures/java/db/DatabaseUtil.java
```diff
@@ -26,21 +27,26 @@
   public static final String PASSWORD = "postgres";
 
   public static TestDatabaseInfo create() {
+    final TestDatabaseInfo testDatabaseInfo = createWithoutMigration();
+    final Flyway flyway =
+        Flyway.configure()
+            .locations("/migrations/postgresql/")
```
usmansaleem, line 33: may be we can extract `/migrations/postgresql/` as a constant? It seems it is used almost everywhere.

usmansaleem (approved):
LGTM - minor observations.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDaoTest.java
jframe: Was going to leave VALIDATOR_ENABLE_FLAG_VERSION as a constant since that the version the enable flag is added in

slashing-protection/src/testFixtures/java/db/DatabaseUtil.java
jframe: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDao.java
```diff
@@ -15,6 +15,8 @@
 import org.jdbi.v3.core.Handle;
 
 public class DatabaseVersionDao {
+  public static final int EXPECTED_DATABASE_VERSION = 10;
+  public static final int VALIDATOR_ENABLE_FLAG_VERSION = 10;
```
siladu, line 19: Looks like this is only used in the tests? Although I see it's related to the EXPECTED_DATABASE_VERSION, I think here it could potentially be confused as a feature flag of some sort.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDaoTest.java
```diff
@@ -81,8 +87,50 @@ public void storesUnregisteredValidatorsInDb(final Handle handle) {
...
-    final byte[] value = Bytes.of(i).toArrayUnsafe();
-    h.execute("INSERT INTO validators (public_key) VALUES (?)", value);
+  @Test
+  public void isEnabledReturnsTrueForEnabledValidator(final Handle handle) {
+    insertValidator(handle, Bytes.of(1), true);
+    assertThat(new ValidatorsDao().isEnabled(handle, 1)).isTrue();
+  }
+
+  @Test
+  public void isEnabledReturnsTrueForExistingValidator() {
```
siladu, line 97: I like that you've tested the migration compatibility

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtectionTest.java
```diff
@@ -20,6 +20,7 @@
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.refEq;
 import static org.mockito.Mockito.atLeast;
+import static org.mockito.Mockito.lenient;
```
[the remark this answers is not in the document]
jframe: To make the tests simpler. Almost all the tests required the validatorDao.isEnabled to return true and the metadataDao.findGenesisValidatorsRoot to return the GVR.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/DatabaseVersionDao.java
jframe: Hmm.. maybe need a better name. Wanted to give a name to the validator enable flag migration. Any suggestions?

siladu: I forgot to mention it but I was thinking it would be better scoped in the test packages rather than the Dao itself (then I guess the name wouldn't matter too much). 
I confused things by mentioning feature flags, it's more about the fact that I'd expect this constant to be used in production code if it's here.

Issue #494, opened 2022-01-24
Error parsing parameters: Failed to load network spec


prestonvanloon:
```
Error parsing parameters: Failed to load network spec: /path/to/e2e_config.yaml
```

I'm receiving the above error without any reason or justification. The docs mention the flag value can be "Predefined network configuration. Accepts a predefined network name, or file path or URL to a YAML configuration file. See the Ethereum 2.0 specification for examples." 

```
Version = web3signer/v21.10.5/linux-x86_64/-ubuntu-openjdk64bitservervm-java-11
```

This is the file I'm trying to load. It's a modified version of the [minimal config](https://github.com/ethereum/consensus-specs/blob/180855b579830c3f757bb63eda3bf3793f37cd90/configs/minimal.yaml). 

```yaml
# e2e config

# Extends the minimal preset
PRESET_BASE: 'minimal'


# Transition
# ---------------------------------------------------------------
# TBD, 2**256-2**10 is a placeholder
TERMINAL_TOTAL_DIFFICULTY: 115792089237316195423570985008687907853269984665640564039457584007913129638912
# By default, don't use these params
TERMINAL_BLOCK_HASH: 0x0000000000000000000000000000000000000000000000000000000000000000
TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH: 18446744073709551615



# Genesis
# ---------------------------------------------------------------
# [customized]
MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: 256 # Override for e2e tests
# Jan 3, 2020
MIN_GENESIS_TIME: 1578009600
# Highest byte set to 0x01 to avoid collisions with mainnet versioning
GENESIS_FORK_VERSION: 0x00000001
# [customized] Faster to spin up testnets, but does not give validator reasonable warning time for genesis
GENESIS_DELAY: 10 # Override for e2e tests


# Forking
# ---------------------------------------------------------------
# Values provided for illustrative purposes.
# Individual tests/testnets may set different values.

# Altair
ALTAIR_FORK_VERSION: 0x01000001
ALTAIR_FORK_EPOCH: 6 # Override for e2e
# Bellatrix
BELLATRIX_FORK_VERSION: 0x02000001
BELLATRIX_FORK_EPOCH: 18446744073709551615
# Sharding
SHARDING_FORK_VERSION: 0x03000001
SHARDING_FORK_EPOCH: 18446744073709551615


# Time parameters
# ---------------------------------------------------------------
[...]

PR #515, opened 2022-02-21
Updating Teku libraries to 22.3.0

usmansaleem:
#494 should get fixed with latest Teku libraries which should allow loading of external network spec and allows to override configuration options.


siladu (approved):
LGTM

PR #516, opened 2022-02-22
Postgres 11

jframe:
Test web3signer using Postgresql 11 as we want this to be recommended version for further db changes and postgres 10.6 crashed on OSX Monterey when running the tests.

* Changes the embedded postgresql library from opentable library to zonky as this let's us specify which postgresql version we want to use
* Update the version of postgresql we are testing against to the latest postgres 11 release


slashing-protection/src/testFixtures/java/db/DatabaseUtil.java
```diff
@@ -0,0 +1,58 @@
...
+      flyway.migrate();
+
+      final String databaseUrl =
+          String.format("jdbc:postgresql://localhost:%d/postgres", db.getPort());
+      final Jdbi jdbi =
+          DbConnection.createConnection(
+              databaseUrl, DatabaseUtil.USERNAME, DatabaseUtil.PASSWORD, null);
+      return new TestDatabaseInfo(db, jdbi);
+    } catch (IOException e) {
+      throw new RuntimeException("Unable to create embedded postgres database", e);
```
usmansaleem, line 44: nit: `UncheckedIOException` might be a better choice to wrap `IOException`.

usmansaleem (approved):
LGTM, minor nit-pick.

usmansaleem (approved):
LGTM

slashing-protection/src/testFixtures/java/db/DatabaseUtil.java
jframe: done

usmansaleem (approved):
LGTM

PR #517, opened 2022-02-23
Signing of BLS keys retrieved from AWS Secrets Manager

lucassaldanha:
Supersedes #511 w/o the acceptance tests for AWS Secrets Manager


core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/AwsKeySigningMetadata.java
```diff
@@ -0,0 +1,75 @@
...
+public class AwsKeySigningMetadata extends SigningMetadata implements AwsSecretsManagerParameters {
+
+  private final AwsAuthenticationMode authenticationMode;
+  private final String region;
+  private final String accessKeyId;
+  private final String secretAccessKey;
+  private final String secretName;
+
+  @JsonCreator
+  public AwsKeySigningMetadata(
```
jframe, line 32: Probably should validate that the accessKeyId and secretAccessKey are supplied when in the SPECIFIED mode. Perhaps using a custom deserialiser as done with the Azure i.e. AzureSecretSigningMetadataDeserializer

jframe, line 32: Adding a file parsing test would be good as well then we would know jackson is all setup correctly, we did this for interlock and yubihsm see e.g. InterlockSigningMetadataFileParsingTest

jframe (approved):
LGTM

PR #518, opened 2022-02-24
Add alias for old key manager flag

jframe:
Add an alias for old key manager flag "--enable-key-manager-api=true" as the old flag went out in a release and there are people still using the old flag.


georgep9 (approved):
LGTM

PR #520, opened 2022-03-02
Disable validator when deleting keystore

### LINKED ISSUE #509: Disable validator before deleting key
Validators need to be disabled at the database level prevent other instances from inserting slashing protection data during and after the key manager delete API has been called.

**Tasks**
- [x] Modify the database schema so the active/disabled status can be set
- [x] In the DbSlashingProtection maySignAttestation and maySignBlock add a check to ensure the validator is active before attempting any checks or signing.
- [x] Add new method to update the disabled status in the appropriate Dao most likely the ValidatorsDao
- [ ] In the DeleteKeystoresProcessor disable the validator before exporting the slashing protection data.

**Acceptance Criteria**
- It is not possible for the slashing protection data to be inserted by any instance of web3signer once the key has been deleted
- If key files are added again the validator marked active again and sign data and insert records into the slashing protection database

jframe:
This disables the validator when the keystore is deleted using the keystore API. This ensures that even if other web3signers are sharing the same database there will be no further signings or slashing protection data will be written to the database.

fixes #509 

* Disable the validator in the validators table when deleting the keystore
* Enable the validator in the validators table when importing the keystore
* If any errors occur during the delete the validator enable status will not be changed
* Moved the validator adding and removing logic into ValidatorManager


signing/src/main/java/tech/pegasys/web3signer/signing/FileValidatorManager.java
```diff
@@ -0,0 +1,81 @@
...
+      // 2. write keystore file to disk
+      keystoreFileManager.createKeystoreFiles(signer.getIdentifier(), keystore, password);
+      // 3. add the new signer to the provider to make it available for signing
+      signerProvider.addSigner(signer).get();
+    } catch (IOException | InterruptedException | ExecutionException e) {
+      throw new IllegalStateException("Unable to add validator", e);
+    }
+  }
+
+  private BlsArtifactSigner decryptKeystoreAndCreateSigner(String jsonKeystoreData, String password)
```
usmansaleem, line 73: nit: `final String jsonKeystoreData ... `

signing/src/main/java/tech/pegasys/web3signer/signing/FileValidatorManager.java
```diff
@@ -0,0 +1,81 @@
...
+      throws JsonProcessingException, KeyStoreValidationException {
+    final KeyStoreData keyStoreData = objectMapper.readValue(jsonKeystoreData, KeyStoreData.class);
+    final Bytes privateKey = KeyStore.decrypt(password, keyStoreData);
```
usmansaleem, line 76: Keystore.decrypt can throw a runtime exception I believe. We might need to handle it at line 68 rather than bubbling it up.

signing/src/test/java/tech/pegasys/web3signer/signing/FileValidatorManagerTest.java
```diff
@@ -0,0 +1,105 @@
...
+    final KeyStoreData keyStoreData =
+        KeyStore.encrypt(
+            BLS_KEY_PAIR.getSecretKey().toBytes(),
+            BLS_KEY_PAIR.getPublicKey().toBytesCompressed(),
+            "password",
+            "",
+            pbkdf2Param,
+            cipher);
+
+    return new ObjectMapper()
```
usmansaleem, line 101: nit: It might be a good idea to use the global ObjectMapper (if it is exposed). Or at minimum, consider declaring it as static constant.

usmansaleem (approved):
LGTM - minor comments.

signing/src/main/java/tech/pegasys/web3signer/signing/FileValidatorManager.java
jframe: done

signing/src/main/java/tech/pegasys/web3signer/signing/FileValidatorManager.java
jframe: Aah yes It does throw a runtime exception after all the KeyStoreValidationException. Added that to the catch clause.

signing/src/test/java/tech/pegasys/web3signer/signing/FileValidatorManagerTest.java
jframe: added a static constant

usmansaleem (approved):
LGTM

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/RegisteredValidatorsTest.java
```diff
@@ -0,0 +1,111 @@
...
+    registeredValidators.registerValidators(List.of(PUBLIC_KEY1));
+    assertThat(registeredValidatorsMap).hasSize(1);
+
+    registeredValidators.registerValidators(List.of(PUBLIC_KEY1, PUBLIC_KEY2, PUBLIC_KEY3));
+    assertThat(registeredValidatorsMap).hasSize(3);
+    // because 'id' is a sequence, the values will be 1, 2, 3
+    assertThat(registeredValidatorsMap)
+        .isEqualTo(Map.of(PUBLIC_KEY1, 1, PUBLIC_KEY2, 2, PUBLIC_KEY3, 3));
+  }
+}
```
siladu, line 111: missing tests for mustGetValidatorIdForPublicKey?

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDaoTest.java
```diff
@@ -88,6 +88,20 @@ public void storesUnregisteredValidatorsInDb(final Handle handle) {
     assertThat(validators.get(4)).isEqualToComparingFieldByField(new Validator(5, Bytes.of(104)));
   }
 
+  @Test
+  public void registeringValidatorsDoesNotEnableDisabledValidators(final Handle handle) {
```
siladu, line 92: Can you point me to the code that prevents disabled validators from being reenabled?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
```diff
@@ -180,6 +189,66 @@ public void deletingReadOnlyKeyReturnError() throws URISyntaxException {
...
+      throws URISyntaxException, JsonProcessingException {
+    final String firstPubkey = createBlsKey("eth2/bls_keystore.json", "somepassword");
+    final String secondPubKey = createBlsKey("eth2/bls_keystore_2.json", "otherpassword");
+    setupSignerWithKeyManagerApi(true);
+
+    final SignerConfiguration signer2Configuration =
+        new SignerConfigurationBuilder()
+            .withKeyStoreDirectory(testDirectory)
+            .withMode("eth2")
+            .withNetwork("minimal")
```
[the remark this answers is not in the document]
siladu: nvm - found it :) https://github.com/ethereum/consensus-specs/tree/master/configs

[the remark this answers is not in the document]
siladu: Would be interesting to know what impact setting that has on this particular test (compared to when it wasn't set before)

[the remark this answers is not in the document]
jframe: For this test it's only needed because I wanted to sign a V2 block. Previously this test didn't have any need to sign blocks. I only affects tests that are signing and specifically require use of the forks. The import test also does has signing part to it but only signs an attestation that is unaffected by this change.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
```diff
@@ -180,6 +189,66 @@ public void deletingReadOnlyKeyReturnError() throws URISyntaxException {
...
+        .contentType(ContentType.JSON)
+        .body("data.size()", is(2))
+        .and()
+        .body("data[0].validating_pubkey", is(firstPubkey))
+        .body("data[1].validating_pubkey", is(secondPubKey));
+
+    final Eth2SigningRequestBody attestationRequest =
+        Eth2RequestUtils.createCannedRequest(ArtifactType.ATTESTATION);
+    signer.eth2Sign(firstPubkey, attestationRequest, ContentType.TEXT).then().statusCode(404);
+    signer2.eth2Sign(firstPubkey, attestationRequest, ContentType.TEXT).then().statusCode(412);
```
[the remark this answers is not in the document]
jframe: Yeah it is a bit misleading. The handler assumes it's due to the slashing protection rules failing but that's not always true. Could be GVR missing, database error, validator being disabled. Perhaps we should be returning a richer error answer from the slashing protection with errors that return an error message or code? Think this should be another ticket though.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/ImportKeystoresAcceptanceTest.java
```diff
@@ -155,6 +156,37 @@ public void importLoadsNewKeys() throws IOException, URISyntaxException {
...
+    validateApiResponse(callListKeys(), "data.validating_pubkey", hasItem(PUBLIC_KEY));
+
+    // Sign with it
+    final Eth2SigningRequestBody request = createAttestationRequest(5, 6, UInt64.ZERO);
+    signer.eth2Sign(PUBLIC_KEY, request).then().assertThat().statusCode(200);
+  }
+
+  @Test
+  public void canSignPreviouslyDeletedKeyAfterImporting() throws IOException, URISyntaxException {
+    setupSignerWithKeyManagerApi();
```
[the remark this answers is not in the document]
jframe: I think I messed up resolving a merge conflict. Looks the like code is in the wrong test. Good catch

signing/src/test/java/tech/pegasys/web3signer/signing/FileValidatorManagerTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class FileValidatorManagerTest {
+  private static final Bytes SALT =
+      Bytes.fromHexString("0x9ac471d9d421bc06d9aefe2b46cf96d11829c51e36ed0b116132be57a9f8c22b");
+  private static final Bytes IV = Bytes.fromHexString("0xcca2c67ec95a1dd13edd986fea372789");
+  private static final BLSKeyPair BLS_KEY_PAIR = BLSTestUtil.randomKeyPair(1);
+  private static final ObjectMapper OBJECT_MAPPER =
+      new ObjectMapper().registerModule(new SigningMetadataModule());
```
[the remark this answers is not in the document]
jframe: done

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/RegisteredValidatorsTest.java
jframe: Yes it is. Added some.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDaoTest.java
jframe: There is no specific code for this, it is achieved by not making any changes to the validator registration. In retrospect I think this code should go. The reason it's not enabling validators is because we don't registered validator that already exist. So I don't think this test case makes sense anymore. Originally I was taking a different approach.
Removing the test.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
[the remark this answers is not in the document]
siladu: makes sense, thanks for the explanation

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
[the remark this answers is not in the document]
siladu: agree, I've created a ticket for the backlog: https://github.com/ConsenSys/web3signer/issues/543

siladu (approved):
LGTM

PR #523, opened 2022-03-03
Incremental slashing db export

jframe:
Add incremental export of the slashing protection database. This will allow for changing the keymanger delete functionality to use a incremental slashing protection export necessary to handle database rollbacks with the validator disable correctly.

This also changes the key manager delete to use the incremental export but does not change it to use a streaming approach. Changing delete to process the deletion of keys and slashing protection data one by one will be done in another PR.

### DIFF
--- a/slashing-protection/src/integration-test/java/tech/pegasys/web3signer/slashingprotection/InterchangeExportIntegrationTestBase.java
+++ b/slashing-protection/src/integration-test/java/tech/pegasys/web3signer/slashingprotection/InterchangeExportIntegrationTestBase.java
@@ -125,7 +125,7 @@ public class InterchangeExportIntegrationTestBase extends IntegrationTestBase {
     final IncrementalExporter incrementalExporter =
         slashingProtection.createIncrementalExporter(exportOutput);
     for (int i = 0; i < VALIDATOR_COUNT; i += 2) {
-      incrementalExporter.addPublicKey(String.format("0x0%x", i + 1));
+      incrementalExporter.export(String.format("0x0%x", i + 1));
     }
     incrementalExporter.finalise();
     incrementalExporter.close();
--- a/slashing-protection/src/integration-test/java/tech/pegasys/web3signer/slashingprotection/InterchangeExportIntegrationTestBase.java
+++ b/slashing-protection/src/integration-test/java/tech/pegasys/web3signer/slashingprotection/InterchangeExportIntegrationTestBase.java
@@ -15,6 +15,7 @@ package tech.pegasys.web3signer.slashingprotection;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import tech.pegasys.web3signer.slashingprotection.interchange.IncrementalExporter;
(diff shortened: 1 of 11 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Exporter.java
```diff
@@ -225,4 +257,28 @@ private void writeAttestations(
     }
     jsonGenerator.writeEndArray();
   }
+
+  public class IncrementalInterchangeV5Exporter implements IncrementalExporter {
+    final JsonGenerator jsonGenerator;
+
+    public IncrementalInterchangeV5Exporter(final OutputStream outputStream) throws IOException {
+      jsonGenerator = mapper.getFactory().createGenerator(outputStream);
+      startInterchangeExport(jsonGenerator);
```
siladu, line 266: I like how the encapsulation of this class allows you to reuse this method as well as keeping the class well scoped.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -146,27 +147,37 @@ public void importDataWithFilter(final InputStream input, final List<String> pub
...
+      interchangeManager.exportDataWithFilter(output, pubkeys);
       LOG.info("Export complete");
     } catch (IOException e) {
       throw new RuntimeException("Failed to export database content", e);
     }
   }
 
+  @Override
+  public IncrementalExporter createIncrementalExporter(final OutputStream out) {
+    try {
```
siladu, line 173: nit: Do we want a log here similar to exportData, e.g.
`LOG.info("Exporting slashing protection database");`
 or was it an intentional omission?

slashing-protection/src/integration-test/java/tech/pegasys/web3signer/slashingprotection/InterchangeExportIntegrationTestBase.java
```diff
@@ -92,6 +93,74 @@ void exportedEntitiesRepresentTheEntriesStoredInTheDatabase() throws IOException
...
+      }
+
+      jdbi.useTransaction(
+          h -> {
+            lowWatermarkDao.updateSlotWatermarkFor(h, validatorId, UInt64.ZERO);
+            lowWatermarkDao.updateEpochWatermarksFor(h, validatorId, UInt64.ZERO, UInt64.ZERO);
+          });
+    }
+
+    // incrementally export only the even the public keys
```
siladu, line 123: I like this approach for the test

siladu (approved):
I like the approach with the nested class holding the intermediate export state. 
Just a nitpick about a potentially missing log, but not sure how important that log is anyway.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
jframe: Will need to move the logging though into the IncrementalExporter instead though

PR #526, opened 2022-03-04
Fix #525 Sorting sql files to be executed from V1,V2 to V10 and not from V10,V1 to V9

### LINKED ISSUE #525: Error while installing slashing protection database v10
There's an error while installing the web3signer with the slashing protection database. After moving all sql files from 
`/slashing-protection/src/main/resources/migrations/postgresql/*.sql`
to
`/docker-entrypoint-initdb.d/`

Scripts will run alphabetically, so V10__validator_enabled_status.sql will be executed before V1__initial.sql and never after V9__upsert_validators.sql

mgarciate:
Fixed #525


usmansaleem:
@mgarciate thank you for the PR. Let me test it locally and we can merge it after that.

usmansaleem (approved):
LGTM. See comments in the #525 for additional approaches as well.

Issue #505, opened 2022-02-07
Wallet files created using key manager import have incorrect structure


jframe:
The key manager API import endpoint is creating wallet files with the incorrect structure. The type of the file isn't specified and the files paths are file URIs.

e.g.
```
--- !<file-keystore>
keystoreFile: "file:///Users/jason/code/web3signer/tmp/keymanager/0x98c8f1b6cbaf4210426743638b7a421fb99db2614a6246853da41413ed55bd886faf6df8a46b84eca166e418351f2b24.json"
keystorePasswordFile: "file:///Users/jason/code/web3signer/tmp/keymanager/0x98c8f1b6cbaf4210426743638b7a421fb99db2614a6246853da41413ed55bd886faf6df8a46b84eca166e418351f2b24.password"
keyType: "BLS"
```

It should be:
```
type: "file-keystore"
keystoreFile: "/Users/jason/code/web3signer/tmp/keymanager/0x98c8f1b6cbaf4210426743638b7a421fb99db2614a6246853da41413ed55bd886faf6df8a46b84eca166e418351f2b24.json"
keystorePasswordFile: "/Users/jason/code/web3signer/tmp/keymanager/0x98c8f1b6cbaf4210426743638b7a421fb99db2614a6246853da41413ed55bd886faf6df8a46b84eca166e418351f2b24.password"
keyType: "BLS"
```

Tasks
1. Make the automated tests break for this bug.
2. Investigate and use the correct Jackson config to produce the right format.
3. Remove the file:/// prefix.
4. Ensure network resources can't be loaded (e.g. over http://)

PR #527, opened 2022-03-06
[#505] Fix Keystore metadata yaml file generation

usmansaleem:
Fix Keystore metadata yaml file generation for key manager import. See #505.


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
```diff
@@ -169,7 +165,8 @@ public void handle(RoutingContext context) {
           final BlsArtifactSigner signer =
               decryptKeystoreAndCreateSigner(jsonKeystoreData, password);
           // 2. write keystore file to disk
-          createKeyStoreYamlFileAt(pubkey, jsonKeystoreData, password);
+          new KeystoreConfigurationFilesCreator(keystorePath, pubkey, jsonKeystoreData)
```
jframe, line 168: We have a KeystoreFileManager so it probably make sense for the create to part of that

usmansaleem: refactored to KeystoreFileManager.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
```diff
@@ -39,6 +41,9 @@
 public class KeystoreFileManager {
```
siladu, line 41: Suggestion: Now this is used by the import flow, should it be pulled up a package from `delete` to `keymanager`?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
```diff
@@ -55,6 +60,39 @@ public void deleteKeystoreFiles(final String pubkey) throws IOException {
...
+   */
+  public void createKeystoreFiles(
+      final String fileNameWithoutExtension, final String jsonKeystoreData, final char[] password)
+      throws IOException {
+    final Path metadataYamlFile = keystorePath.resolve(fileNameWithoutExtension + YAML_EXTENSION);
+    final Path keystoreJsonFile = keystorePath.resolve(fileNameWithoutExtension + JSON_EXTENSION);
+    final Path keystorePasswordFile =
+        keystorePath.resolve(fileNameWithoutExtension + PASSWORD_EXTENSION);
+
+    // create yaml file first (so that if it fails we haven't written password file before it)
```
siladu, line 81: Do you think we also need to cleanup the other files in `removeSignersAndCleanupImportedKeystoreFiles`?
Or at least the keystore json file if e.g. writing the password file fails.

I know it's not directly related to your changes, but your comment made me realise there was actually three files to cleanup not just the yaml file.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
```diff
@@ -55,6 +60,39 @@ public void deleteKeystoreFiles(final String pubkey) throws IOException {
     }
   }
 
+  /**
+   * Create Keystore metadata, json and password files
+   *
+   * @param fileNameWithoutExtension File name, usually public key, without extension.
+   * @param jsonKeystoreData Keystore Json data which will be written to
+   *     fileNameWithoutExtension.json
+   * @param password password char[] which will be written as pubkey.password. All the values in the
```
siladu, line 69: s/pubkey.password/fileNameWithoutExtension.password ?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
siladu: Also really like that this comment explains the _why_ :)

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -26,13 +26,16 @@
...
 import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
 
 public class YamlSignerParser implements SignerParser {
 
   public static final YAMLMapper YAML_MAPPER =
       YAMLMapper.builder()
           .addModule(new SigningMetadataModule())
+          .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)
```
siladu, line 37: Tests don't break when I comment this out

jframe (approved):
LGTM

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/SigningMetadataModule.java
```diff
@@ -36,6 +37,8 @@ public SigningMetadataModule() {
 
     addDeserializer(Bytes.class, new BytesDeserialiser());
     addSerializer(Bytes.class, new BytesSerializer());
+
+    addSerializer(Path.class, new NioPathSerializer());
```
siladu, line 41: NioPathSerializer is already [a built in jackson class](https://fasterxml.github.io/jackson-databind/javadoc/2.13/com/fasterxml/jackson/databind/ext/NioPathSerializer.html) that does the opposite of what you're doing here (i.e. it does `path.toUri().toString()`).

Can use the built in [`ToStringSerializer`](https://fasterxml.github.io/jackson-databind/javadoc/2.13/com/fasterxml/jackson/databind/ser/std/ToStringSerializer.html) I think...
```suggestion
    addSerializer(Path.class, new ToStringSerializer());
```

siladu (commented):
Some suggestions and a question about the password array.

siladu (changes requested):
Marking my previous comments as a blocker in case they're missed now that there's been an approval

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParser.java
jframe: Think it's still cleaner if we don't have this in the file we generate as our docs specify the format without the yaml doc start marker. I think the yaml parser handles it fine though.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
```diff
@@ -55,6 +60,39 @@ public void deleteKeystoreFiles(final String pubkey) throws IOException {
...
+    createYamlFile(metadataYamlFile, data);
+
+    // keystore json file
+    Files.writeString(keystoreJsonFile, jsonKeystoreData, StandardCharsets.UTF_8);
+
+    // password file
+    Files.writeString(keystorePasswordFile, new String(password), StandardCharsets.UTF_8);
+
+    // reset password array argument
+    Arrays.fill(password, ' ');
```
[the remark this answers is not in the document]
usmansaleem: This is mainly a security recommendation ... i.e. the password should not be sitting in memory for longer than needed. But I guess we are already using String to represent password everywhere in our codebase.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
usmansaleem: I didn't realize it was in delete subpackage. Will move it.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/SigningMetadataModule.java
```diff
@@ -81,4 +84,13 @@ public void serialize(
       gen.writeString(value.toString());
     }
   }
+
+  public static class NioPathSerializer extends JsonSerializer<Path> {
+    @Override
+    public void serialize(Path path, JsonGenerator gen, SerializerProvider serializers)
+        throws IOException {
+      // serialize Path without uri (Jackson's implementation uses uri)
+      gen.writeString(path.toString());
```
jframe, line 93: Will this give us the absolute or relative path? I guess relative should be fine as long as it's relative to the keystore directory.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/SigningMetadataModule.java
```diff
@@ -81,4 +84,13 @@ public void serialize(
       gen.writeString(value.toString());
     }
   }
+
+  public static class NioPathSerializer extends JsonSerializer<Path> {
```
jframe, line 88: Name seems a bit odd. But maybe irrelevant if we can use the ToStringSerializer that siladu suggested.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
usmansaleem: good catch ... I also realized I missed the javadoc wording after refactoring it.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParser.java
usmansaleem: Probably because for reading the yaml file it really doesn't matter as long as its a valid yaml file. Having a doc start marker should still work, its just that it does not match our examples. Hence when writing the yaml files, we ought to not write it.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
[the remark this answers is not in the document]
jframe: This removes the password from memory. If we just use a string it could stay in memory for a long time.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/SigningMetadataModule.java
usmansaleem: Thats a good idea.

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParser.java
```diff
@@ -26,13 +26,16 @@
...
+          .disable(YAMLGenerator.Feature.USE_NATIVE_TYPE_ID)
```
jframe, line 38: Why it this needed?

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/SigningMetadataModule.java
usmansaleem: 👍🏼

core/src/main/java/tech/pegasys/web3signer/core/multikey/metadata/parser/YamlSignerParser.java
usmansaleem: Native type id would generate `!<file-keystore>`. Disabling it will generate `type="file-keystore"`

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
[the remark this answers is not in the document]
siladu: Discussed this with @usmansaleem. It's already String elsewhere so if we want to achieve this we probably want a wider PR to address it throughout the codebase.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/delete/KeystoreFileManager.java
```diff
@@ -55,6 +59,37 @@ public void deleteKeystoreFiles(final String pubkey) throws IOException {
...
+      throws IOException {
+    final Path metadataYamlFile =
+        keystorePath.resolve(fileNameWithoutExtension + METADATA_YAML_EXTENSION);
+    final Path keystoreJsonFile =
+        keystorePath.resolve(fileNameWithoutExtension + KEYSTORE_JSON_EXTENSION);
+    final Path keystorePasswordFile =
+        keystorePath.resolve(fileNameWithoutExtension + KEYSTORE_PASSWORD_EXTENSION);
+
+    // create yaml file first (so that if it fails we haven't written password file before it)
+    final FileKeyStoreMetadata data =
```
jframe, line 82: Thinking the create should be atomic and transactional so if any individual file fails to create we clean up by deleting the partially created one. So basically thinking we should add a catch and do the deletes in this function.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/KeystoreFileManager.java
```diff
@@ -55,6 +59,44 @@ public void deleteKeystoreFiles(final String pubkey) throws IOException {
...
+        keystorePath.resolve(fileNameWithoutExtension + METADATA_YAML_EXTENSION);
+    final Path keystoreJsonFile =
+        keystorePath.resolve(fileNameWithoutExtension + KEYSTORE_JSON_EXTENSION);
+    final Path keystorePasswordFile =
+        keystorePath.resolve(fileNameWithoutExtension + KEYSTORE_PASSWORD_EXTENSION);
+
+    final FileKeyStoreMetadata data =
+        new FileKeyStoreMetadata(keystoreJsonFile, keystorePasswordFile, KeyType.BLS);
+    try {
+      // create yaml file first (so that if it fails we haven't written password file before it)
```
siladu, line 85: nit: Is this comment relevant anymore now there's an atomic delete?

siladu (approved):
Looks good, just a nit about a comment

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
```diff
@@ -221,43 +223,25 @@ private void handleInvalidRequest(final RoutingContext routingContext, final Exc
...
-        Files.deleteIfExists(keystorePath.resolve(pubkey + ".yaml"));
-      } catch (Exception e) {
-        LOG.error("Failed to cleanup imported keystore file for key: " + pubkey);
+      } catch (final InterruptedException | ExecutionException e) {
+        LOG.warn("Unable to remove signer for {} due to {}", pubkey, e.getMessage());
       }
+
+      deleteFile(keystorePath.resolve(pubkey + METADATA_YAML_EXTENSION));
+      deleteFile(keystorePath.resolve(pubkey + KEYSTORE_JSON_EXTENSION));
+      deleteFile(keystorePath.resolve(pubkey + KEYSTORE_PASSWORD_EXTENSION));
```
jframe, line 236: Don't think deleting the files is necessary anymore now that the KeyStoreFileManager cleans up on failure

usmansaleem: This might still be required if error happens outside create operation ... since its 'deleteIfExist' it won't cause unnecessary exceptions.

jframe (approved):

Issue #524, opened 2022-03-03
Key Manager Delete - return not_active status when keys have no slashing protection data


siladu:
placeholder

PR #528, opened 2022-03-07
Key Manager Delete API - return not_active status when keys have slashing protection data

siladu:
Following a delete, validators can be registered in the database but still not have signed anything and therefore have no slashing protection data.
When re-deleting such a key, return a status not_found.
In the more typical case where there is slashing protection data, i.e. the key has signed something, return not_active.

Fixes https://github.com/ConsenSys/web3signer/issues/524

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
@@ -67,10 +67,23 @@ public class DeleteKeystoresAcceptanceTest extends KeyManagerTestBase {
   }
 
   @Test
-  public void deletingNonExistingKeyReturnNotFound() throws URISyntaxException {
+  public void deletingExistingKeyWithNoSlashingProtectionDataTwiceReturnsNotFound()
+      throws URISyntaxException {
+    final String pubKey =
+        "0xa46bf94016af71e55ca0518fe6a8bd3852e01b3f959780a4faf3bbe461ac553c0a83f232cc5f2a4b827d8d3455b706e4";
     createBlsKey("eth2/bls_keystore_2.json", "otherpassword");
     setupSignerWithKeyManagerApi(true);
-    callDeleteKeystores(composeRequestBody())
+    callDeleteKeystores(composeRequestBody(pubKey))
+        .then()
+        .contentType(ContentType.JSON)
+        .assertThat()
+        .statusCode(200)
+        .body("data[0].status", is("deleted"))
+        .and()
+        .body("slashing_protection", is(emptySlashingData));
+
+    // call API again with same key should return not_found
+    callDeleteKeystores(composeRequestBody(pubKey))
         .then()
         .contentType(ContentType.JSON)
         .assertThat()
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtection.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtection.java
@@ -34,7 +34,7 @@ public interface SlashingProtection {
 
   void registerValidators(List<Bytes> validators);
 
-  boolean isRegisteredValidator(Bytes publicKey);
+  boolean slashingProtectionDataExistsFor(Bytes publicKey);
 
   void export(OutputStream output);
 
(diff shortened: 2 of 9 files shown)


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
```diff
@@ -67,10 +67,23 @@ public void invalidRequestBodyReturnsError() throws URISyntaxException {
   }
```
jframe, line 68: What about an AT showing that 2nd delete returns not_active if there is slashing protection data?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDao.java
```diff
@@ -64,4 +65,26 @@ public boolean isEnabled(final Handle handle, final int validatorId) {
         .mapTo(Boolean.class)
         .first();
   }
+
+  public boolean hasSigned(final Handle handle, final Bytes publicKey) {
+
+    Optional<Integer> maybeValidatorId =
```
jframe, line 71: This shouldn't be necessary to look up the validatorId as we have the validator id in the slashing protection service and can pass that in

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java
siladu: That test already exists: https://github.com/ConsenSys/web3signer/blob/dba2a0614cb81080d0180a4b221348500ebd0a23/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/keymanager/DeleteKeystoresAcceptanceTest.java#L97-L120

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDao.java
siladu: done 👍

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDao.java
```diff
@@ -64,4 +64,15 @@ public boolean isEnabled(final Handle handle, final int validatorId) {
         .mapTo(Boolean.class)
         .first();
   }
+
+  public boolean hasSigned(final Handle handle, final int validatorId) {
+    return handle
+        .createQuery(
+            "SELECT EXISTS(SELECT 1 FROM SIGNED_ATTESTATIONS WHERE validator_id = ?)"
+                + " OR EXISTS(SELECT 1 FROM SIGNED_BLOCKS WHERE validator_id = ?)")
+        .bind(0, validatorId)
```
jframe, line 73: You can simplify this to just bind the one value by using a named binding https://jdbi.org/#_binding_arguments

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtection.java
```diff
@@ -34,7 +34,7 @@ boolean maySignBlock(
 
   void registerValidators(List<Bytes> validators);
 
-  boolean isRegisteredValidator(Bytes publicKey);
+  boolean slashingProtectionDataExistsFor(Bytes publicKey);
```
jframe, line 37: nit: hasSlashingProtectionData would make clearer we are returning a boolean

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDao.java
siladu: done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtection.java
siladu: @jframe I've gone for hasSlashingProtectionDataFor as a compromise. 

Agree that hasX fits in with convention, however I also wanted to make clear that it was the passed in public key that has the data. This was to distinguish from the optional slashingProtection object in the expression...
```
final boolean slashingProtectionDataExistsForPubKey =
              slashingProtection
                  .map(sp -> sp.hasSlashingProtectionDataFor(Bytes.fromHexString(pubkey)))
                  .orElse(false);
```

Overall, this expression is saying "we have slashingProtection enabled and have slashing protection data for this key".

For the same reason, I have left the tightly scoped boolean `slashingProtectionDataExistsForPubKey` unchanged.

jframe (approved):
LGTM

georgep9 (approved):
LGTM!

PR #529, opened 2022-03-07
Key manager delete to use a streaming approach

jframe:
Delete and export the slashing protection data one a time instead of exporting all the slashing protection data at once. This will allows for an individual slashing protection data record to fail and also makes it easier to rollback database changes when we add the disable key functionality.


siladu (approved):
LGTM

PR #532, opened 2022-03-15
Filling a gap in the test coverage to ensure IOException will bubble up

siladu:
Relates to Key Manager Delete API, specifically exporting slashing protection data. This test is ensuring that an IOException during the increment export of an individual key is bubbled up.

DeleteKeystoresProcessorTest already covers the error handling once this exception has bubbled up.


jframe (approved):
LGTM

PR #535, opened 2022-03-16
Adding ip and net utils in docker image

usmansaleem:
-- these utilities can be used in k8 or docker-compose environment to wait for other dependent services


jframe (approved):
LGTM

PR #536, opened 2022-03-17
Inline the metric categories

jframe:
Move Web3SignerMetricCategory to common and inline the metric categories. This makes it easier to add new metrics categories and use in the w3s code.


siladu (approved):
LGTM

PR #541, opened 2022-03-26
Cacheable AWS Secrets Manager

### LINKED ISSUE #531: Cache AWS client when for key loading
When loading AWS keys from the secrets manager we are creating the AWS client each time. This makes loading of the keys slower than if we are able to use the same client.

**Tasks**
* Reuse the same AwsSecretsManager for each key that uses the same connection details

georgep9:
## PR Description
Updated AwsSecretsManager factory to use [new provider class](https://github.com/ConsenSys/signers/commit/7de7e6cb2be4e48833f73624cb0a4a55862336fa) for creating/serving cacheable managers.

## Fixed Issue(s)
Fixes #531

## Documentation

## Changelog


georgep9:
May need to implement unit tests to validate whether;
- initialized `AwsSecretsManagerProvider` has set cache `maximumSize` equal to cli option `--aws-cache-maximum-size`
- default size 1 is set when `--aws-cache-maximum-size` not given

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -180,6 +180,11 @@
       description = "Enable access logs (default: ${DEFAULT-VALUE})")
   private final Boolean accessLogsEnabled = false;
 
+  @Option(
+      names = {"--aws-cache-maximum-size"},
+      description = "Maximum cache size for AWS Secrets Manager (default: ${DEFAULT-VALUE})")
```
jframe, line 185: Can we add a paramLabel too

jframe, line 185: Think description needs to mention that this to cache the connections. When I read this as it is it sounds like we are caching the secrets.
Perhaps something like "Maximum number of connections to cache to the AWS secrets manager."

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -180,6 +180,11 @@
       description = "Enable access logs (default: ${DEFAULT-VALUE})")
   private final Boolean accessLogsEnabled = false;
 
+  @Option(
+      names = {"--aws-cache-maximum-size"},
```
jframe, line 184: Maybe, "--aws-connection-cache-size"?

siladu (approved):
LGTM

jframe (approved):

Issue #554, opened 2022-04-27
Docker image vulnerabilities in 22.4.0


jmcruz1983:
Running docker scan, 
some high vulnerabilities are present in the new release,

```
docker scan --severity high consensys/web3signer:22.4.0 

Testing consensys/web3signer:22.4.0...

Organization:     <REDACTED>
Package manager:   deb
Project name:      docker-image|consensys/web3signer
Docker image:      consensys/web3signer:22.4.0
Platform:          linux/amd64
Base image:        ubuntu:22.04
Licenses:          enabled

✔ Tested 121 dependencies for known issues, no vulnerable paths found.

According to our scan, you are currently using the most secure version of the selected base image

-------------------------------------------------------

Testing consensys/web3signer:22.4.0...

Tested 205 dependencies for known issues, found 3 issues.


Issues to fix by upgrading:

  Upgrade com.fasterxml.jackson.core:jackson-databind@2.13.1 to com.fasterxml.jackson.core:jackson-databind@2.13.2.1 to fix
  ✗ Denial of Service (DoS) [High Severity][https://snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-2421244] in com.fasterxml.jackson.core:jackson-databind@2.13.1
    introduced by com.fasterxml.jackson.core:jackson-databind@2.13.1

  Upgrade com.github.jnr:jnr-posix@3.0.47 to com.github.jnr:jnr-posix@3.1.8 to fix
  ✗ Use After Free [High Severity][https://snyk.io/vuln/SNYK-JAVA-COMGITHUBJNR-1570422] in com.github.jnr:jnr-posix@3.0.47
    introduced by com.github.jnr:jnr-posix@3.0.47

  Upgrade com.google.protobuf:protobuf-java@3.17.2 to com.google.protobuf:protobuf-java@3.18.2 to fix
  ✗ Denial of Service (DoS) [High Severity][https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEPROTOBUF-2331703] in com.google.protobuf:protobuf-java@3.17.2
    introduced by com.google.protobuf:protobuf-java@3.17.2



Organization:      <REDACTED>
Package manager:   maven
Target file:       /opt/web3signer/lib
Project name:      consensys/web3signer:22.4.0:/opt/web3signer/lib
Docker image:      consensys/web3signer:22.4.0
Licenses:          enabled

-------------------------------------------------------

[...]

PR #556, opened 2022-04-28
Update various dependencies

siladu:
Fixes https://github.com/ConsenSys/web3signer/issues/554

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -13,7 +13,7 @@
 
 dependencyManagement {
   dependencies {
-    dependency 'com.fasterxml.jackson.core:jackson-databind:2.13.1'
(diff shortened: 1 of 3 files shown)


gradle/versions.gradle
```diff
@@ -13,7 +13,7 @@
 
 dependencyManagement {
   dependencies {
-    dependency 'com.fasterxml.jackson.core:jackson-databind:2.13.1'
+    dependency 'com.fasterxml.jackson.core:jackson-databind:2.13.2.2'
```
jframe, line 16: Can use a dependency set and also update the other jackson library versions as well i.e. the dataformat-yml package

jframe (approved):
LGTM

PR #563, opened 2022-05-10
Fix AWS environment mode not supporting web identity token

jframe:
AWS environment mode is not working with web identity token due to the sts library missing. This PR updates to the latest version of the signers library which includes the sts library.


usmansaleem (approved):
LGTM

siladu (approved):
LGTM

PR #564, opened 2022-05-10
Fix AWS ATs failing with NoClassDefFoundError

jframe:
Fix AWS ATs failing with java.lang.NoClassDefFoundError. This was caused by conflicting versions of the AWS SDK for the ATs.


usmansaleem (approved):
LGTM

PR #569, opened 2022-05-17
Keystore bulk loading

jframe:
Keystore bulk loading. Config takes seperate keystore and password directories keeping compatibility with the Teku approach.

This adds 3 new command arguments for the keystore bulk loading on the eth2 subcommand:
* --keystores-path The path to a directory storing Eth2 keystore
* --keystores-passwords-path The path to a directory with the corresponding passwords file for the keystores.
* --keystores-password-file The path to a file that contains the password that all keystores use.

## Documentation

## Changelog


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsDefaultImpl.java
```diff
@@ -97,6 +98,19 @@ public List<String> createCmdLineParams() {
         params.add("--azure-tenant-id");
         params.add(azureParams.getTenantId());
       }
+      if (signerConfig.getKeystoresParameters().isPresent()) {
```
usmansaleem, line 101: Don't forget to add same to `tech.pegasys.web3signer.dsl.signer.runner.CmdLineParamsConfigFileImpl` as well. AT CLI DSL uses both config and and standard cli options.

signing/src/main/java/tech/pegasys/web3signer/signing/BlsBKeystoreBulkLoader.java
```diff
@@ -0,0 +1,102 @@
...
+        .collect(Collectors.toList());
+  }
+
+  public Collection<ArtifactSigner> loadKeystoresUsingPasswordFile(
+      final Path keystoresDirectory, final Path passwordFile) {
+    final List<Path> keystoreFiles = keystoreFiles(keystoresDirectory);
+    final String password;
+    try {
+      password = Files.readString(passwordFile);
+    } catch (IOException e) {
```
usmansaleem, line 60: nit: final

signing/src/main/java/tech/pegasys/web3signer/signing/BlsBKeystoreBulkLoader.java
```diff
@@ -0,0 +1,102 @@
...
+      throw new IllegalStateException("Unable to read the password file", e);
```
usmansaleem, line 61: nit: You can instead wrap in `UncheckedIOException` instead of `IllegalStateException`.

signing/src/testFixtures/java/tech/pegasys/web3signer/KeystoreUtil.java
```diff
@@ -0,0 +1,71 @@
...
+      final String password) {
+    createKeystoreFile(keyPair, keystoreDir, password);
+    createKeystorePasswordFile(keyPair, passwordDir, password);
+  }
+
+  public static void createKeystorePasswordFile(
+      final BLSKeyPair keyPair, final Path passwordDir, final String password) {
+    try {
+      Files.writeString(passwordDir.resolve(keyPair.getPublicKey().toString() + ".txt"), password);
+    } catch (IOException e) {
```
usmansaleem, line 51: nit: final. and UncheckedIOException.

usmansaleem (changes requested):
DSL configfileimpl. All other changes looks good to me.

usmansaleem (commented):
LGTM

signing/src/main/java/tech/pegasys/web3signer/signing/config/KeystoresParameters.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.nio.file.Path;
+
+public interface KeystoresParameters {
+
+  Path getKeystoresPath();
+
+  Path getKeystoresPasswordsPath();
+
+  Path getKeystoresPasswordFile();
```
siladu, line 23: Since teku compatibility is a goal of this PR, I'm wondering if there's a further benefit to making it even closer to how this works https://docs.teku.consensys.net/en/latest/Reference/CLI/CLI-Syntax/#validator-keys

i.e. --validator-keys="<keystoresPath>:<keystoresPasswordsPath>"
The naming and similar structure may be easier for users to translate if they're already used to teku.

commandline/src/main/java/tech/pegasys/web3signer/commandline/config/PicoKeystoresParameters.java
```diff
@@ -0,0 +1,61 @@
...
+      description = "The path to a directory storing Eth2 keystores",
+      paramLabel = DefaultCommandValues.MANDATORY_PATH_FORMAT_HELP)
+  private Path keystoresPath;
+
+  @Option(
+      names = {"--keystores-passwords-path"},
+      description =
+          "The path to a directory with the corresponding passwords file for the keystores."
+              + " Filename for the password without the extension must match the keystore filename."
+              + " This cannot be set if --keystores-password-file is also specified.",
```
siladu, line 35: There's no mention of the files needing to end in txt

siladu: I guess the same goes for the other options too (presumably json is requirement for keystores-path?)

signing/src/test/java/tech/pegasys/web3signer/signing/BlsBKeystoreBulkLoaderTest.java
```diff
@@ -0,0 +1,144 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collection;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class BlsBKeystoreBulkLoaderTest {
```
siladu, line 31: These tests take 13 secs on my machine. Will the tests be significantly faster if reuse the @TempDirs or is the time crypto/keystore related?

signing/src/main/java/tech/pegasys/web3signer/signing/BlsBKeystoreBulkLoader.java
```diff
@@ -0,0 +1,103 @@
...
+  public Collection<ArtifactSigner> loadKeystoresUsingPasswordDir(
+      final Path keystoresDirectory, final Path passwordsDirectory) {
+    final List<Path> keystoreFiles = keystoreFiles(keystoresDirectory);
+    return keystoreFiles.parallelStream()
+        .map(
+            keystoreFile ->
+                createSignerForKeystore(
+                    keystoreFile,
+                    key -> Files.readString(passwordsDirectory.resolve(key + ".txt"))))
+        .flatMap(Optional::stream)
```
siladu, line 51: nice!

signing/src/main/java/tech/pegasys/web3signer/signing/BlsBKeystoreBulkLoader.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.commons.io.FilenameUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+
+public class BlsBKeystoreBulkLoader {
```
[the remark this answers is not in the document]
jframe: typo :)

signing/src/main/java/tech/pegasys/web3signer/signing/config/KeystoresParameters.java
jframe: I don't think needs to exactly the same as the Teku implementation just compatible and inspired by Teku as in we can pass in a directory of keystores and keystore passwords. The feature should also make sense in how Web3Signer works as well.

The approach I've taken follows the bulk loading approach we have done elsewhere. Combining the bulk loading with individual key configuration doesn't make sense for web3signer. We have the existing config files for this kind of configuration. So I only included the directory part.

As far as command line options go we unfortunately cannot have something as generic as what done with Teku as we have multiple key sources so --validator-keys is not clear for web3signer what source they are coming from which why have taken the same naming approach which we took with Azure bulk loading.

commandline/src/main/java/tech/pegasys/web3signer/commandline/config/PicoKeystoresParameters.java
jframe: Good point. Keystore files must have a json extension and the password files a txt extension. However I wonder how important that restriction is? If we ensure that the directories were different which is a sensible security measure then the file extensions don't matter at all.

signing/src/test/java/tech/pegasys/web3signer/signing/BlsBKeystoreBulkLoaderTest.java
jframe: Takes 4s on my machine which I think is ok. It's probably the decrypt taking a long time?

jframe: Since this is only for tests we can make it more insecure as well. Perhaps use the settings that eth2-val-tools uses in their insecure mode?

jframe: Reduced the number of pbkdf iterations to 2 as this is a test. Only takes 1.5s on my machine now. Hopefully it similarly faster on your machine now.

commandline/src/main/java/tech/pegasys/web3signer/commandline/config/PicoKeystoresParameters.java
siladu: True. 

Couple of concerns with loading any file extension, if that's what you're suggesting...

1. Any non-keystore files will cause errors and be ignored which is fine, but potentially a lot of wasted processing.

2. What happens if there are duplicates files with only differing extensions? I think this could be a problem, especially in the password directory.

jframe: Could find a way of dealing with that. But I prefer the simplicity of using .json extension for the keystore and a .txt extension for the password. Added those details in the CLI docs.

siladu (approved):
LGTM

usmansaleem (approved):
LGTM

PR #580, opened 2022-05-26
update to latest teku libraries to support ropsten

jframe:
## PR Description
Add support for the ropsten beacon chain network

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Changelog
 
 ## Next version
+### Breaking Changes
+- Removed network definition for kintsugi testnet
+
 ### Features Added
 - Eth2 keystore bulk loading allowing a directory of keystores to be loaded without config files
-- Updated Teku libraries to provide support for `ropsten` network
+- Added support for ropsten testnet
 
 ## 22.5.0
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next version
 ### Features Added
 - Eth2 keystore bulk loading allowing a directory of keystores to be loaded without config files
+- Updated Teku libraries to provide support for `ropsten` network
 
 ## 22.5.0
 ### Breaking Changes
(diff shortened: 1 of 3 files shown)


commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -51,7 +51,7 @@ public class Eth2SubCommand extends ModeSubCommand {
       names = {"--network"},
       paramLabel = "<NETWORK>",
       description =
-          "Predefined network configuration to use. Possible values: [mainnet, pyrmont, prater, kintsugi, kiln, gnosis, minimal], file path"
+          "Predefined network configuration to use. Possible values: [mainnet, pyrmont, prater, kintsugi, kiln, gnosis, minimal, ropsten], file path"
```
usmansaleem, line 54: I wonder if we should create a String constant from Teku's Eth2Network enum and use that in description. That would free us from modifying it every time we support a new network.

usmansaleem (approved):
LGTM

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
siladu, line 54: Could also remove kintsugi to keep in line with teku: https://github.com/ConsenSys/teku/commit/b329dc05bce59c3cc484f260013ac584d71ebe40

jframe: Another reason to use the enum somehow

jframe: Good idea. I'll try and work how to use that with pico

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Next version
 ### Features Added
 - Eth2 keystore bulk loading allowing a directory of keystores to be loaded without config files
+- Updated Teku libraries to provide support for `ropsten` network
```
siladu, line 6: Worth mentioning kintsugi support is removed here too?

siladu (approved):
LGTM

Issue #561, opened 2022-05-04
db query duration metric


jmcruz1983:
Currently web3signer offers `signing_bls_signing_duration` metric with percentiles,
but this metric is the total time spent on signing operation without detailing how much it takes on postgres db query 
when running slashing protection with postgres db,

As feature request, could be offered a new metrics that shows the time spend on db query with percentiles?
Thanks

**Tasks**
* Determine what db queries we want to have timing for
* Expose metrics for these

PR #587, opened 2022-06-02
Add timer metric for total database duration

siladu:
Fixes https://github.com/ConsenSys/web3signer/issues/561

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/metrics/SlashingProtectionMetrics.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/metrics/SlashingProtectionMetrics.java
@@ -17,7 +17,6 @@ import static tech.pegasys.web3signer.common.Web3SignerMetricCategory.ETH2_SLASH
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import org.hyperledger.besu.plugin.services.metrics.Counter;
 import org.hyperledger.besu.plugin.services.metrics.OperationTimer;
-import tech.pegasys.web3signer.common.Web3SignerMetricCategory;
 
 public class SlashingProtectionMetrics {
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/tech/pegasys/web3signer/core/metrics/SlashingProtectionMetrics.java
```diff
@@ -34,6 +36,12 @@ public SlashingProtectionMetrics(final MetricsSystem metricsSystem) {
             ETH2_SLASHING_PROTECTION,
             "prevented_signings",
             "The number of prevented signings due to violation of slashing conditions.");
+
+    this.databaseTimer =
+        metricsSystem.createTimer(
+            ETH2_SLASHING_PROTECTION,
+            "slashing_protection_database_duration",
```
jframe, line 43: slashing_protection prefix can be removed. The metrics name is appended with the metric category so will already have a "eth2_slashingprotection" prefix

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/Eth2SignForIdentifierHandler.java
```diff
@@ -149,7 +149,7 @@ private void signWithSlashingProtection(
       final Eth2SigningRequestBody eth2SigningRequestBody,
       final Bytes signingRoot,
       final String signature) {
-    try {
+    try (final TimingContext ignored = slashingMetrics.getDatabaseTimer().startTimer()) {
```
jframe, line 152: Would be better to put this in the DbSlashingProtection class so it's as close as possible to the time taken for the db queries.

core/src/main/java/tech/pegasys/web3signer/core/metrics/SlashingProtectionMetrics.java
```diff
@@ -34,6 +36,12 @@ public SlashingProtectionMetrics(final MetricsSystem metricsSystem) {
             ETH2_SLASHING_PROTECTION,
             "prevented_signings",
             "The number of prevented signings due to violation of slashing conditions.");
+
+    this.databaseTimer =
+        metricsSystem.createTimer(
```
jframe, line 41: Would be good to create a timer with a label so we can database queries broken down into attestation and block duration times.

siladu:
Put the metrics as close to the DB as I could without crossing into the slashing module, which would require a larger refactor with the only practical benefit being excluding one non-DB related code...although I appreciate that could change in the future so maybe the inside the slashing module is still the right place.

jframe (approved):
LGTM

PR #596, opened 2022-06-24
Register validator support

### LINKED ISSUE #577: Support register validator api endpoint
https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/register_validator.yaml
https://github.com/ethereum/consensus-specs/pull/2884

Note, we do not need to implement this API. Teku will send sign request

Tasks
- [ ] Add new signing type (`ArtifactType`)
- [ ] Add to `Eth2SigningRequestBody` (entrypoint)
- [ ] Update data models
- [ ] Update OpenAPI schema
- [ ] Process request to perform signing
- [ ] Integration tests
est. 2
- [ ] Acceptance Tests to verify OpenAPI
- [ ] Communicate API changes to unblock validator client teams
est. 1

To be estimated separately:
- [ ] https://github.com/ConsenSys/teku/issues/5644 - Wire up the request from teku: https://github.com/ConsenSys/teku/blob/78013c1b0520dfdbfcb8abfde7c69b5841993d38/validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/ExternalSigner.java#L260

wcgcyx:
## PR Description
This PR adds support for register validator endpoint.

## Fixed Issue(s)
fixes #577


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/eth2/RegisterValidatorMessage.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class RegisterValidatorMessage {
+  private final ValidatorRegistration validatorRegistration;
+
+  private final UInt64 epoch;
+
+  @JsonCreator
+  public RegisterValidatorMessage(
```
siladu, line 26: Again similar point to before, wondering if this could just be RegisterValidator

siladu (commented):
One question about naming, otherwise LGTM. I'd rather leave Jason or Uzi to approve as I'm not too familiar with the openAPI spec for this.

core/src/main/java/tech/pegasys/web3signer/core/service/http/ArtifactType.java
```diff
@@ -23,5 +23,6 @@ public enum ArtifactType {
   VOLUNTARY_EXIT,
   SYNC_COMMITTEE_MESSAGE,
   SYNC_COMMITTEE_SELECTION_PROOF,
-  SYNC_COMMITTEE_CONTRIBUTION_AND_PROOF
+  SYNC_COMMITTEE_CONTRIBUTION_AND_PROOF,
+  REGISTER_VALIDATOR_MESSAGE
```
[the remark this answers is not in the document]
wcgcyx: Right, it could just be register validator. It will be more concise in this way. I will update the PR.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/Eth2RequestUtils.java
```diff
@@ -473,7 +486,38 @@ private static Eth2SigningRequestBody createSyncCommitteeContributionAndProofReq
...
+
+  private static Eth2SigningRequestBody createRegisterValidatorRequest() {
+    final UInt64 epoch = dataStructureUtil.randomEpoch();
+    final ValidatorRegistration validatorRegistration =
+        new ValidatorRegistration(
+            dataStructureUtil.randomBytes20(),
+            dataStructureUtil.randomUInt64(),
+            dataStructureUtil.randomUInt64(),
+            BLSPubKey.fromHexString(
+                "0x8f82597c919c056571a05dfe83e6a7d32acf9ad8931be04d11384e95468cd68b40129864ae12745f774654bbac09b057"));
```
jframe, line 501: nit: be nice to create a constant for this public key since we use in at least two create request methods

core/src/main/java/tech/pegasys/web3signer/core/service/http/ArtifactType.java
jframe, line 27: Unsure of the name, the type in Teku is ValidatorRegistration. Would VALIDATOR_REGISTRATION or VALIDATOR_REGISTRATION_MESSAGE be more correct?

wcgcyx: Maybe register validator would be better? since the request consists both a validator registration and an epoch.

jframe: 🤷 Not sure. That's inconsistent with the name used in Teku. Would rather it was consistent everywhere.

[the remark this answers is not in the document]
jframe: Agree the message suffix isn't needed

jframe: Looking at the spec, the message type is ValidatorRegistration so would rather is was consistent with that. We do add the epoch as well but don't think that matters.

wcgcyx: Naming updated

siladu (approved):
LGTM 👍

core/src/main/resources/openapi-specs/eth2/signing/schemas.yaml
```diff
@@ -135,7 +135,20 @@ components:
...
+    ValidatorRegistrationSigning:
+      allOf:
+        - type: object
+          properties:
+            type:
+              type: "string"
+            signingRoot:
+              description: 'signing root for optional verification if field present'
+              type: "string"
+            validator_registration_request:
```
jframe, line 147: Don't think we need this additional level of nesting in the request data. For other requests we have just treat the top level as the message. So would have expected to have epoch and signing root at the top level without the additional intermediary validator_registration_request object.

jframe (changes requested):

usmansaleem (approved):
lgtm

jframe (approved):
LGTM

PR #601, opened 2022-07-04
Release version metric

### LINKED ISSUE #480: W3S Version in Metrics
Web3Signer version can currently be obtained via web3signer -V 

 teku expose their version via a prometheus metric
 
 `beacon_teku_version{endpoint="metrics", version="21.12.1"} = 1`
 
 This metric is useful for displaying version number on a grafana dashboard, can we include an equivalent metric for the exposed w3s metrics ie
 
  `web3signer_version{version="21.10.5"} = 1`
  
  cc @usmansaleem (following on from discussion in discord)

gfukushima:
## PR Description
Exposing web3singer_version metric so it can be displayed on a grafana dashboard.

## Fixed Issue(s)

Fixes #480

## Documentation

## Changelog

### DIFF
--- a/common/src/main/java/tech/pegasys/web3signer/common/ApplicationInfo.java
+++ b/common/src/main/java/tech/pegasys/web3signer/common/ApplicationInfo.java
@@ -10,7 +10,7 @@
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
  */
-package tech.pegasys.web3signer.commandline;
+package tech.pegasys.web3signer.common;
 
 public class ApplicationInfo {
   private static final String CLIENT_IDENTITY = "web3signer";
--- a/common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
+++ b/common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
@@ -24,7 +24,8 @@ public enum Web3SignerMetricCategory implements MetricCategory {
   ETH2_SLASHING_PROTECTION("eth2_slashingprotection"),
   FILECOIN("filecoin"),
   HTTP("http"),
-  SIGNING("signing");
+  SIGNING("signing"),
+  WEB3SIGNER("web3signer");
 
   private final String name;
 
--- a/common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
+++ b/common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
@@ -24,8 +24,7 @@ public enum Web3SignerMetricCategory implements MetricCategory {
   ETH2_SLASHING_PROTECTION("eth2_slashingprotection"),
   FILECOIN("filecoin"),
   HTTP("http"),
-  SIGNING("signing"),
-  WEB3SIGNER("web3signer");
+  SIGNING("signing");
 
   private final String name;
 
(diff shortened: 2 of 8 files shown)


common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
```diff
@@ -24,7 +24,8 @@ public enum Web3SignerMetricCategory implements MetricCategory {
   ETH2_SLASHING_PROTECTION("eth2_slashingprotection"),
   FILECOIN("filecoin"),
   HTTP("http"),
-  SIGNING("signing");
+  SIGNING("signing"),
+  WEB3SIGNER("web3signer");
```
siladu, line 28: Could use StandardMetricCategory.PROCESS instead of creating a new one?

common/src/main/java/tech/pegasys/web3signer/common/ApplicationInfo.java
```diff
@@ -10,7 +10,7 @@
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
  */
-package tech.pegasys.web3signer.commandline;
+package tech.pegasys.web3signer.common;
```
siladu, line 13: Seems right to move this out of commandline. Wondering if it should be in core rather than common though?

common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
gfukushima: Agree, although I've created a new one to meet the requirements of the issue which named the metric using the  "web3signer" prefix. Could easily adjust to use StandardMetricCategory.PROCESS instead.

common/src/main/java/tech/pegasys/web3signer/common/ApplicationInfo.java
gfukushima: Core sounds like a better package to have the ApplicationInfo class although the PlatformDetector class that has been moved as well but it doesn't seem appropriate to me to move it to core.
Maybe move both under the "util" directory of the "core" package?

siladu: Looks like PlatformDetector is only used by ApplicationInfo so probably makes sense to keep those together. 

I just noticed ApplicationInfo is shared between app and core, so common is fine too. The app module is a bit of an odd one, wonder if that should be rolled into core at some point?

common/src/main/java/tech/pegasys/web3signer/common/Web3SignerMetricCategory.java
siladu: Not sure if there's a standard naming convention for this...if there's not then I go with whatever is going to be easier for people to find, which might well be web3signer_version :)

siladu (approved):
couple of nitpicks being discussed but no blockers

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -172,6 +175,13 @@ public void run() {
     }
   }
 
+  private void createVersionMetric(MetricsSystem metricsSystem) {
+    metricsSystem
+        .createLabelledGauge(
+            Web3SignerMetricCategory.WEB3SIGNER, "version", "Release information", "version")
```
jframe, line 181: Not sure why we need a labelled gauge. But we are taking that approach then then I think the metric should be release with the labelled metric of version.
i.e. 
`metricsSystem
        .createLabelledGauge(
            StandardMetricCategory.PROCESS, "release", "Release information", "version")
        .labels(() -> 1, ApplicationInfo.version();`

jframe (approved):

jframe (commented):
Can you also add a change log entry for this

siladu (approved):
LGTM

PR #610, opened 2022-07-11
Add health check endpoint

### LINKED ISSUE #538: Liveness endpoint
Add a liveliness endpoint to Web3Signer so that infrastructure can know when web3signer is up and running and able to accept requests. This would it easier for users that have Web3Signer in kubernetes and simplify their deployement.

**Tasks**
* Investigate format for the health and liveliness endpoints
* Determine what endpoints we should have to integrate with kubernetes and to allow for easier deployment

wcgcyx:
## PR Description
This PR adds a health check endpoint to check if w3s is healthy. This PR also updates the existing up check handler, if w3s fails to load signers from any resource, up check will return FAIL.

## Expected behaviour
~~1. On startup, if w3s fails to load signers from one or more of the resources (e.g. AWS). Up check endpoint will return a response of FAIL with status to be 500.~~
2. Under eth2 mode with slashing protection enabled, w3s will periodically check if db is healthy and reachable. If w3s is healthy, health check returns a response with status 200. If w3s is unhealthy, health check returns a response with status 503.

## Fixed Issue(s)
fixes #538 

## Documentation

## Changelog


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/HealthcheckHandler.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public void handle(final RoutingContext routingContext) {
+    if (slashingProtectionContext.isPresent()) {
+      try (Handle ignored = slashingProtectionContext.get().getSlashingProtectionJdbi().open()) {
+      } catch (final Exception e) {
+        routingContext
+            .response()
+            .setStatusCode(500)
+            .putHeader(CONTENT_TYPE, TEXT_PLAIN_UTF_8)
+            .end("Lost connection to db");
```
siladu, line 43: Might be good to specify "slashing protection db" here.

core/src/main/java/tech/pegasys/web3signer/core/Eth1Runner.java
```diff
@@ -74,6 +76,10 @@ protected Router populateRouter(final Context context) {
         .failureHandler(errorHandler);
 
     addReloadHandler(routerBuilder, signerProvider, RELOAD.name(), context.getErrorHandler());
+    registerHealthCheck(
+        context.getRouterBuilder(),
+        context.getErrorHandler(),
+        new HealthcheckHandler(Optional.empty()));
```
jframe, line 82: Is there any useful check that be done for eth1 to ensure it is able to sign?

core/src/main/java/tech/pegasys/web3signer/core/FilecoinRunner.java
```diff
@@ -71,6 +73,10 @@ protected Router populateRouter(final Context context) {
         context.getArtifactSignerProvider(),
         RELOAD.name(),
         context.getErrorHandler());
+    registerHealthCheck(
+        context.getRouterBuilder(),
+        context.getErrorHandler(),
+        new HealthcheckHandler(Optional.empty()));
```
jframe, line 79: Similar concern to the eth1 healthpoint. Is there anything useful this does for filecoin. Think this is just a upcheck without the db health check?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/HealthcheckHandler.java
```diff
@@ -0,0 +1,49 @@
...
+  private final Optional<SlashingProtectionContext> slashingProtectionContext;
+
+  public HealthcheckHandler(final Optional<SlashingProtectionContext> slashingProtectionContext) {
+    this.slashingProtectionContext = slashingProtectionContext;
+  }
+
+  @Override
+  public void handle(final RoutingContext routingContext) {
+    if (slashingProtectionContext.isPresent()) {
+      try (Handle ignored = slashingProtectionContext.get().getSlashingProtectionJdbi().open()) {
```
jframe, line 37: Think this is creating a potential DOS vector for the database by attempting to open a new connection each time, a better way might be periodically that the connection works on a timer and just check that status from this handler.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/HealthcheckHandler.java
```diff
@@ -0,0 +1,49 @@
...
+      } catch (final Exception e) {
```
jframe, line 38: I opening a connection sufficient to test that the DB connection works? Normally I've a simple query like `SELECT 1` to confirm that the database is healthy.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/HealthcheckHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+import tech.pegasys.web3signer.slashingprotection.SlashingProtectionContext;
+
+import java.util.Optional;
+
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
+import org.jdbi.v3.core.Handle;
+
+public class HealthcheckHandler implements Handler<RoutingContext> {
```
jframe, line 26: Are the vertx health checks https://vertx.io/docs/vertx-health-check/java/ of any use instead of rolling our own?

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -103,11 +108,12 @@ public void run() {
     try {
       createVersionMetric(metricsSystem);
       metricsEndpoint.start(vertx);
-
+      boolean failToLoad = false;
```
siladu, line 111: rename suggestion: failToLoad -> failedToLoadAnySigners (or is it failedToLoadSomeSigners?)

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/UpcheckHandler.java
```diff
@@ -19,8 +19,23 @@
 import io.vertx.ext.web.RoutingContext;
 
 public class UpcheckHandler implements Handler<RoutingContext> {
+
+  private final boolean failToLoad;
```
siladu, line 23: As above, failedToLoadAnySigners

siladu: If the service is up but the signers failed to load, is there any mitigation via RPC or is the only recourse to restart the service?

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
wcgcyx: failed to load some signers maybe? as long as there is an error when loading from one source, it will be set to true.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/UpcheckHandler.java
wcgcyx: I think with this we just have a way to tell if w3s fails to load signers from one of the sources and if yes, we can restart it.

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -140,6 +146,45 @@ public Router populateRouter(final Context context) {
...
+                  future.get(3000, TimeUnit.MILLISECONDS);
+                  isDbDown = false;
+                } catch (Exception e) {
+                  // Have exception in database health check (timeout or error).
+                  isDbDown = true;
+                } finally {
+                  future.cancel(true);
+                }
+              },
+              3000,
```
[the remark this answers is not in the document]
wcgcyx: Configuration added

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -140,6 +146,41 @@ public Router populateRouter(final Context context) {
...
+              3000,
+              TimeUnit.MILLISECONDS);
+
+      super.registerHealthCheckProcedure(
+          "slashing-protection-db-health-check",
+          promise -> {
+            promise.complete(isDbDown ? Status.KO() : Status.OK());
+          });
```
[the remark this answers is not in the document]
wcgcyx: Moved into a class under slashing protection db

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -94,6 +97,8 @@ public class Eth2Runner extends Runner {
   private final Spec eth2Spec;
   private final boolean isKeyManagerApiEnabled;
 
+  private boolean isDbDown;
```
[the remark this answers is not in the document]
wcgcyx: Thought that there is only one process writing to the variable but yes an atomic boolean is better. I've updated it.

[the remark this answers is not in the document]
wcgcyx: Done

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
wcgcyx: Updated.

siladu:
@wcgcyx I think this PR warrants a changelog entry and the doc-change-required label, at least for the new CLI options

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+                          Executors.newCachedThreadPool()
+                              .submit(
+                                  () ->
+                                      protectionContext
+                                          .getSlashingProtectionJdbi()
+                                          .withHandle(
+                                              h ->
+                                                  h.createQuery("SELECT 1")
+                                                      .mapTo(Integer.class)
+                                                      .one()));
```
siladu, line 55: nit: This is quite hard to read, mainly due to spotless, but partly due to nested lambdas.
Slight improvement could be to replace ifPresent will a `if (slashingProtectionContext.isPresent()) {`

Could also extract out the Runnable passed to scheduleAtFixedRate.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public void run() {
+    slashingProtectionContext.ifPresent(
+        protectionContext ->
+            Executors.newScheduledThreadPool(1)
+                .scheduleAtFixedRate(
+                    () -> {
+                      Future<Integer> future =
+                          Executors.newCachedThreadPool()
```
siladu, line 46: Do we need this inner pool? Unless I'm missing something, I would have thought it was OK to block the outer scheduled thread pool since it only exists for this purpose.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+                              .submit(
+                                  () ->
+                                      protectionContext
+                                          .getSlashingProtectionJdbi()
+                                          .withHandle(
+                                              h ->
+                                                  h.createQuery("SELECT 1")
+                                                      .mapTo(Integer.class)
+                                                      .one()));
+                      try {
```
siladu, line 56: Should this try contain the DB call itself?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+                        isDbDown.set(false);
+                      } catch (Exception e) {
+                        // Have exception in database health check (timeout or error).
+                        isDbDown.set(true);
+                      } finally {
+                        future.cancel(true);
+                      }
+                    },
+                    this.dbHealthCheckIntervalMilliseconds,
+                    this.dbHealthCheckIntervalMilliseconds,
```
siladu, line 68: Should these both be interval or should one use the timeout?

commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliSlashingProtectionParameters.java
```diff
@@ -92,6 +92,22 @@ public class PicoCliSlashingProtectionParameters implements SlashingProtectionPa
       arity = "1")
   boolean pruningAtBootEnabled = true;
 
+  @Option(
+      names = {"--slashing-protection-db-health-check-timeout-milliseconds"},
+      paramLabel = "<timeout in milliseconds>",
+      description =
+          "Number of milliseconds after which the database health check will be failed (Default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private int dbHealthCheckTimeoutMilliseconds = 3000;
```
usmansaleem, line 101: consider changing to `long` as they are milliseconds.

commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliSlashingProtectionParameters.java
```diff
@@ -92,6 +92,22 @@ public class PicoCliSlashingProtectionParameters implements SlashingProtectionPa
...
+
+  @Option(
+      names = {"--slashing-protection-db-health-check-interval-milliseconds"},
+      paramLabel = "<interval in milliseconds>",
+      description =
+          "Number of milliseconds between the database health check operation (Default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private int dbHealthCheckIntervalMilliseconds = 3000;
+
```
usmansaleem, line 110: same as above, consider changing type to `long`.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+                    TimeUnit.MILLISECONDS));
+  }
+
+  public boolean isDbDown() {
```
jframe, line 72: Would prefer this was written in the positive to make the code clearer i.e. isDbUp

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Override
+  public void run() {
+    slashingProtectionContext.ifPresent(
+        protectionContext ->
+            Executors.newScheduledThreadPool(1)
+                .scheduleAtFixedRate(
+                    () -> {
+                      Future<Integer> future =
```
jframe, line 45: Is this future just so that a timeout can be set? You can set a timeout on the query itself see https://jdbi.org/apidocs/org/jdbi/v3/core/statement/SqlStatement.html#setQueryTimeout(int) instead of using the future. Then I think this inner thread pool and future could be removed and you could do something similar to DbPrunerRunner

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+    this.dbHealthCheckTimeoutMilliseconds = dbHealthCheckTimeoutMilliseconds;
+    this.dbHealthCheckIntervalMilliseconds = dbHealthCheckIntervalMilliseconds;
+    this.isDbDown = new AtomicBoolean(false);
+  }
+
+  @Override
+  public void run() {
+    slashingProtectionContext.ifPresent(
+        protectionContext ->
+            Executors.newScheduledThreadPool(1)
```
jframe, line 42: We shouldn't create the thread pool each time. It would be better to create the thread pool in the runner and schedule running the DbHealthCheck there so this only need to worry about doing the health check.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,75 @@
...
+                      Future<Integer> future =
+                          Executors.newCachedThreadPool()
+                              .submit(
+                                  () ->
+                                      protectionContext
+                                          .getSlashingProtectionJdbi()
+                                          .withHandle(
+                                              h ->
+                                                  h.createQuery("SELECT 1")
+                                                      .mapTo(Integer.class)
```
jframe, line 54: We don't actually care about the result so could just use `h.execute("SELECT 1")` instead

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -103,11 +108,12 @@ public void run() {
     try {
       createVersionMetric(metricsSystem);
       metricsEndpoint.start(vertx);
-
+      boolean failedToLoadSomeSigners = false;
       try {
         artifactSignerProvider.load().get(); // wait for signers to get loaded ...
       } catch (final InterruptedException | ExecutionException e) {
         LOG.error("Error loading signers", e);
+        failedToLoadSomeSigners = true;
```
jframe, line 116: I don't think is going to catch all cases where signers aren't loaded. In the SignerLoader class we catch most the exceptions log errors and continue for the remaining signers.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 22.8.0
+### Features Added
+- Added health check endpoint [#538](https://github.com/ConsenSys/web3signer/issues/538). 
+- Introduced `--slashing-protection-db-health-check-timeout-milliseconds` to specify the timeout of the slashing db health check procedure.
+- Introduced `--slashing-protection-db-health-check-interval-milliseconds` to specify the interval between slashing db health check procedures.
```
siladu, line 7: I like that the names have the units in :)

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -326,6 +328,26 @@ public void run() {
...
     }
+    slashingProtectionContext.ifPresent(this::scheduleDbHealthCheck);
+  }
+
+  private void scheduleDbHealthCheck(final SlashingProtectionContext protectionContext) {
+    final DbHealthCheck dbHealthCheck =
+        new DbHealthCheck(
+            protectionContext, slashingProtectionParameters.getDbHealthCheckTimeoutMilliseconds());
+
+    Executors.newScheduledThreadPool(1)
```
siladu, line 339: I think we might need to shut this thread pool down upon graceful stop of the web3signer process

core/src/main/resources/openapi-specs/eth1/web3signer.yaml
```diff
@@ -106,8 +106,50 @@ paths:
               schema:
                 type: string
               example: 'OK'
-        '500':
-          description: 'Internal Web3Signer server error'
```
siladu, line 110: The 500 was there before this PR, so should you be completely removing it here?

core/src/main/resources/openapi-specs/eth2/signing/paths/upcheck.yaml
```diff
@@ -13,5 +13,3 @@ get:
           schema:
             type: string
           example: 'OK'
-    '500':
```
siladu, line 16: as above

core/src/main/resources/openapi-specs/filecoin/web3signer.yaml
```diff
@@ -39,8 +39,49 @@ paths:
               schema:
                 type: string
               example: 'OK'
-        '500':
```
siladu, line 42: as above

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,54 @@
...
+      CompletableFuture<Integer> future =
+          CompletableFuture.supplyAsync(
+              () ->
+                  slashingProtectionContext
+                      .getSlashingProtectionJdbi()
+                      .withHandle(h -> h.execute("SELECT 1")));
+      // Check db health with timeout.
+      future.get(this.dbHealthCheckTimeoutMilliseconds, TimeUnit.MILLISECONDS);
+      isDbUp.set(true);
+    } catch (Exception e) {
```
siladu, line 45: Wondering if this should be `catch (Throwable e)` as it looks like jdbi could throw a RuntimeException and we don't want to kill the health check thread in this case.

wcgcyx: Updated

siladu (approved):
LGTM

commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliSlashingProtectionParameters.java
```diff
@@ -92,6 +92,22 @@ public class PicoCliSlashingProtectionParameters implements SlashingProtectionPa
...
+      arity = "1")
+  private long dbHealthCheckTimeoutMilliseconds = 3000;
+
+  @Option(
+      names = {"--slashing-protection-db-health-check-interval-milliseconds"},
+      paramLabel = "<interval in milliseconds>",
+      description =
+          "Number of milliseconds between the database health check operation (Default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private long dbHealthCheckIntervalMilliseconds = 3000;
```
jframe, line 109: Default for the interval is probably a bit short. I think that 30s would be a more typical value.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -266,6 +282,12 @@ private void registerUpcheckRoute(
         .failureHandler(errorHandler);
   }
 
+  protected void registerHealthCheckProcedure(
+      final String name,
+      io.vertx.core.Handler<io.vertx.core.Promise<io.vertx.ext.healthchecks.Status>> procedure) {
```
jframe, line 287: Handler doesn't need to be fully qualified we already have that type imported. Do the other types here Promos and the Status need be fully qualified as well or can we use imports?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbHealthCheck.java
```diff
@@ -0,0 +1,54 @@
...
+  private final SlashingProtectionContext slashingProtectionContext;
+  private final long dbHealthCheckTimeoutMilliseconds;
+  private final AtomicBoolean isDbUp;
+
+  public DbHealthCheck(
+      final SlashingProtectionContext slashingProtectionContext,
+      final long dbHealthCheckTimeoutMilliseconds) {
+    this.slashingProtectionContext = slashingProtectionContext;
+    this.dbHealthCheckTimeoutMilliseconds = dbHealthCheckTimeoutMilliseconds;
+    this.isDbUp = new AtomicBoolean(true);
```
jframe, line 30: nit: would be nicer to initialise this as part of the field declaration

jframe (approved):
LGTM

PR #633, opened 2022-08-26
Add final modifier to fields that can be final

jtraglia:
## PR Description

* There were some fields that could have been marked as final. Made those final.
* Made public some fields private. I see no reason for these being public.

## Documentation

## Changelog


siladu (approved):
LGTM

PR #642, opened 2022-09-06
Log network configuration in eth2 mode

### LINKED ISSUE #640: Logging network details on W3S startup
- Create a log when W3S starts to show what network is being used.
- Log the fork / epoch activated

usmansaleem:
## PR Description
Log network configuration in eth2 mode

## Fixed Issue(s)

fixes #640 

## Documentation

## Changelog

### DIFF
--- a/commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
+++ b/commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
@@ -19,8 +19,11 @@ import static tech.pegasys.web3signer.commandline.PicoCliAwsSecretsManagerParame
 import static tech.pegasys.web3signer.signing.config.AzureAuthenticationMode.CLIENT_SECRET;
 import static tech.pegasys.web3signer.signing.config.AzureAuthenticationMode.USER_ASSIGNED_MANAGED_IDENTITY;
 
+import tech.pegasys.teku.infrastructure.collections.TekuPair;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.networks.Eth2NetworkConfiguration;
+import tech.pegasys.teku.spec.ForkSchedule;
+import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.networks.Eth2Network;
 import tech.pegasys.web3signer.commandline.PicoCliAwsSecretsManagerParameters;
 import tech.pegasys.web3signer.commandline.PicoCliAzureKeyVaultParameters;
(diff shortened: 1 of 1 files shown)


commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -125,6 +133,28 @@ public Runner createRunner() {
...
+            .collect(Collectors.toMap(TekuPair::getLeft, TekuPair::getRight));
+    forkSchedule
+        .getActiveMilestones()
+        .forEach(
+            m -> {
+              final String specName = m.getSpecMilestone().name();
+              final UInt64 forkEpoch = m.getFork().getEpoch();
+              final UInt64 boundarySlot = milestoneSlotMap.get(m.getSpecMilestone());
+              LOG.info(
+                  "Spec Name: {}, Fork Epoch: {}, Boundary Slot: {}",
```
siladu, line 151: Can't see "Boundary" in the teku docs but the consensus specs do mention "fork boundary". 
It's not a blocker, but maybe "First slot" would be more intuitive for readers of this log?

siladu (approved):
Just a comment about the name, but looks good 👍

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
usmansaleem: 👍🏼

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -125,6 +133,29 @@ public Runner createRunner() {
...
+    forkSchedule
+        .getActiveMilestones()
+        .forEach(
+            m -> {
+              final String specName = m.getSpecMilestone().name();
+              final UInt64 forkEpoch = m.getFork().getEpoch();
+              final UInt64 boundarySlot = milestoneSlotMap.get(m.getSpecMilestone());
+              logString.append(
+                  String.format(
+                      "Spec Name: %s, Fork Epoch: %s, First Slot: %s%n",
```
jframe, line 153: A BIt unsure what the correct terminology should be here. Is it milestones or specs? Thought milestones would be the more correct answer but not sure.

Also, do we need the slot? Was thinking of something more succinct, perhaps we could just do something simple like Besu and print the name with the epoch? i.e. maybe something like this
`Using milestones PHASE0: 0, ALTAIR: 74240, BELLATRIX: 144896`

siladu: The slot, while it can be determined from the epoch, was useful when debugging the recent issues since the requests contain the slot

jframe: Good point. Was just trying to find a way to make it more succinct. But we can't do that after all.

PR #643, opened 2022-09-08
Suppression for CVE-2022-38752 as no fix is available yet

siladu:


gradle/owasp-suppression.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
     <!-- See https://jeremylong.github.io/DependencyCheck/general/suppression.html for examples -->
-    <suppress until="2022-08-01">
+    <suppress until="2022-09-15">
```
jframe, line 4: Is this long enough? Maybe give ourselves a month to fix it instead?

jframe (approved):
LGTM

Issue #647, opened 2022-10-07
Allow docker user flag to work Docker image


jframe:
Web3Signer docker image doesn't work with the user flag as web3signer executable and directory are owned by web3signer local user.

running `docker run -u 501 consensys/web3signer eth2 --slashing-protection-enabled=false` on my local machine gives this error.

`
docker run -u 501 consensys/web3signer eth2 --slashing-protection-enabled=false
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/opt/web3signer/bin/web3signer": stat /opt/web3signer/bin/web3signer: permission denied: unknown.
`

joshuafernandes:
```docker run -u 501:1000 consensys/web3signer eth2 --slashing-protection-enabled=false
Unable to find image 'consensys/web3signer:latest' locally
latest: Pulling from consensys/web3signer
d19f32bd9e41: Pull complete 
fbaf32f481f6: Pull complete 
051e050cb15c: Pull complete 
a8b027633188: Pull complete 
a1d08d73e8d4: Pull complete 
Digest: sha256:19b0760ebe03975ca6750e8eeb509009915552decf1bfeb5b899fd754d957e40
Status: Downloaded newer image for consensys/web3signer:latest
2022-10-07 07:49:45.938+00:00 | main | INFO  | Web3SignerApp | Web3Signer has started with args eth2,--slashing-protection-enabled=false
2022-10-07 07:49:45.943+00:00 | main | INFO  | Web3SignerApp | Version = web3signer/v22.8.1/linux-x86_64/-eclipseadoptium-openjdk64bitservervm-java-17
2022-10-07 07:49:46.879+00:00 | pool-2-thread-1 | INFO  | SignerLoader | Loading signer configuration metadata files from .
2022-10-07 07:49:46.884+00:00 | pool-2-thread-1 | INFO  | SignerLoader | Signer configuration metadata files read in memory 0 in 00:00:00.003
2022-10-07 07:49:46.888+00:00 | ForkJoinPool-1-worker-1 | INFO  | SignerLoader | Parsing configuration metadata files
2022-10-07 07:49:46.890+00:00 | ForkJoinPool-1-worker-1 | INFO  | SignerLoader | Total configuration metadata files processed: 0
[...]

PR #650, opened 2022-10-11
Fix installation directory permission in Web3Signer docker image

usmansaleem:
## PR Description
Fixes installation directory permission in Docker image.

In our Docker file, we have `RUN adduser --disabled-password --gecos "" --home /opt/web3signer` that sets the top level `/opt/web3signer` with restricted permissions as this becomes the home folder of user `web3signer`, everything else under `/opt/web3signer` is already correct permission because we extract them from our build .tar.gz.

The fix is either to use a different `--home` in `adduser` command or fix the top level directory permission i.e. `/opt/web3signer` to allow access to the directory to the world (it defaults to user and group due to being home folder).

To test the fix via local build:
~~~
./gradlew --no-daemon installDist
./gradlew --no-daemon distDocker
docker run --rm -u 501 consensys/web3signer:develop eth2 --slashing-protection-enabled=false --version
~~~

## Fixed Issue(s)

#647 

## Documentation

## Changelog

### DIFF
--- a/docker/jdk11/Dockerfile
+++ b/docker/jdk11/Dockerfile
@@ -16,7 +16,7 @@ COPY --from=jre-build /javaruntime $JAVA_HOME
 
 RUN apt-get -y update && apt-get -y install curl iputils-ping net-tools && rm -rf /var/lib/api/lists/*
 RUN adduser --disabled-password --gecos "" --home /opt/web3signer web3signer && \
-    chown web3signer:web3signer /opt/web3signer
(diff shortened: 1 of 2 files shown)


joshuafernandes:
Hi @usmansaleem just wanted to get some context on this mate. Can the existing container not just be run like so:

```
docker run --rm -u 501:1000 consensys/web3signer:develop ```

joshuafernandes (approved):
LGTM

jframe (approved):
LGTM

Issue #598, opened 2022-06-30
Enable pruning by default


jframe:
Enable slashing protection by default. It is safe to have pruning enabled and will help with database performance if there is a lot of history.

PR #668, opened 2022-10-31
Better database pruning default values

siladu:
Enable pruning by default with these default values: `slashing-protection-pruning-epochs-to-keep = 250`, `slashing-protection-pruning-at-boot-enabled = false` and `slashing-protection-pruning-interval = 12`

Fixes: https://github.com/ConsenSys/web3signer/issues/598
Fixes: https://github.com/ConsenSys/web3signer/issues/630


jframe (approved):
LGTM

PR #671, opened 2022-11-03
Improve performance of slashing data import

### LINKED ISSUE #670: Performance improvement for slashing protection import
Slashing protection import subcommand takes a longer than expected time to import data. For ~4400 validators it is taking ~90 minutes.

jframe:
## PR Description
This improves the performance of slashing data import by parallelising the import of multiple validators at the same time. To do this the import transaction is now done per validator instead of for all the validators so that multiple database connections can be used. Because of this the import now allows the partial import of valid data. That is any invalid data for a validator import now only affects that validator not all of the import.

In my local testing importing a slashing protection file with 4395 validators.
Current import takes: 1:31:14
With #671 changes takes: 0:19:20

## Fixed Issue(s)

fixes #670 

## Documentation

## Changelog

### DIFF
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Importer.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Importer.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.IntStream;
 
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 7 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Importer.java
```diff
@@ -100,22 +101,24 @@ private void importDataInternal(final InputStream input, final Optional<List<Str
...
+          .parallel()
+          .forEach(
+              i -> {
+                try {
+                  jdbi.useTransaction(
+                      h -> {
+                        final JsonNode validatorNode = dataNode.get(i);
+                        importValidator(h, validatorNode, pubkeys);
+                      });
+                } catch (Exception e) {
```
usmansaleem, line 114: nit: `final` in catch block. Also, I would append `e.getMessage()` with the error log to have a better idea what went wrong during the transaction.

usmansaleem (approved):
LGTM - minor nitpicks.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Importer.java
jframe: good idea. done

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Importer.java
```diff
@@ -140,6 +146,7 @@ private void parseValidator(
 
     final ArrayNode signedAttestationNode = parentNode.withArray("signed_attestations");
     importAttestations(handle, validator, signedAttestationNode);
+    LOG.info("Imported slashing protection data for validator {}", validator.getPublicKey());
```
siladu, line 149: Could be too spammy with 1000s of validators?

siladu: ah, I suppose we probably want to know which succeeded though

siladu: On second thought a full record of successful and failed validators is probably really useful

siladu (approved):

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/interchange/InterchangeV5Importer.java
jframe: Actually less spammy than it was before now I have hidden the other logs at debug level. Think it's probably ok wouldn't want more than one line of logging though.

PR #679, opened 2022-11-23
Watermark repair subcommand

### LINKED ISSUE #493: Subcommand to adjust the signing protection watermark
Add a subcommand that will allow adjusting of the slashing protection watermark. This would be useful if all the slashing protection data has been lost and allow web3signer to be safely used to sign again by setting the watermark to a future slot.

Tasks
* Create new subcommand at the same level as Eth2Export, Eth2Import, etc. (e.g. SlashingProtectionSubCommand or WatermarkSubCommand?)
* Create service layer (may be small enough to go in subcommand) which uses tech.pegasys.teku.spec.Spec to determine the epoch/slot as the default for the command.
* Allow user to configure their own epoch/slot?
* Consider adding an acceptance test

**Notes**
* An approach similar to the Teku signing-protection repair subcommand https://docs.teku.consensys.net/en/latest/Reference/CLI/Subcommands/Slashing-Protection/#repair could be used.
* If automatically setting the slot and epoch watermark we can use the Teku spec object and calculate the exact epoch and slot
* Optional nice to have to do in another story: take the opportunity to prune the DB.
* Probably only need to configure one of epoch or slot.
* Epoch source and target can be the same value.

**Acceptance Criteria**
* Subcommand that allows setting the attestation and block watermark to the provided values
* Can automatically set the slot and epoch to next epoch given the network name
* Should set watermark for all the validators the web3signer instance knows about (rather than everything in the database), c.f. how pruning works: https://github.com/ConsenSys/web3signer/blob/6fa9df2fa42a8a67da19eb6fe5f210c4b9467d76/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java#L264

Tasks
- [ ] Create watermark subcommand
- [ ] Update watermarks for all known validators

jframe:
## PR Description
Adds a subcommand to allow updating of the low watermarks for validators.

## Fixed Issue(s)

fixes #493 

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsConfigFileImpl.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsConfigFileImpl.java
@@ -171,6 +171,18 @@ public class CmdLineParamsConfigFileImpl implements CmdLineParamsBuilder {
                 YAML_STRING_FMT,
                 "eth2.import.from",
                 signerConfig.getSlashingImportPath().get().toAbsolutePath()));
+      } else if (signerConfig.getWatermarkRepairParameters().isPresent()) {
+        params.add("watermark-repair"); // sub-sub command
+        yamlConfig.append(
+            String.format(
+                YAML_NUMERIC_FMT,
+                "eth2.watermark-repair.slot",
+                signerConfig.getWatermarkRepairParameters().get().getSlot()));
+        yamlConfig.append(
+            String.format(
+                YAML_NUMERIC_FMT,
+                "eth2.watermark-repair.epoch",
+                signerConfig.getWatermarkRepairParameters().get().getEpoch()));
       }
     }
 
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsConfigFileImpl.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsConfigFileImpl.java
@@ -48,7 +48,6 @@ public class CmdLineParamsConfigFileImpl implements CmdLineParamsBuilder {
   private static final String YAML_STRING_FMT = "%s: \"%s\"%n";
   private static final String YAML_NUMERIC_FMT = "%s: %d%n";
   private static final String YAML_BOOLEAN_FMT = "%s: %b%n";
-  private static final String YAML_LIST_STRING_FMT = "%s: [\"%s\"]%n";
 
   public CmdLineParamsConfigFileImpl(final SignerConfiguration signerConfig, final Path dataPath) {
     this.signerConfig = signerConfig;
@@ -412,6 +411,6 @@ public class CmdLineParamsConfigFileImpl implements CmdLineParamsBuilder {
   private String formatStringList(final String key, final List<String> stringList) {
     final StringJoiner stringJoiner = new StringJoiner(",", "\"", "\"");
     stringList.forEach(stringJoiner::add);
-    return String.format(YAML_LIST_STRING_FMT, key, stringJoiner);
+    return String.format("%s: [\"%s\"]%n", key, stringJoiner);
   }
 }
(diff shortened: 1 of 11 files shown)


acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsConfigFileImpl.java
```diff
@@ -171,6 +173,23 @@ public List<String> createCmdLineParams() {
                 YAML_STRING_FMT,
                 "eth2.import.from",
                 signerConfig.getSlashingImportPath().get().toAbsolutePath()));
+      } else if (signerConfig.getWatermarkRepairParameters().isPresent()) {
```
usmansaleem, line 176: may be we can now refactor a method e.g. `manageSubCommands()` as the list is growing.

usmansaleem (approved):
LGTM

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/runner/CmdLineParamsConfigFileImpl.java
```diff
@@ -388,4 +407,10 @@ private String awsBulkLoadingOptions(
 
     return yamlConfig.toString();
   }
+
+  private String formatStringList(final String key, final List<String> stringList) {
+    final StringJoiner stringJoiner = new StringJoiner(",", "\"", "\"");
+    stringList.forEach(stringJoiner::add);
```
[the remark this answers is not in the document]
jframe: Nice. That worked.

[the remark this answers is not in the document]
jframe: done

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2WatermarkRepairSubCommand.java
```diff
@@ -0,0 +1,108 @@
...
+      arity = "0..*")
+  List<String> validatorIds = new ArrayList<>();
+
+  @Override
+  public void run() {
+    final LowWatermarkDao lowWatermarkDao = new LowWatermarkDao();
+    final ValidatorsDao validatorsDao = new ValidatorsDao();
+    final List<Bytes> validatorIdPublicKeys =
+        validatorIds.stream()
+            .filter(s -> !s.isEmpty())
```
usmansaleem, line 77: May be StringUtils:isEmpty can be used to handle both empty and null values.

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2WatermarkRepairSubCommand.java
```diff
@@ -0,0 +1,108 @@
...
+            .map(Bytes::fromHexStringLenient)
```
usmansaleem, line 78: flatMap(this::convertToBytesFromHex)...
where convertToBytes uses Bytes with a try/catch and returns Optional<Bytes>

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2WatermarkRepairSubCommand.java
jframe: done

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2WatermarkRepairSubCommand.java
jframe: Rather throw a parameter exception if the conversion fails as the command line arguments are invalid in this case rather than just ignoring them

usmansaleem: fair point 👍🏼

usmansaleem (approved):
LGTM

Issue #689, opened 2022-12-02
Support yaml multi-doc config files


jframe:
Add support for yaml multidoc config files so that a single config file can be used rather than multiple config files. This would make it easier to configure particularly in a kubernetes environment where providing multiple files would mean changing the k8s config each time.

usmansaleem:
Dev note: Reloading of config file changes. Consider using hash of file to monitor file changes ? Will it work in K8?

PR #692, opened 2022-12-07
Support multidoc yaml signing metadata configuration

usmansaleem:
## PR Description
Support multidoc yaml signing metadata configuration files

## Fixed Issue(s)

See #689 

## Documentation

## Changelog

### DIFF
--- a/signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
+++ b/signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
@@ -55,6 +55,7 @@ public class SignerLoader {
 
     final Map<Path, String> configFilesContent =
         getConfigFilesContents(configsDirectory, fileExtension);
+    // TODO (Uzi): Cache files with their hash values instead of file name to pick any changes during reload
     loadedConfigurationFiles.addAll(configFilesContent.keySet());
 
     final String timeTaken =
@@ -71,6 +72,7 @@ public class SignerLoader {
       final Path configsDirectory, final String fileExtension) {
     // read configuration files without converting them to signers first.
     // Avoid reading files which are already loaded/processed
+    // TODO (Uzi): Cache files with their hash values instead of file name to pick any changes during reload
     try (final Stream<Path> fileStream = Files.list(configsDirectory)) {
       return fileStream
           .filter(path -> !loadedConfigurationFiles.contains(path))
--- a/signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
+++ b/signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
@@ -20,12 +20,15 @@ import tech.pegasys.web3signer.signing.config.metadata.SigningMetadata;
 import tech.pegasys.web3signer.signing.config.metadata.SigningMetadataException;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
 
@@ -48,12 +51,16 @@ public class YamlSignerParser implements SignerParser {
   @Override
   public List<ArtifactSigner> parse(final String fileContent) {
     try {
-      final SigningMetadata metaDataInfo =
-          YAML_MAPPER.readValue(fileContent, SigningMetadata.class);
+      final List<SigningMetadata> metadataInfoList = readSigningMetadata(fileContent);
 
-      return signerFactories.stream()
-          .filter(factory -> factory.getKeyType() == metaDataInfo.getKeyType())
-          .map(metaDataInfo::createSigner)
+      return metadataInfoList.stream()
+          .flatMap(
+              metadataInfo ->
+                  signerFactories.stream()
+                      .filter(
+                          factory ->
+                              Objects.equals(factory.getKeyType(), metadataInfo.getKeyType()))
+                      .map(metadataInfo::createSigner))
           .collect(Collectors.toList());
     } catch (final JsonParseException | JsonMappingException e) {
       throw new SigningMetadataException("Invalid signing metadata file format", e);
@@ -65,5 +72,16 @@ public class YamlSignerParser implements SignerParser {
     } catch (final Exception e) {
       throw new SigningMetadataException("Unknown failure", e);
     }
+  }
+
+  private static List<SigningMetadata> readSigningMetadata(String fileContent) throws IOException {
+    final List<SigningMetadata> signingMetadataList = new ArrayList<>();
+    try (final MappingIterator<SigningMetadata> signingMetadataIterator =
+        YAML_MAPPER.readerFor(SigningMetadata.class).readValues(fileContent)) {
+      while (signingMetadataIterator.hasNextValue()) {
+        signingMetadataList.add(signingMetadataIterator.nextValue());
+      }
+    }
+    return signingMetadataList;
   }
 }
--- a/signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserMultiReadTest.java
+++ b/signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserMultiReadTest.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2022 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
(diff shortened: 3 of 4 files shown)


signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -46,16 +48,18 @@ public class SignerLoader {
   private static final Logger LOG = LogManager.getLogger();
   private static final long FILES_PROCESSED_TO_REPORT = 10;
   private static final int MAX_FORK_JOIN_THREADS = 5;
-  private final Set<Path> loadedConfigurationFiles = new HashSet<>();
+
+  private static final Map<Path, FileTime> loadedConfigFiles = new ConcurrentHashMap<>();
```
jframe, line 52: Is there a performance impact from using ConcurrentHashMap? I think I remember that being a reason for refactoring this at some point to avoid use of ConcurrentHashMap

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserTest.java
```diff
@@ -233,7 +235,7 @@ void keyStoreMetaDataInfoWithHttpURIFailsWithException() throws IOException {
 
   @Test
   void aSignerIsCreatedForEachMatchingFactory() throws IOException {
-    lenient().when(otherBlsArtifactSignerFactory.getKeyType()).thenReturn(KeyType.BLS);
+    lenient().when(secpArtifactSignerFactory.getKeyType()).thenReturn(KeyType.BLS);
```
jframe, line 238: Should return a KeyType of secp in this case

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
```diff
@@ -66,4 +73,15 @@ public List<ArtifactSigner> parse(final String fileContent) {
       throw new SigningMetadataException("Unknown failure", e);
     }
   }
+
+  private static List<SigningMetadata> readSigningMetadata(String fileContent) throws IOException {
```
jframe, line 77: Based on this https://github.com/FasterXML/jackson-dataformats-text/issues/66 this could be simplified to use the readAll method. So something like this should work.

```
final JsonParser parser = YAML_MAPPER.createParser(fileContent);
final List<SigningMetadata> metadataInfoList = YAML_MAPPER.readValues(parser, new TypeReference<SigningMetadata>(){}).readAll();
```

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
```diff
@@ -48,12 +51,16 @@ public YamlSignerParser(final Collection<AbstractArtifactSignerFactory> signerFa
...
-      return signerFactories.stream()
-          .filter(factory -> factory.getKeyType() == metaDataInfo.getKeyType())
-          .map(metaDataInfo::createSigner)
+      return metadataInfoList.stream()
+          .flatMap(
+              metadataInfo ->
+                  signerFactories.stream()
+                      .filter(
+                          factory ->
+                              Objects.equals(factory.getKeyType(), metadataInfo.getKeyType()))
```
jframe, line 62: What's the benefit of using Object.equals for the enum?

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -67,18 +71,37 @@ public Collection<ArtifactSigner> load(
     return processMetadataFilesInParallel(configFilesContent, signerParser);
   }
 
-  private Map<Path, String> getConfigFilesContents(
+  private static void updateLoadedConfigFilesMap() {
```
jframe, line 74: this doesn't need to be static

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -67,18 +71,37 @@ public Collection<ArtifactSigner> load(
...
+    // remove loaded config files which do not exist anymore
+    new HashSet<>(loadedConfigFiles.keySet())
```
jframe, line 76: Wrapping this in another set shouldn't necessary if we are using a concurrent hashmap

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserMultiReadTest.java
```diff
@@ -0,0 +1,92 @@
...
+    Mockito.reset();
+    lenient().when(blsArtifactSignerFactory.getKeyType()).thenReturn(KeyType.BLS);
+    lenient().when(secp256k1ArtifactSignerFactory.getKeyType()).thenReturn(KeyType.SECP256K1);
+
+    signerParser =
+        new YamlSignerParser(List.of(blsArtifactSignerFactory, secp256k1ArtifactSignerFactory));
+  }
+
+  @Test
+  void readMultiDoc() {
```
jframe, line 54: A test that shows that multidoc with different config types would be good

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
usmansaleem: We can consider two approaches here. First make SignerLoader a singelton, second use this Map as a static member variable which is thread-safe. 

The main reason to make it concurrent is to cater for "removal" of config file paths. Since `load` can be called via reload endpoint asynchronously, there is a corner case where an entry is removed in one thread while it is being accessed during loading. On the other hand, if we want to keep track of Paths of all the files that we have read, then we can revert to standard HashMap.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
usmansaleem: 👍🏼

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
usmansaleem: no benefit, using `==` or `Object.equals` would result in same result.

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
usmansaleem: wouldn't it cause an issue in `forEach`? as we would be removing an item from same iterator.

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserMultiReadTest.java
usmansaleem: will add one.

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -67,18 +71,37 @@ public Collection<ArtifactSigner> load(
...
-          .filter(path -> !loadedConfigurationFiles.contains(path))
           .filter(path -> matchesFileExtension(fileExtension, path))
           .map(
               path -> {
                 try {
+                  // only read file if its last modified time has been changed or not already loaded
+                  final FileTime lastModifiedTime = Files.getLastModifiedTime(path);
+                  if (loadedConfigFiles.containsKey(path)) {
+                    if (loadedConfigFiles.get(path).compareTo(lastModifiedTime) == 0) {
+                      return null;
```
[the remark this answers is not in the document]
usmansaleem: yes, good idea.

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserTest.java
usmansaleem: Originally it wasn't, but I've fixed it to use secp type correctly.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParser.java
usmansaleem: converted, however, still kept try/resource block.

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserMultiReadTest.java
usmansaleem: added. Encrypted Keystore and Raw file types within same file.

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -67,30 +69,49 @@ public Collection<ArtifactSigner> load(
...
-                  LOG.error("Error reading config file: {}", path, e);
-                  return null;
-                }
-              })
-          .filter(Objects::nonNull)
-          .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+          .filter(this::isNewOrModifiedMetadataFile)
+          .map(this::getMetadataFileContent)
+          .filter(Optional::isPresent)
+          .map(Optional::get)
```
siladu, line 81:
```suggestion
          .flatMap(Optional::stream)
```

siladu (approved):
LGTM, just a small suggested change

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
jframe: Yeah it would do

jframe (approved):

PR #695, opened 2022-12-19
Increase dependency check to use high cvss score

jframe:
## PR Description
Increaese dependency check to use high cvss score

## Documentation

## Changelog


siladu (approved):
LGTM

PR #720, opened 2023-02-23
Change signing indexes and sequences to use bigint to fix out of index out of bounds error

### LINKED ISSUE #705: Slashing protection database fails reaching max index
Web3signer slashing protection database inserts will eventually fail due postgresql reaching the maximum value of the index.

When this happens we will get an error like this:
`org.postgresql.util.PSQLException: ERROR: nextval: reached maximum value of sequence "signed_attestations_id_seq" (2147483647) [statement:"INSERT INTO signed_attestations (validator_id, signing_root, source_epoch, target_epoch) VALUES (?, ?, ?, ?)"`

A workaround is to reset the sequences:
```
ALTER SEQUENCE signed_attestations_id_seq RESTART WITH 1;
ALTER SEQUENCE signed_blocks_id_seq RESTART WITH 1;
```

A better solution is to change the index to use serial8 instead of serial https://www.cybertec-postgresql.com/en/error-nextval-reached-maximum-value-of-sequence/

jframe:
## PR Description
Inserting blocks or attestations fails once the maximum value for the autoincrementing index is reached. This currently has a value of 2147483647. This is an issue particularly for gnosis chain.

Change signing indexes and sequences to use bigint a 64-bit value so that index now has a max value of 2^63-1 instead.

Tested migration times
1,000 validators with 1,000 signings for each validator took 00:02.936s
10,000 validators with 10,000 signings for each validator took 05:00.126s

I manually tested this by setting the current value to the max value and attempting to insert a new signing.

## Fixed Issue(s)

fixes #705 

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
 # Changelog
 
 ## Next release
+
 ### Features Added
 -
 ### Bugs fixed
 - Upgrade to Vertx 4.3.8 to address CVE-2023-24815
-- 
+- Fix issue with slashing protection database failing once reaching max integer index value [#705](https://github.com/ConsenSys/web3signer/issues/705)
```
siladu, line 9: Is this a breaking change if someone wants to downgrade?

usmansaleem (approved):
Looks good to me.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/SignedAttestationsDaoTest.java
```diff
@@ -280,6 +280,21 @@ public void findsNearestAttestationForTargetEpoch(final Handle handle) {
         .isEqualToComparingFieldByField(attestation(1, 7, 8, 1));
   }
 
+  @Test
+  public void insertsAttestationWithIdGreaterThanIntBoundary(final Handle handle) {
```
siladu, line 284: nit: Wonder if it could be clearer that we're dealing with a former boundary case, rather than a current boundary case. At a glance, I would assume this is testing the actual boundary

CHANGELOG.md
jframe: Yeah, it is a breaking change to the database. I'll add a note for that

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/dao/SignedAttestationsDaoTest.java
jframe: It's testing the old boundary. I'll change to name to make it clear the intent is to make sure we can insert ids with a value greater than the integer max value

siladu (approved):
LGTM

Issue #719, opened 2023-02-22
"YAMLException: The incoming YAML document exceeds the limit: 3145728 code points."


gg718:
While testing PR [#692](https://github.com/ConsenSys/web3signer/pull/692), I came across the following issue with files >3MB:

> `Error parsing signing metadata file /tmp/test.yaml: YAMLException: The incoming YAML document exceeds the limit: 3145728 code points.`

Is there a reason for this limit, or can it be increased?

Ping @usmansaleem

usmansaleem:
Hi @gg718, this "limitation" was not intended and got introduced due to underlying SnakeYaml implementation (1.32) which is used by Jackon dataformat libraries. See https://github.com/FasterXML/jackson-dataformats-text/tree/2.15/yaml#maximum-input-yaml-document-size-3-mb

We will make some modifications in Web3Signer to set the default value to 1GB and also make it configureable using a cli option as well.

PR #721, opened 2023-02-23
Set yaml default code point limit

usmansaleem:
## PR Description
Set yaml default code point limit.
 -- Default to 100MB
 -- Introduced cli option `--key-store-config-files-size`

## Fixed Issue(s)

#719 
## Documentation

## Changelog


signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Optional;
+
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import org.yaml.snakeyaml.LoaderOptions;
+
+public class YamlMapperProvider {
+  private static final int YAML_DEFAULT_CODE_POINT_LIMIT =
```
jframe, line 25: This is a really high limit. Maybe 100MB?

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,66 @@
...
+      1_073_741_824; // 1024 * 1024 * 1024 = 1GB
+  private static final String SYSTEM_PROPERTY = "web3signer.yamlCodePointLimit";
```
jframe, line 27: Seems like this should a pico cli option so we get all the ways to configure it like other options

jframe, line 27: The name probably won't mean anything to anyone who doesn't know snakeyaml.
If the limit the same of the max file size? Maybe maxYamlFileSize?

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
usmansaleem: agreed, updated the code to use cli option instead.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
usmansaleem: agreed, updated the code to use cli option instead.

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +82,20 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
-      description = "The path to a directory storing toml files defining available keys",
+      description = "The path to a directory storing yaml files defining available keys",
       paramLabel = DefaultCommandValues.MANDATORY_PATH_FORMAT_HELP,
       arity = "1")
   private Path keyStorePath = Path.of("./");
 
+  @Option(
+      names = {"--key-store-config-files-size"},
+      description =
+          "The key store configuration file size in bytes. Useful when loading large number of "
```
jframe, line 93: Similarly we should mention that this is the max file size

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +82,20 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
   @Option(
       names = {"--key-store-path"},
-      description = "The path to a directory storing toml files defining available keys",
+      description = "The path to a directory storing yaml files defining available keys",
       paramLabel = DefaultCommandValues.MANDATORY_PATH_FORMAT_HELP,
       arity = "1")
   private Path keyStorePath = Path.of("./");
 
+  @Option(
+      names = {"--key-store-config-files-size"},
```
jframe, line 91: maybe --key-store-config-files-max-size?

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +82,20 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
+      description =
+          "The key store configuration file size in bytes. Useful when loading large number of "
+              + "configurations from same yaml file. Defaults to 100MB (104857600) ",
+      paramLabel = "<NUMBER>",
+      arity = "1")
+  private final Integer keystoreConfigFileSize = 104_857_600;
```
jframe, line 97: Is integer large enough considering this is bytes? maybe long would be better

usmansaleem: the yaml codePoint method takes an int.

usmansaleem:
Close due to #729

Issue #682, opened 2022-12-01
Investigate changes needed for Capella in Web3Signer


jframe:
For Capella CL updates, there’s new block bodies and blinded block bodies to cater for sending bls_to_execution changes and withdrawals. 

- Identify changes for Web3Signer in terms of withdrawl signing requests (New API or modifications to existing API?). 
- Coordinate with EL team (such as Teku) to finalize the changes
- Create new issues if required.

**Timebox to 0.5 day**

james-prysm:
might be a duplicate or related to https://github.com/ConsenSys/web3signer/issues/673

PR #722, opened 2023-02-23
Capella changes

usmansaleem:
## PR Description
Update Web3Signer to sign block for Capella spec.

- Update OpenApi spec
- Introduced manual fork overridden cli option
- Update Acceptance Test
- Update Teku library to 23.2.0

## Fixed Issue(s)

#682 
## Documentation

## Changelog


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2BlockSigningAcceptanceTest.java
```diff
@@ -88,14 +88,18 @@ void signAndVerifyLegacyBlockSignature() throws Exception {
...
-  @Test
-  void emptyBlockRequestReturnsBadRequestStatus() throws JsonProcessingException {
-    final Eth2BlockSigningRequestUtil util =
-        new Eth2BlockSigningRequestUtil(SpecMilestone.BELLATRIX);
-    setupEth2Signer(Eth2Network.MINIMAL, SpecMilestone.BELLATRIX);
+  @ParameterizedTest(
+      name = "#{index} - Empty block request for spec {0} should return bad request status")
+  @EnumSource(
+      value = SpecMilestone.class,
+      names = {"BELLATRIX", "CAPELLA"})
```
jframe, line 95: Why does this only apply to bellatrix and capella? should also include phase0 and altair?

core/src/main/resources/openapi-specs/eth2/signing/schemas.yaml
```diff
@@ -464,6 +465,19 @@ components:
       required:
         - version
         - block_header
+    BlockRequestCapella:
+      type: object
+      properties:
+        version:
+          type: string
+          enum:
+            - CAPELLA
```
jframe, line 474: Annoying that we need to define this just for the enum value. But think we have to do this way with the current API.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/Eth2BlockSigningAcceptanceTest.java
usmansaleem: good point, updated it to cover phase0 and altair as well.

jframe (approved):
LGTM

PR #729, opened 2023-02-27
Fix yaml config file size

usmansaleem:
## PR Description
Fix yaml config file size default limit
-- Default to 100MB
-- Introduced cli option `--key-store-config-file-max-size`

## Fixed Issue(s)

#719 
## Documentation

## Changelog


signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,56 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Optional;
+
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import org.yaml.snakeyaml.LoaderOptions;
+
+public enum YamlMapperProvider {
```
siladu, line 25: nit: I like the enum approach but is there any advantage to it if we still need to use a synchronized init method to achieve lazy init?

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserTest.java
```diff
@@ -73,6 +75,11 @@ class YamlSignerParserTest {
 
   private YamlSignerParser signerParser;
 
+  @BeforeAll
+  static void init() {
+    YamlMapperProvider.INSTANCE.init(Optional.empty());
```
siladu, line 80: Can't we just create a test fixture version of the YamlMapper and use it everywhere?

siladu (commented):
Wondering if we need the extra init() step for the singleton...also can we remove a lot of init code by using a single test fixture?

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +83,21 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
       arity = "1")
   private Path keyStorePath = Path.of("./");
 
+  @Option(
+      names = {"--key-store-config-file-size"},
+      description =
+          "The key store configuration file size in bytes. Useful when loading large number of "
+              + "configurations from same yaml file. Defaults to (${DEFAULT-VALUE}) 100MB",
+      paramLabel = "<NUMBER>",
+      defaultValue = "104857600",
```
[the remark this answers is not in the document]
jframe: If we don't specify a defaultValue it should take the default value from the initial value assigned to keystoreConfigFileSize

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +83,21 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
+      description = "The path to a directory storing yaml files defining available keys",
       paramLabel = DefaultCommandValues.MANDATORY_PATH_FORMAT_HELP,
       arity = "1")
   private Path keyStorePath = Path.of("./");
 
+  @Option(
+      names = {"--key-store-config-file-size"},
+      description =
+          "The key store configuration file size in bytes. Useful when loading large number of "
+              + "configurations from same yaml file. Defaults to (${DEFAULT-VALUE}) 100MB",
```
jframe, line 95: The second sentence would be clearer as "Useful when loading a large number of configurations from the same yaml file"

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -299,6 +310,18 @@ public void validateArgs() {
     final PicoCliTlsServerOptionsValidator picoCliTlsServerOptionsValidator =
         new PicoCliTlsServerOptionsValidator(spec, picoCliTlsServerOptions);
     picoCliTlsServerOptionsValidator.validate();
+
+    // validate keystore config file size and initialize the YamlMapperProvider singelton
```
jframe, line 314: typo: /singelton/singleton/

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
jframe: I don't like the enum approach. This is not really an enum it's just using it for instantiation and doesn't hide the initialisation that you usually have with a singleton.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,56 @@
...
+  INSTANCE;
+  private YAMLMapper yamlMapper;
+  private static final int yamlCodePointLimit = 100 * 1024 * 1024; // 100 MB default
+
+  public synchronized void init(Optional<Integer> yamlFileSizeInBytes) {
+    if (yamlMapper != null) {
+      return;
+    }
+    final LoaderOptions loaderOptions = new LoaderOptions();
+    if (yamlFileSizeInBytes.isPresent() && yamlFileSizeInBytes.get() > 0) {
```
jframe, line 35: Do we need to validate that yamlFileSizeInBytes is > 0? We can't pass in such a value due to the PicoCLI validation.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,56 @@
...
+      loaderOptions.setCodePointLimit(yamlFileSizeInBytes.get());
+    } else {
+      loaderOptions.setCodePointLimit(yamlCodePointLimit);
```
jframe, line 38: It would be simpler to not use an optional rely on the default value that will get assigned by picocli. If you go rid of the optional then you remove `if (yamlFileSizeInBytes.isPresent() && yamlFileSizeInBytes.get() > 0)` case and just do `loaderOptions.setCodePointLimit(yamlCodePointLimit)`

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,56 @@
...
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import org.yaml.snakeyaml.LoaderOptions;
+
+public enum YamlMapperProvider {
+  INSTANCE;
+  private YAMLMapper yamlMapper;
+  private static final int yamlCodePointLimit = 100 * 1024 * 1024; // 100 MB default
+
+  public synchronized void init(Optional<Integer> yamlFileSizeInBytes) {
```
jframe, line 30: This is creating a new instance each time. Should this be only allowed once? As above would rather we just follow a typical singleton pattern and did the initialisation on the getYamlMapper call if it was uninitialised.

usmansaleem: Although this wasn't creating instance each time, I have converted this class to standard singleton pattern.

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +83,21 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
-      description = "The path to a directory storing toml files defining available keys",
+      description = "The path to a directory storing yaml files defining available keys",
       paramLabel = DefaultCommandValues.MANDATORY_PATH_FORMAT_HELP,
       arity = "1")
   private Path keyStorePath = Path.of("./");
 
+  @Option(
+      names = {"--key-store-config-file-size"},
+      description =
+          "The key store configuration file size in bytes. Useful when loading large number of "
```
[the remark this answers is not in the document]
usmansaleem: updated to use suggested cli option name.

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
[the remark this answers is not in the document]
usmansaleem: removed the defaultValue annotation. This was left by mistake because I was experimenting with using `@Option` on setter method.

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -299,6 +310,18 @@ public void validateArgs() {
     final PicoCliTlsServerOptionsValidator picoCliTlsServerOptionsValidator =
         new PicoCliTlsServerOptionsValidator(spec, picoCliTlsServerOptions);
     picoCliTlsServerOptionsValidator.validate();
+
+    // validate keystore config file size and initialize the YamlMapperProvider singelton
+    if (keystoreConfigFileSize <= 0) {
+      throw new ParameterException(
+          spec.commandLine(),
+          String.format(
+              "Invalid value '%s' for option '--key-store-config-file-size': "
```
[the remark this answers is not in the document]
usmansaleem: done.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import org.yaml.snakeyaml.LoaderOptions;
+
+public enum YamlMapperProvider {
+  INSTANCE;
+  private YAMLMapper yamlMapper;
+  private static final int yamlCodePointLimit = 100 * 1024 * 1024; // 100 MB default
```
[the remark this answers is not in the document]
usmansaleem: I've removed it. The default of 3MB should work for all of our local tests and when we don't call init method explicitly.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,56 @@
...
+        builder
+            .addModule(new SigningMetadataModule())
+            .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)
+            .disable(YAMLGenerator.Feature.USE_NATIVE_TYPE_ID)
+            .enable(ACCEPT_CASE_INSENSITIVE_ENUMS)
+            .build();
+  }
+
+  public YAMLMapper getYamlMapper() {
+    checkNotNull(yamlMapper, "YamlMapperProvider must be initialized");
```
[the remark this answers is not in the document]
usmansaleem: I've removed this check. Now we are instantiating YamlMapper here with defaults (which is internally 3MB).

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
usmansaleem: I've converted enum to standard class with singelton pattern.

signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserTest.java
usmansaleem: The requirement of explicit init is removed as getInstance().getMapper() would initialize it with defaults if not already initialised.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
usmansaleem: I've removed this validation.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
usmansaleem: The reason for Optional is to perform default initialization if init is not explicitly called i.e. getMapper would call init with Optional.empty.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,59 @@
...
+            .enable(ACCEPT_CASE_INSENSITIVE_ENUMS)
+            .build();
+  }
+
+  public YAMLMapper getYamlMapper() {
+    if (yamlMapper == null) {
+      init(Optional.empty());
+    }
+    return yamlMapper;
+  }
```
siladu, line 58: Nit: could we save a method call by passing the value as a param to getYamlMapper?

siladu (approved):

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,59 @@
...
+  private YAMLMapper yamlMapper;
+  private static final YamlMapperProvider INSTANCE = new YamlMapperProvider();
+
+  private YamlMapperProvider() {}
+
+  public static YamlMapperProvider getInstance() {
+    return INSTANCE;
+  }
+
+  public synchronized void init(Optional<Integer> yamlFileSizeInBytes) {
```
jframe, line 34: This doesn't need to be optional as we will always have a value because of the defaulting of the value in Web3SignerBaseCommand

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlMapperProvider.java
```diff
@@ -0,0 +1,59 @@
...
+            .addModule(new SigningMetadataModule())
+            .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)
+            .disable(YAMLGenerator.Feature.USE_NATIVE_TYPE_ID)
+            .enable(ACCEPT_CASE_INSENSITIVE_ENUMS)
+            .build();
+  }
+
+  public YAMLMapper getYamlMapper() {
+    if (yamlMapper == null) {
+      init(Optional.empty());
```
jframe, line 55: Don't think we ever want to pass through an Optional.empty(), thinking we should through an exception rather than allow this to be initialised incorrectly

jframe (changes requested):

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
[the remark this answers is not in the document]
usmansaleem: Reintroduced the defaultValue as I've changed the approach to create YamlMapper by using PicoCli custom type converter.

commandline/src/main/java/tech/pegasys/web3signer/commandline/convertor/YamlMapperCliConverter.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package tech.pegasys.web3signer.commandline.convertor;
+
+import tech.pegasys.web3signer.signing.config.metadata.parser.YamlMapperFactory;
+
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import picocli.CommandLine;
+
+public class YamlMapperCliConverter implements CommandLine.ITypeConverter<YAMLMapper> {
+  public static final String KEY_STORE_CONFIG_FILE_SIZE_OPTION_NAME =
```
siladu, line 21: nit: would this make more sense to live in Web3SginerBaseCommand?

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -82,11 +85,22 @@ public class Web3SignerBaseCommand implements Config, Runnable {
...
   private Path keyStorePath = Path.of("./");
 
+  @Option(
+      names = {KEY_STORE_CONFIG_FILE_SIZE_OPTION_NAME},
+      description =
+          "The key store configuration file size in bytes. Useful when loading a large number of configurations from "
+              + "the same yaml file. Defaults to (${DEFAULT-VALUE}) 100MB",
+      paramLabel = "<NUMBER>",
+      defaultValue = "104857600",
+      converter = YamlMapperCliConverter.class,
```
siladu, line 100: What happens if you set this with a config file instead? Or if you don't set it at all - how/when does the YamlMapper get created in those situations?

core/src/main/java/tech/pegasys/web3signer/core/config/Config.java
```diff
@@ -37,6 +38,8 @@ public interface Config {
 
   Path getKeyConfigPath();
 
+  YAMLMapper getKeyConfigYamlMapper();
```
siladu, line 41: Seems a reasonable approach for this isolated usage of YAMLMapper. 
If we needed other instances of YAMLMapper in the prod code, they would exist outside of this config I guess?

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -89,12 +91,10 @@ public class Web3SignerBaseCommand implements Config, Runnable {
   private Path keyStorePath = Path.of("./");
 
   @Option(
-      names = {"--key-store-config-file-size"},
+      names = {KEY_STORE_CONFIG_FILE_SIZE_OPTION_NAME},
       description =
-          "The key store configuration file size in bytes. Useful when loading large number of "
-              + "configurations from same yaml file. Defaults to (${DEFAULT-VALUE}) 100MB",
+          "The key store configuration file size in bytes. Useful when loading a large number of configurations from the same yaml file. Defaults to (${DEFAULT-VALUE}) 100MB",
```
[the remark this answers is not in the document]
usmansaleem: agreed with your point 👍🏼

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
usmansaleem: Config file and cli both eventually ends up here. If we don't set it either via config file and cli, the default value gets assigned.

core/src/main/java/tech/pegasys/web3signer/core/config/Config.java
usmansaleem: correct ... as you can see the test code uses YamlMapperFactory directly for unit/integration testing. The main goal is to keep all the configurations for creating YamlMapper at one spot.

commandline/src/main/java/tech/pegasys/web3signer/commandline/convertor/YamlMapperCliConverter.java
usmansaleem: I don't have strong opinion on that, I initially had it in Web3SignerBaseCommand, can move it back there again.

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
siladu: nice!

siladu (approved):
thanks for the changes - LGTM!

commandline/src/main/java/tech/pegasys/web3signer/commandline/convertor/YamlMapperCliConverter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static tech.pegasys.web3signer.commandline.Web3SignerBaseCommand.KEY_STORE_CONFIG_FILE_SIZE_OPTION_NAME;
+
+import tech.pegasys.web3signer.signing.config.metadata.parser.YamlMapperFactory;
+
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import picocli.CommandLine;
+import picocli.CommandLine.TypeConversionException;
+
+public class YamlMapperCliConverter implements CommandLine.ITypeConverter<YAMLMapper> {
```
jframe, line 23: Don't really this approach. This is using picocli convertor to initialise part of our application. Would prefer that picocli was just used for the configuration. Transforming a --key-store-config-file-max-size to a YamlMapper doesn't make sense to me. It's a value that is used as part of creating it not converting to another type.

Having said that don't dislike it enough to block merging this PR.

core/src/main/java/tech/pegasys/web3signer/core/config/Config.java
jframe, line 41: This shouldn't be part of the config. To me, this class is meant to represent the configuration that is passed into the application. The YamlMapper it the output of that.

jframe (approved):
LGTM

PR #735, opened 2023-03-09
Fix slf4j 2 logging

jframe:
## PR Description
Use slf4j2 API in the log4j bridge. This allows logging from libraries that are using slf4j2 in particular jdbi logs weren't being logged.

The slf4j 2 API is backwards compatible with slf4j 1.

It also fixes the errors on startup complaining about missing SLF4J providers
```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J: Ignoring binding found at [jar:file:/Users/jframe/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-slf4j-impl/2.19.0/1a0c9615ba9fd5b96db8c1136afbef4394286e93/log4j-slf4j-impl-2.19.0.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
```

## Documentation

## Changelog

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,5 +33,5 @@ dependencies {
   implementation 'org.apache.logging.log4j:log4j-api'
 
   runtimeOnly 'org.apache.logging.log4j:log4j-core'
-  runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+  runtimeOnly 'org.apache.logging.log4j:log4j-slf4j2-impl'
 }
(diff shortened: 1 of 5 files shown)


siladu (approved):
nice!

usmansaleem (approved):
LGTM

PR #746, opened 2023-03-24
Fix healthcheck assertion in Azure AT

usmansaleem:
## PR Description
Fix healthcheck assertion in Azure AT

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/bulkloading/AzureKeyVaultAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/bulkloading/AzureKeyVaultAcceptanceTest.java
@@ -12,6 +12,7 @@
  */
 package tech.pegasys.web3signer.tests.bulkloading;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.hamcrest.Matchers.contains;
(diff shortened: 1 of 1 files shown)


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/bulkloading/AzureKeyVaultAcceptanceTest.java
```diff
@@ -60,12 +62,31 @@ void ensureSecretsInKeyVaultAreLoadedAndReportedViaPublicKeysApi() {
...
     response.then().statusCode(200).contentType(ContentType.JSON).body("", contains(EXPECTED_KEY));
 
-    signer
-        .healthcheck()
+    // Since our Azure vault contains some invalid keys, the healthcheck would return 503.
+    final Response healthcheckResponse = signer.healthcheck();
+    healthcheckResponse
         .then()
-        .statusCode(200)
+        .statusCode(503)
```
jframe, line 69: This is ok for now in fixing the build. But the expectation shouldn't be that we expect to fail loading keys i.e. have a 503 status code. Otherwise, we can't tell that having all valid keys works correctly. We could add a tag or use a prefix for the invalid key so they aren't used for this test.

jframe (approved):

PR #748, opened 2023-03-28
Azure bulk loading - filters results using tags

usmansaleem:
## PR Description
Azure bulk loading - filter results based on tags. Use cli option to specify the tag names and values pair. Added acceptance test to return results based on tag.

```
      --azure-secrets-tags=<TAG_NAME=TAG_VALUE>
                            Specify optional tags to filter the secrets. For
                              example --azure-secrets-tags ENV=prod
                              --azure-secrets-tags=CLIENT=A.
```
## Documentation

## Changelog

### DIFF
--- a/commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliAzureKeyVaultParameters.java
+++ b/commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliAzureKeyVaultParameters.java
@@ -15,6 +15,9 @@ package tech.pegasys.web3signer.commandline;
 import tech.pegasys.web3signer.signing.config.AzureAuthenticationMode;
 import tech.pegasys.web3signer.signing.config.AzureKeyVaultParameters;
 
+import java.util.Collections;
+import java.util.Map;
+
 import picocli.CommandLine.Option;
 
 public class PicoCliAzureKeyVaultParameters implements AzureKeyVaultParameters {
@@ -61,6 +64,14 @@ public class PicoCliAzureKeyVaultParameters implements AzureKeyVaultParameters {
       paramLabel = "<CLIENT_SECRET>")
   private String clientSecret;
 
+  @Option(
+      names = {"--azure-secrets-tags"},
+      description =
+          "Specify optional tags to filter the secrets. "
+              + "For example --azure-secrets-tags ENV=prod --azure-secrets-tags=CLIENT=A.",
+      paramLabel = "<TAG_NAME=TAG_VALUE>")
+  private Map<String, String> tags = Collections.emptyMap();
+
   @Override
   public boolean isAzureKeyVaultEnabled() {
     return azureKeyVaultEnabled;
@@ -89,5 +100,10 @@ public class PicoCliAzureKeyVaultParameters implements AzureKeyVaultParameters {
   @Override
   public String getClientSecret() {
     return clientSecret;
+  }
+
+  @Override
+  public Map<String, String> getTags() {
+    return tags;
   }
 }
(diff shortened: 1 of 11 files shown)


signing/src/main/java/tech/pegasys/web3signer/signing/config/AzureKeyVaultParameters.java
```diff
@@ -25,4 +28,8 @@ public interface AzureKeyVaultParameters {
   String getClientId();
 
   String getClientSecret();
+
+  default Map<String, String> getTags() {
```
jframe, line 32: This default method is only used in one place the AzureSecretSigningMetadata. Would be clearer to just return the emptyMap for that one case.

signing/src/main/java/tech/pegasys/web3signer/signing/config/AzureKeyVaultParameters.java
```diff
@@ -12,6 +12,9 @@
  */
 package tech.pegasys.web3signer.signing.config;
 
+import java.util.Collections;
+import java.util.Map;
+
 public interface AzureKeyVaultParameters {
```
jframe, line 18: The AwsKeySigningMetadata doesn't have an implementation so we won't be able to use tags in the key config files

signing/src/main/java/tech/pegasys/web3signer/signing/config/AzureKeyVaultParameters.java
usmansaleem: agreed, will make change.

signing/src/main/java/tech/pegasys/web3signer/signing/config/AzureKeyVaultParameters.java
usmansaleem: tags only make sense for bulk load. For key config file, we always specify the "secret name" to load.

commandline/src/main/java/tech/pegasys/web3signer/commandline/PicoCliAzureKeyVaultParameters.java
```diff
@@ -61,6 +64,14 @@ public class PicoCliAzureKeyVaultParameters implements AzureKeyVaultParameters {
       paramLabel = "<CLIENT_SECRET>")
   private String clientSecret;
 
+  @Option(
+      names = {"--azure-secrets-tags"},
+      description =
+          "Specify optional tags to filter the secrets. "
+              + "For example --azure-secrets-tags ENV=prod --azure-secrets-tags=CLIENT=A.",
+      paramLabel = "<TAG_NAME=TAG_VALUE>")
```
siladu, line 72: AWS implements --aws-secrets-tag-names-filter and --aws-secrets-tag-values-filter
so this is inconsistent from a UX perspective, though not sure how much that matters in this instance.

Is there a reason this was implemented differently?

jframe (approved):

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/AzureSecretSigningMetadata.java
```diff
@@ -87,4 +90,11 @@ public String getSecretName() {
   public ArtifactSigner createSigner(final ArtifactSignerFactory factory) {
     return factory.create(this);
   }
+
+  @Override
+  public Map<String, String> getTags() {
+    // tags support is not applicable for config file mode as
+    // user is already providing the secret name to load the secret from.
+    return Collections.emptyMap();
```
siladu, line 98: Could/should we throw unimplemented exception here? Feels like we should log a warning too.

PR #756, opened 2023-04-17
Upgrade jackson and vertx to upgrade snakeyaml to 2.0

siladu:


usmansaleem (approved):
LGTM

gradle/versions.gradle
```diff
@@ -13,8 +13,8 @@
 
 dependencyManagement {
   dependencies {
-    dependency 'com.fasterxml.jackson.core:jackson-databind:2.14.2'
-    dependency 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.14.2'
+    dependency 'com.fasterxml.jackson.core:jackson-databind:2.15.0-rc3'
```
jframe, line 16: Normally rather not use an RC but in this case fixing a CVE think it's worth it

jframe (approved):
LGTM

gradle/versions.gradle
siladu: Created an issue as a reminder: https://github.com/ConsenSys/web3signer/issues/757

PR #775, opened 2023-05-03
Eth RPC proxy

jframe:
## PR Description
This is the first part of adding an RPC proxy to Web3Signer for the execution client RPCs. It adds the base functionality of http proxy in eth1 mode disabled by default behind a feature toggle. It does not include integration and acceptance tests. This will be added in a subsequent PR.

The RPC proxy will proxy any unmatched RPC requests to a downstream execution client. By having the proxy it means that all RPC requests can go through web3signer with web3signer providing additional RPCs for eth_sendtransaction and eth_accounts (not part of this PR).

It is disabled by default behind a feature flag --downstream-http-proxy-enabled.

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/KeyLoadAndSignAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/KeyLoadAndSignAcceptanceTest.java
@@ -59,9 +59,9 @@ public class KeyLoadAndSignAcceptanceTest extends SigningAcceptanceTestBase {
   public void receiveA404IfRequestedKeyDoesNotExist(final KeyType keyType)
       throws JsonProcessingException {
     if (keyType == KeyType.BLS) {
-      setupEth1Signer();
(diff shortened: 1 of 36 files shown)


gfukushima (approved):
lgtm

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/KeyLoadAndSignAcceptanceTest.java
```diff
@@ -59,9 +59,9 @@ public class KeyLoadAndSignAcceptanceTest extends SigningAcceptanceTestBase {
   public void receiveA404IfRequestedKeyDoesNotExist(final KeyType keyType)
       throws JsonProcessingException {
     if (keyType == KeyType.BLS) {
-      setupEth1Signer();
+      setupEth2Signer(Eth2Network.MINIMAL, SpecMilestone.PHASE0);
```
gfukushima, line 62: Don't fully understand why we reversed this logic.

PR #785, opened 2023-05-11
fix for nightly trivy

joshuafernandes:
## PR Description
Fix for nightly trivy - remove the test files and only use the folder names to pull down the containers and test


siladu (approved):
LGTM

Issue #763, opened 2023-04-24
Eth RPC proxy to web3 provider


jframe:
Proxy Ethereum RPC requests to the web3 provider so that web3signer can be used for all Ethereum RPCs.

This mirrors the functionality of the RPC pass-through/proxy in EthSigner.

**Tasks**
- [x] Web3 provider configuration in Eth1 subcommand for downstream host, proxy and TLS options
- [x] Add route to Vertx in Eth1Runner for unhandled HTTP requests
- [x] Proxy HTTP handler. Can use [pass-through handler](https://github.com/ConsenSys/ethsigner/blob/master/ethsigner/core/src/main/java/tech/pegasys/ethsigner/core/requesthandler/passthrough/PassThroughHandler.java) from EthSigner
- [x] Proxy support. Code can be adapted from EthSigner
- [x] TLS support. Code can be adapted from EthSigner
- [x] Acceptance tests
- [x] Restrict proxy requests to RPC requests
- [x] Documentation

PR #788, opened 2023-05-18
Only proxy RPC requests

jframe:
## PR Description
Ensures the Eth 1 proxy functionality only applies to RPC requests. Any unhandled rest/json requests will not be proxied to the downstream provider.

Also removed the --downstream-http-proxy-enabled flag now this restricted to RPCs.

## Fixed Issue(s)

part of #763 

## Documentation

## Changelog

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/PassThroughHandler.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/PassThroughHandler.java
@@ -15,6 +15,7 @@ package tech.pegasys.web3signer.core.service.jsonrpc.handlers;
 import tech.pegasys.web3signer.core.service.ForwardedMessageResponder;
 import tech.pegasys.web3signer.core.service.VertxRequestTransmitter;
 import tech.pegasys.web3signer.core.service.VertxRequestTransmitterFactory;
+import tech.pegasys.web3signer.core.service.jsonrpc.JsonDecoder;
 import tech.pegasys.web3signer.core.service.jsonrpc.JsonRpcRequest;
 import tech.pegasys.web3signer.core.service.jsonrpc.JsonRpcRequestHandler;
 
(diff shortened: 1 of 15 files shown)


core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/PassThroughHandler.java
```diff
@@ -52,6 +62,17 @@ public void handle(final RoutingContext context) {
         request.method(), headersToSend, request.path(), context.getBodyAsString());
   }
 
+  private boolean isRpc(final RoutingContext context) {
+    final HttpMethod method = context.request().method();
+    try {
+      final JsonRpcRequest request =
+          jsonDecoder.decodeValue(context.body().buffer(), JsonRpcRequest.class);
+      return method.equals(HttpMethod.POST) && request.getVersion().equals("2.0");
```
siladu, line 70: nit: Is being able to deserialize to JsonRpcRequest sufficient to return true? 

Taking that even further, perhaps could use `mapper.canDeserialize()` though not sure that saves anything compared to full deserialization

siladu (approved):

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/PassThroughHandler.java
jframe: For Ethereum it always versions "2.0" with a post request, so that is why I added the additional constraint.

Looking at `mapper.canDeserialize()` https://javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.15.1/com/fasterxml/jackson/databind/ObjectMapper.html#canDeserialize-com.fasterxml.jackson.databind.JavaType-. It doesn't look it will do what I want. It will tell me whether Jackson knows whether understands how to deserialise a JsonRPC, that is it has the appropriate deserialiser. But it won't tell me if the body request can be deserialised as a jsonRPC. It doesn't take any input just a type.

PR #796, opened 2023-06-02
Support prometheus push metrics 

### LINKED ISSUE #790: Implement prometheus push gateway
**Description**

We should be able to push metrics to Prometheus rather than just exposing the same metrics through /metrics, if possible. It would help with different Prometheus configurations and also we would need just one port open, the one for the signer itself.

siladu:
Add configuration for and pass through to Besu library, tested with Besu's instructions https://besu.hyperledger.org/en/stable/public-networks/how-to/monitor/metrics/#run-prometheus-with-besu-in-push-mode

Example CLI Usage: 
```
web3signer --metrics-push-enabled --metrics-push-port 9091 --metrics-push-host=127.0.0.1  \
--metrics-push-interval=30 --metrics-push-prometheus-job="web3signer-job" \
eth2 --slashing-protection-enabled=false
```

Example config.yaml Usage `web3signer --config-file=config.yaml eth2`:
```
metrics-push-enabled: true
metrics-push-port: 9091
metrics-push-host: 127.0.0.1
metrics-push-interval: 60
metrics-push-prometheus-job: "web3signer-job"

eth2.slashing-protection-enabled: false
```

Minimal Usage (using defaults):
```
web3signer --metrics-push-enabled eth2 --slashing-protection-enabled=false
```

Defaults:

```
--metrics-push-enabled Enable the metrics push gateway integration
                               (default: false)
--metrics-push-host=<HOST>
                             Host of the Prometheus Push Gateway for push mode
                               (default: 127.0.0.1)
--metrics-push-port=<PORT>
                             Port of the Prometheus Push Gateway for push mode
                               (default: 9001)
--metrics-push-interval=<INTEGER>
                             Interval in seconds to push metrics when in push
                               mode (default: 15)
--metrics-push-prometheus-job=<metricsPrometheusJob>
                             Job name to use when in push mode (default:
                               web3signer-job)
```

## Fixed Issue(s)

Fixes #790 

## Documentation

## Changelog


commandline/src/test/java/tech/pegasys/web3signer/commandline/CommandlineParserTest.java
```diff
@@ -131,6 +131,33 @@ void eth2SubcommandRequiresSlashingDatabaseUrlWhenSlashingEnabled() {
...
+
+    parser.registerSubCommands(new MockEth2SubCommand());
+    final int result = parser.parseCommandLine(cmdline.split(" "));
+
+    assertThat(result).isZero();
+    assertThat(commandError.toString()).isEmpty();
+  }
+
+  @Test
+  void metricsFailsToParse() {
```
jframe, line 148:
```suggestion
  void metricsEnabledWithMetricsPushedEnabledFailsToParse() {
```

core/src/main/java/tech/pegasys/web3signer/core/config/BaseConfig.java
```diff
@@ -49,6 +49,16 @@ public interface BaseConfig {
...
+  Boolean isMetricsPushEnabled();
+
+  String getMetricsPushHost();
+
+  Integer getMetricsPushPort();
+
+  Integer getMetricsPushIntervalSeconds();
+
+  String getMetricsPrometheusJob();
+
```
usmansaleem, line 61: nit: `MetricsPushOptions` value object, similar to TlsOptions, can also be used to eliminate need of multiple methods in BaseConfig.

usmansaleem (approved):
LGTM

jframe (approved):

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -266,6 +270,14 @@ public List<String> getMetricsHostAllowList() {
     return metricsHostAllowList;
   }
 
+  @Override
+  public Optional<MetricsPushOptions> getMetricsPushOptions() {
+    if (metricsPushOptions.isMetricsPushEnabled()) {
```
usmansaleem, line 275: From memory, the mixin might have some weird side effects if they are not specified on command line. Make sure that `metricsPushOptions` is not `null` before calling `isMetricsPushEnabled`

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
```diff
@@ -334,6 +346,13 @@ public void validateArgs() {
     final PicoCliTlsServerOptionsValidator picoCliTlsServerOptionsValidator =
         new PicoCliTlsServerOptionsValidator(spec, picoCliTlsServerOptions);
     picoCliTlsServerOptionsValidator.validate();
+
+    if (isMetricsEnabled() && metricsPushOptions.isMetricsPushEnabled()) {
```
usmansaleem, line 350: same as above, might be a good idea to either check for null of call getMetricsPushOptions method.

commandline/src/main/java/tech/pegasys/web3signer/commandline/Web3SignerBaseCommand.java
usmansaleem: However, if there is an AT which is testing the default behavior of metrics push disabled, then nothing to change here.

usmansaleem (approved):
LGTM - minor comment around checking for null.

jframe (approved):

bgravenorst:
Removing label because this issue is now in the Doc backlog.

PR #816, opened 2023-06-22
Optimize pruning connection creation

### LINKED ISSUE #672: Don't create pruning database connection when pruning is disabled
Don't create pruning database connection when pruning is disabled.

See `tech.pegasys.web3signer.slashingprotection.SlashingProtectionContextFactory` where the pruning connection is established without checking whether pruning is enabled or not.

gfukushima:
## PR Description
This PR simply adds a check to avoid creating a connection to the DB for pruning when pruning is disabled. 

## Fixed Issue(s)

Fixes #672 

## Documentation

## Changelog

## Testing

### DIFF
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbPruner.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbPruner.java
@@ -32,13 +32,13 @@ import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
 
 public class DbPruner {
-  private final Jdbi jdbi;
+  private final Optional<Jdbi> jdbi;
   private final SignedBlocksDao signedBlocksDao;
   private final SignedAttestationsDao signedAttestationsDao;
   private final LowWatermarkDao lowWatermarkDao;
 
   public DbPruner(
-      final Jdbi jdbi,
+      final Optional<Jdbi> jdbi,
       final SignedBlocksDao signedBlocksDao,
       final SignedAttestationsDao signedAttestationsDao,
       final LowWatermarkDao lowWatermarkDao) {
@@ -60,6 +60,10 @@ public class DbPruner {
   }
 
   private void pruneBlocks(final int validatorId, final long slotsToKeep) {
+
+    final Jdbi jdbi =
+        this.jdbi.orElseThrow(() -> new NullPointerException("Pruner has no connection created"));
+
     final boolean hasWatermark =
         jdbi.inTransaction(
             READ_UNCOMMITTED,
@@ -92,6 +96,9 @@ public class DbPruner {
   }
 
   private void pruneAttestations(final int validatorId, final long epochsToKeep) {
+    final Jdbi jdbi =
+        this.jdbi.orElseThrow(() -> new NullPointerException("Pruner has no connection created"));
+
     final boolean hasWatermark =
         jdbi.inTransaction(
             READ_UNCOMMITTED,
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbPruner.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbPruner.java
@@ -62,7 +62,7 @@ public class DbPruner {
   private void pruneBlocks(final int validatorId, final long slotsToKeep) {
 
     final Jdbi jdbi =
-        this.jdbi.orElseThrow(() -> new NullPointerException("Pruner has no connection created"));
+        this.jdbi.orElseThrow(() -> new IllegalStateException("Pruner has no connection created"));
 
     final boolean hasWatermark =
         jdbi.inTransaction(
@@ -97,7 +97,7 @@ public class DbPruner {
 
   private void pruneAttestations(final int validatorId, final long epochsToKeep) {
     final Jdbi jdbi =
-        this.jdbi.orElseThrow(() -> new NullPointerException("Pruner has no connection created"));
+        this.jdbi.orElseThrow(() -> new IllegalStateException("Pruner has no connection created"));
 
     final boolean hasWatermark =
         jdbi.inTransaction(
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
@@ -55,7 +55,7 @@ public class DbSlashingProtection implements SlashingProtection {
   private final InterchangeManager interchangeManager;
   private final LowWatermarkDao lowWatermarkDao;
   private final GenesisValidatorRootValidator gvrValidator;
-  private final DbPruner dbPruner;
+  private final Optional<DbPruner> dbPruner;
   private final long pruningEpochsToKeep;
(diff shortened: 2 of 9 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionContextFactory.java
```diff
@@ -38,12 +38,14 @@ public static SlashingProtectionContext create(
...
-            slashingProtectionParameters.getPruningDbPoolConfigurationFile(),
-            slashingProtectionParameters.isDbConnectionPoolEnabled());
+        slashingProtectionParameters.isPruningEnabled()
+            ? DbConnection.createPruningConnection(
+                slashingProtectionParameters.getDbUrl(),
+                slashingProtectionParameters.getDbUsername(),
+                slashingProtectionParameters.getDbPassword(),
+                slashingProtectionParameters.getPruningDbPoolConfigurationFile(),
+                slashingProtectionParameters.isDbConnectionPoolEnabled())
+            : null;
```
usmansaleem, line 48: its not evident from here, is there a null guard in DbSlashingProtection for pruningJdbi? A better alternative might be to modify DBSlashingProtecting constructor to take `Optional(Jdbi) pruningJdbi` which would make intent more clear that pruningJdbi is optional.

usmansaleem: Similarly, SlashingProtectionContext should also take `Optional<Jdbi> pruningJdbi` and the usage of it should take appropriate action only if it is present.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbPruner.java
```diff
@@ -60,6 +60,10 @@ public void pruneForValidator(
   }
 
   private void pruneBlocks(final int validatorId, final long slotsToKeep) {
+
+    final Jdbi jdbi =
+        this.jdbi.orElseThrow(() -> new NullPointerException("Pruner has no connection created"));
```
usmansaleem, line 65: if (jdbi.isEmpty()) {
return; //no-op since jdbi is empty/pruning is disabled
}

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionContextFactory.java
gfukushima: Done some changes, can you please see if that matches your suggestion?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbPruner.java
gfukushima: We've talked about and I'd rather throw an Exception here to let whoever is trying to call the pruner with no connection that something is wrong rather than just doing nothing.

usmansaleem (approved):
LGTM

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionContextFactoryTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.jdbi.v3.core.Jdbi;
+import org.junit.jupiter.api.Test;
+
+class SlashingProtectionContextFactoryTest {
+
+  private static final String USERNAME = "postgres";
+  private static final String PASSWORD = "postgres";
+
+  @Test
+  public void dontCreatPruningConnectionWhenPrunningIsDisabled() {
```
siladu, line 30:
```suggestion
  public void dontCreatePruningConnectionWhenPruningIsDisabled() {
```

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/SlashingProtectionContextFactoryTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    SlashingProtectionContext factory =
+        SlashingProtectionContextFactory.create(
+            new TestSlashingProtectionParameters(
+                testDatabaseInfo.databaseUrl(), USERNAME, PASSWORD, false));
+    assertThat(factory.getPruningJdbi()).isEqualTo(Optional.empty());
+  }
+
+  @Test
+  public void creatPruningConnectionWhenPrunningIsEnabled() {
```
siladu, line 41:
```suggestion
  public void createPruningConnectionWhenPruningIsEnabled() {
```

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -86,8 +86,12 @@ public DbSlashingProtection(
             signedAttestationsDao,
             metadataDao,
             lowWatermarkDao);
+
     this.dbPruner =
-        new DbPruner(pruningJdbi, signedBlocksDao, signedAttestationsDao, lowWatermarkDao);
+        pruningJdbi != null
+            ? Optional.of(
+                new DbPruner(pruningJdbi, signedBlocksDao, signedAttestationsDao, lowWatermarkDao))
+            : Optional.empty();
```
siladu, line 94: pruningJdbi is an optional but we're checking for nullability so don't think this will actually work?

siladu: I don't like using two optionals for this, think it would be better with 1 optional by passing the optionally constructed DbPruner into this class, i.e. pull up the construction of the DbPruner class that calls this constructor.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
```diff
@@ -247,6 +251,9 @@ public boolean maySignBlock(
 
   @Override
   public void prune() {
+    if (dbPruner.isEmpty()) {
+      throw new IllegalStateException("prune() called when dbPruner isn't initialized");
+    }
```
siladu, line 256: Feels a bit messy having a prune method check whether prune is enabled, would be better IMO to Decorate DbSlashingProtection with the pruning functionality, e.g. wrap it in PruningDbSlashingProtection and add the prune method to that. That would get rid of all the optionals then too.

siladu (changes requested):
Some suggestions in the comments. Requesting changes because I don't see how mixing Optional and null check is actually working. I think it's a pretty small refactoring to use the Decorator pattern for the pruning logic and end result will be much cleaner.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbSlashingProtection.java
gfukushima: Spiked on this please see https://github.com/ConsenSys/web3signer/pull/820

gfukushima: Closing this PR in favour of #823

gfukushima:
Closing in favour of #823

PR #817, opened 2023-06-22
Hashicorp connection, override http protocol version

### LINKED ISSUE #811: [23.6.0] vault-agent - http2: invalid Upgrade request header: ["h2c"]
Just upgraded to Web3Signer v23.6.0 and started seeing this error in my [vault-agent](https://developer.hashicorp.com/vault/docs/agent) proxy logs:

> `2023-06-19T10:30:06.722Z [ERROR] agent.apiproxy.client: request failed: error="Get \"https://vault.example.com/v1/secret/data/goerli/12345\": http2: invalid Upgrade request header: [\"h2c\"]" method=GET url=https://vault.example.com/v1/secret/data/goerli/12345`

This error was not happening in v23.3.1 so seems to be a new bug.

Config example:

```yaml
keyName: key
keyPath: /v1/secret/data/goerli/12345
keyType: BLS
serverHost: localhost
serverPort: 8200
timeout: 30000
tlsEnabled: false
token: unused_handled_by_vault_agent
type: hashicorp
```

Where `localhost:8200` is the local port my Vault Agent is listening on.

usmansaleem:
## PR Description
Hashicorp connection, override http protocol version. Hashicorp Signing Metadata file can now optionally take `httpProtocolVersion` with possible values of `HTTP_1_1` and `HTTP_2`. It defaults to `HTTP_2`. This should help in situations where vault is behind load balancer/vault agent and unable to handle Http 2 connections over h2c.

## Fixed Issue(s)

fixes #811 

## Documentation

## Changelog

## Testing

### DIFF
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
@@ -107,7 +107,9 @@ public class MetadataFileHelpers {
   }
 
   public void createHashicorpYamlFileAt(
-      final Path metadataFilePath, final HashicorpSigningParams node) {
+      final Path metadataFilePath,
+      final HashicorpSigningParams node,
+      Optional<String> httpProtocolVersion) {
     try {
       final Map<String, String> signingMetadata = new HashMap<>();
 
@@ -130,6 +132,8 @@ public class MetadataFileHelpers {
       signingMetadata.put("keyName", node.getSecretName());
       signingMetadata.put("token", node.getVaultToken());
       signingMetadata.put("keyType", node.getKeyType().toString());
+
+      httpProtocolVersion.ifPresent(version -> signingMetadata.put("httpProtocolVersion", version));
 
       createYamlFile(metadataFilePath, signingMetadata);
     } catch (final Exception e) {
--- a/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
@@ -50,6 +50,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariables;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.EnumSource;
 
 public class BlsSigningAcceptanceTest extends SigningAcceptanceTestBase {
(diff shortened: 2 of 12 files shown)


keystorage/src/main/java/tech/pegasys/web3signer/keystorage/hashicorp/config/ConnectionParameters.java
```diff
@@ -58,4 +59,49 @@ public long getTimeoutMilliseconds() {
   public URI getVaultURI() {
     return vaultURI;
   }
+
+  public HttpClient.Version getHttpProtocolVersion() {
+    return httpProtocolVersion;
+  }
+
+  public static final class Builder {
```
siladu, line 67: nice improvement 👍

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -131,6 +133,8 @@ public void createHashicorpYamlFileAt(
       signingMetadata.put("token", node.getVaultToken());
       signingMetadata.put("keyType", node.getKeyType().toString());
 
+      httpProtocolVersion.ifPresent(version -> signingMetadata.put("httpProtocolVersion", version));
```
siladu, line 136: nit: "magic string" - is there not a type we can use here? e.g. `HttpClient.Version`?

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
```diff
@@ -120,7 +121,32 @@ public void ableToSignUsingHashicorp() throws JsonProcessingException {
...
+          Optional.empty());
+
+      signAndVerifySignature(ArtifactType.BLOCK);
+    } finally {
+      hashicorpNode.shutdown();
+    }
+  }
+
+  @ParameterizedTest(name = "{index} - Using http protocol version: {0}, with TLS: {1}")
+  @CsvSource({"HTTP_1_1, true", "HTTP_2, true", "HTTP_1_1, false", "HTTP_2, false"})
```
siladu, line 134: It appears this is doing a test for each permutation of protocol and withTLS. In what way is the assert condition changing for each of these?

I am surprised that (HTTP_2, false) is working because wasn't that the case that wasn't working for the user?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/hashicorp/HashicorpConnectionFactory.java
```diff
@@ -52,6 +52,8 @@ private HttpClient getHttpClient(ConnectionParameters connectionParameters) {
         _key -> {
           final HttpClient.Builder httpClientBuilder =
               HttpClient.newBuilder()
+                  .followRedirects(HttpClient.Redirect.NORMAL)
```
siladu, line 55: Why do we need this now?

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
usmansaleem: this is part of our Acceptance Test DSL class. All the keys and values are string literals.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
usmansaleem: Http2 + Hashicorp Vault (nonTLS) is supposed to work (as in our acceptance test example). The scenario that *might not* work is Web3Signer -> Hashicorp Vault Agent (nonTLS) -> Load Balancer (TLS) -> Hashicorp Vault (TLS or nonTLS). In this scenario, Http2 over h2c (plain text) may fail by the Load balancer because Hashicorp Vault Agent forward this header when it is supposed to eliminate it.

usmansaleem: The assertion is that we are able to connect to Hashicorp Vault with both Http1.1 and Http/2 on TLS and nonTLS.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/hashicorp/HashicorpConnectionFactory.java
usmansaleem: The default in Java's HttpClient is that redirects are disabled. Changing this default means that if there is a proxy that redirects connection to a different host, HttpClient will automagically follow the redirect.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/hashicorp/config/ConnectionParameters.java
usmansaleem: Yes, I realised that this class is used in our tests as well and we were forced to set values that were not required by the specific tests. Builder patterns neatly allows the defaults to be set.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
siladu: > Http2 + Hashicorp Vault (nonTLS) is supposed to work (as in our acceptance test example).
> The assertion is that we are able to connect to Hashicorp Vault with both Http1.1 and Http/2 on TLS and nonTLS.

This is why I'm confused - I was expecting HTTP_2 with nonTLS to not be able to connect...am I looking at the wrong test? This test isn't related to Hashicorp?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/hashicorp/HashicorpConnectionFactory.java
siladu: Not currently clear if that is relevant to the HTTP2 problem that this PR is solving or just something you noticed was missing? If that latter, does that mean the original PR to change to Java client also broke some proxy use cases?

siladu (approved):

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/hashicorp/HashicorpConnectionFactory.java
siladu: Discussed with @usmansaleem, it was the latter.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/BlsSigningAcceptanceTest.java
siladu: discussed with @usmansaleem - user's issue involved a proxy in front of the vault which this AT isn't attempting to test.

PR #822, opened 2023-06-27
Add eth_signTransaction

### LINKED ISSUE #767: Eth RPC for eth_signTransaction
Implement the RPC for eth_signTransaction in web3signer in eth1 mode.

https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_signtransaction

gfukushima:
## PR Description
Add JSON RPC method `eth_signTransaction` under the eth1 subcommand
## Fixed Issue(s)

Fixes #767 
## Documentation

## Changelog

## Testing


acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/KeyIdentifiersAcceptanceTestBase.java
```diff
@@ -153,6 +154,8 @@ private void createSecpKey(final String privateKeyHexString) {
   protected void initAndStartSigner(final String mode) {
     final SignerConfigurationBuilder builder = new SignerConfigurationBuilder();
     builder.withKeyStoreDirectory(testDirectory).withMode(mode);
+    if (mode.equals("eth1"))
+      builder.withChainIdProvider(new ConfigurationChainId(DEFAULT_CHAIN_ID));
```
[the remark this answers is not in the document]
siladu: Also, is there a reason why we need to explicitly set this to a default value rather than just using the default when we call build() below?

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/EthSendTransactionJsonParameters.java
```diff
@@ -0,0 +1,124 @@
...
+    this.value = decodeBigInteger(value);
+  }
+
+  @JsonSetter("data")
+  public void data(final String data) {
+    this.data = data;
+  }
+
+  @JsonSetter("privateFrom")
+  public void privateFrom(final String privateFrom) {
```
siladu, line 79: Just wondering if we need this and privateFor

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/EthSendTransactionJsonParameters.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright 2019 ConsenSys AG.
```
siladu, line 2: nit: copyrights on new files

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.web3j.crypto.Sign.SignatureData;
+import org.web3j.crypto.TransactionEncoder;
+import org.web3j.utils.Numeric;
+
+public class TransactionSerializer {
+  private static final Logger LOG = LogManager.getLogger();
+
+  protected final ArtifactSignerProvider signer;
```
siladu, line 33: Why did this interface change from Signer?

core/src/test/java/tech/pegasys/web3signer/core/util/ByteUtilsTest.java
siladu, line 1: Looks like we've already got this test https://github.com/Consensys/web3signer/blob/master/signing/src/test/java/tech/pegasys/web3signer/signing/util/ByteUtilsTest.java

siladu (commented):
Just extra ByteUtilsTest blocking approval, everything else is a nit

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
gfukushima: Signer was an interface that ethSigner used to use from the signers library.
ArtifactSigner is the equivalent on W3S an so are the rest of the classes like ArtifactSignerProvider, ArtifactSignature.

core/src/test/java/tech/pegasys/web3signer/core/util/ByteUtilsTest.java
gfukushima: Thanks for catching this

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/publickeys/KeyIdentifiersAcceptanceTestBase.java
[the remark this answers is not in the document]
gfukushima: removed and added a default to the builder class

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/EthSendTransactionJsonParameters.java
gfukushima: removed since it's currently not used. If any features we add in the future requires this we can add it back.

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
siladu: Even if we hadn't merged the signers library in, w3s would have had access to the Signer interface too (via the lib the same way Ethsigner does) so I'm still confused why we should change this.

siladu: Thought I'd mentioned it because maybe we should delete that Signer interface...
but on closer inspection it looks like ArtifactSigner might be wrapping that Signer interface. 
Maybe not for this PR, but feels like might be a refactoring opportunity here.

siladu (approved):
LGTM

Issue #824, opened 2023-06-27
Should we support env var and yaml versions of short name cli options?


siladu:
The status quo is confusing. W3S kind of half supports them: https://github.com/ConsenSys/web3signer/blob/master/commandline/src/main/java/tech/pegasys/web3signer/commandline/valueprovider/EnvironmentVariableDefaultProvider.java#L36 

If you have aliases of `--logging, -l` then only `--logging` will be supported.
But if you have one name of `-l` then it'll be supported.
I also think using longestName is a bug because `--logging, --log, -l` will only support `--logging`.

Similar question for YAML file support.

Besu explicitly filters out single dash prefixed options https://github.com/hyperledger/besu/blob/f43ae51d6f00984c8f87b67999f7a499987dbd49/besu/src/main/java/org/hyperledger/besu/cli/util/EnvironmentVariableDefaultProvider.java#L57-L58

usmansaleem:
Closing as being stale. Please reopen or comment if this is still an issue.

Issue #586, opened 2022-06-01
Rename key-store-path command line option


jframe:
Rename the command line option --key-store-path as it is confusing since it implies that it is related to key stores. This is the key config directory and is for more than just keystore files. It is also similar in name the recently added keystore bulk loading command line options https://docs.web3signer.consensys.net/en/latest/Reference/CLI/CLI-Subcommands/#keystores-path.

Tasks
- [x] Keep old name as an alias and deprecate
- [x] New name

Note to dev: Make sure to specify the new cli option as the first option so that any error reporting by PicoCli mentions new option, for example:
```
  @Option(
      names = {"--key-config-path", "--key-store-path"},
      description = "The path to a directory storing yaml files defining available keys",
```

The AT should test `key-config-path` and `key-store-path` i.e. both the new and deprecated value and make sure that the config file AT picks them up.

jframe:
@james-prysm No decision has been made on what the name would change to. Was thinking --key-config-path would work well but that's not locked in yet. Will update this ticket once we decide on the new command line option name.

siladu:
@james-prysm we went with --key-config-path but --key-store-path still works

james-prysm:
awesome thanks for letting me know

PR #825, opened 2023-06-27
Support aliases in environment variables and YAML configuration

siladu:
## PR Description

Supporting PR for https://github.com/Consensys/web3signer/pull/826

Before this PR, the only aliases that were supported were the ones with the longest name. This makes sense when your aliases are always short name versions of the options, e.g. `--logging, -l`.

It doesn't make sense when you want to use aliases as a way to rename an option as for https://github.com/Consensys/web3signer/issues/586

This PR allows 'short name options' with a single leading dash to be converted into environment variables and YAML config items. 
I have opened another issue to discuss whether we should handle these differently to the long name options: https://github.com/ConsenSys/web3signer/issues/824 

## Documentation

## Changelog

## Testing

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -119,9 +119,9 @@ dependencyManagement {
       entry 'besu'
       entry ('core') {
         exclude group: 'com.github.jnr', name: 'jnr-unixsocket'
+        exclude group: 'org.bouncycastle', name: 'bcprov-jdk15on'
       }
-      entry 'crypto'
-      entry ('utils') {
+      entry ('crypto') {
         exclude group: 'org.bouncycastle', name: 'bcprov-jdk15on'
       }
     }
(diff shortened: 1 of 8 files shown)


gradle/versions.gradle
```diff
@@ -119,9 +119,9 @@ dependencyManagement {
       entry 'besu'
       entry ('core') {
         exclude group: 'com.github.jnr', name: 'jnr-unixsocket'
+        exclude group: 'org.bouncycastle', name: 'bcprov-jdk15on'
```
gfukushima, line 122: Why is this still needed?

gfukushima (approved):
lgtm

gradle/versions.gradle
siladu: My guess is that since the artefact name itself changed, it's not possible for gradle to automatically resolve to the jdk18 version. We're using a version of web3j that pulls in the old one. I tried upgrading web3j first and it caused other issues so aborted that, but we should revisit.

I believe the utils module is a dependency of core. In an earlier commit I had introduced the utils entry simply to be able to exclude bouncycastle. In this PR I realised I could simplify that by moving the exclude.

usmansaleem (approved):
LGTM

Issue #501, opened 2022-01-27
Eth1 secp256k1 remote signing using AWS KMS


jframe:
Implement Eth1 secp256k1 signing using the AWS KMS so that Ethereum secp private keys can be used to sign transactions without the key leaving the AWS KMS.

Tasks
- [x] Add a new key metadata format. The [AzureKeySigningMetadata](https://github.com/Consensys/web3signer/blob/master/signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/AzureKeySigningMetadata.java) can be used as a guide
- [x] Create an AWS signer to sign secp256k1 keys using KMS with the [AWS SDK](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/java_kms_code_examples.html)
- [x] Acceptance tests

PR #837, opened 2023-07-10
Remote Key signing using Aws KMS

usmansaleem:
## PR Description
Remote Key signing using Aws KMS. Integrate AWS KMS SECP256K1 remote signing.

## Summary of Changes:
**signing/secp256k1/aws**
- `AwsKmsSigner.java`: Implements Signer interface. For each kms configuration metadata file, an instance of this class will be registered.
- `AwsKmsSignerFactory.java`: Create an instance of `AwsKmsSigner` based on `AwsKmsMetadata`. The instance of this class is initialized in `Eth1Runner` and `FilecoinRunner`. This class utilizes `CachedAwsKmsClientFactory`.
- `AwsKmsClient.java`: Wraps Aws Kms Client library. Exposes `sign` and `get ECPublicKey`. Used by `AwsKmsSigner`
- `CachedAwsKmsClientFactory.java`: Factory class for providing cached instances of `AwsKmsClient`. The `AwsKmsClientKey.java` is used as a key for cache entry. The credentials and region are used for equals and hashcode. It is anticipated that same credentials/region will be specified in AwsKmsMetadata yaml files to perform various operations of multiple keys.

**signing/secp256k1/util**
- `Eth1SignatureUtil.java` - Refactored signature calculation for internal usage i.e. R, S and V. Azure return P1363Encoded signature while AWS returns ANS.1/DER encoded signature.
 
**common/config**
`AwsAuthenticationMode.java` is moved into common as this file is common both for Aws Secrets Manager and Aws Kms Client. In addition, `AwsCredentials.java` is also created in common. This file is currently used by Aws Kms Client, however, in future, it should also be used by Aws Secrets Manager related classes.

**signing/config**
- `AwsCredentialsProviderFactory.java`: Factory class that return AWS library's `AwsCredentialsProvider` from AwsAuthenticationMode and AwsCredentials (which are derived from metadata deserializer).

**signing/config/metadata**
- `AwsKmsMetadata.java` - Represents new configuration type `aws-kms`. Supports only `SECP256K1`. 
- `AwsKmsMetadataDeserializer.java` is responsible to convert yaml ->  `AwsKmsMetadata` during metadata files loading. It performs validation of required fields. Very similar to existing`AwsKeySigningMetadataDeserializer.java`.
- A sample yaml file with all the options would look like:
```
type: "aws-kms"
authenticationMode: "SPECIFIED"
accessKeyId: "acc_key_id"
secretAccessKey: "sec_acc_key"
sessionToken: "sess_token"
kmsKeyId: "aaabbbcccddd"
region: "us-east-2"
endpointOverride: "http://localhost:4566"
```

**core**
`Eth1Runner` and `FilecoinRunner` are modified to construct `AwsKmsSignerFactory` with `awsKmsClientCacheSize` and `applySha3Hash` flag.

**commandline/subcommands**
`Eth1SubCommand.java` and `FilecoinSubCommand.java` are modified to introduce cli option `--aws-kms-client-cache-size` which is used to construct `AwsKmsSignerFactory` in the runner classes mentioned above.

## Fixed Issue(s)

#501 

## Documentation

## Changelog

## Testing


keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+      final Optional<URI> endpointOverride) {
+    this.authMode = authMode;
+    this.awsCredentials = awsCredentials;
+    this.region = region;
+    this.endpointOverride = endpointOverride;
+
+    initKmsClient();
+  }
+
+  void initKmsClient() {
```
siladu, line 74: nit: could be private

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+public class AwsKeyManagerService implements AutoCloseable {
+  private final AwsAuthenticationMode authMode;
+  private final AwsCredentials awsCredentials;
+  private final String region;
+
+  private KmsClient kmsClient;
+
+  private Optional<URI> endpointOverride;
+
+  public AwsKeyManagerService(
```
siladu, line 61: nit: we have classes with the full name and some with AwsKMS, think it would be better to be consistent one way or the other for the sake of searchability

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+    this.region = region;
+    this.endpointOverride = endpointOverride;
+
+    initKmsClient();
+  }
+
+  void initKmsClient() {
+    if (kmsClient != null) {
+      return;
+    }
```
siladu, line 77: Is this thread-safe? What would be the consequence of constructing the client more than once for concurrent requests?

siladu: Maybe we should use the Factory to maintain a singleton of this class?

Maybe related question: why is this class constructed from both AwsKMSSigner as well as AwsKMSSignerFactory?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+
+      awsCredentialsProvider = StaticCredentialsProvider.create(awsSdkCredentials);
+    }
+    return awsCredentialsProvider;
+  }
+
+  public ECPublicKey getECPublicKey(final String kmsKeyId) {
+    checkArgument(kmsClient != null, "KmsClient is not initialized");
+
+    // Question ... do we need to set grantTokens?
```
siladu, line 114: Question for yourself or reviewers? :)

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
siladu: Or maybe you intended for the difference in name to reflect two different levels of abstraction?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+    final GetPublicKeyRequest getPublicKeyRequest =
+        GetPublicKeyRequest.builder().keyId(kmsKeyId).build();
+    final GetPublicKeyResponse publicKeyResponse = kmsClient.getPublicKey(getPublicKeyRequest);
+    KeySpec keySpec = publicKeyResponse.keySpec();
+    if (keySpec != KeySpec.ECC_SECG_P256_K1) {
```
siladu, line 119: Our Azure code uses two SignatureAlgorithms (one deprecated) - do we need to worry about that for AWS or is it an Azure specific thing?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+        new X509EncodedKeySpec(publicKeyResponse.publicKey().asByteArray());
+    try {
+      KeyFactory keyFactory = KeyFactory.getInstance("EC");
+      return (ECPublicKey) keyFactory.generatePublic(encodedKeySpec);
+    } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public byte[] sign(final String kmsKeyId, final byte[] data) {
```
siladu, line 133: For Azure and file-based, the sign operation is done at the Signer level of abstraction (in the signing package) - is there a reason this one is pushed down to key vault level of abstraction?

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/AwsKMSMetadata.java
```diff
@@ -0,0 +1,76 @@
...
+    this.authenticationMode = authenticationMode;
+    this.region = region;
+    this.awsCredentials = awsCredentials;
+    this.kmsKeyId = kmsKeyId;
+    this.endpointOverride = endpointOverride;
+  }
+
+  public boolean isEnabled() {
+    return true;
+  }
```
siladu, line 50: Do we need this?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/util/Eth1SignatureUtil.java
siladu, line 1: Nice reuse and method naming 👍

siladu (commented):
First pass comments and questions 👍

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
usmansaleem: I wanted this class to be wrapper around AWS' KmsClient. For secrets manager, we had used name `AwsSecretsManager` ... `secrets manager` is the service we are using. On the same note, I used AwsKeyManagerService, `key manager service` is the service we are using. However, I agree with your initial observation, this can be easily rename to `AwsKMS` as in the signing module, we are using `KMS`.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+      final AwsAuthenticationMode authMode,
+      final AwsCredentials awsCredentials,
+      final String region,
+      final Optional<URI> endpointOverride) {
+    this.authMode = authMode;
+    this.awsCredentials = awsCredentials;
+    this.region = region;
+    this.endpointOverride = endpointOverride;
+
+    initKmsClient();
```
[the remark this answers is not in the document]
usmansaleem: this is not meant to fail as it doesn't perform any actual authentication. Having said that, even if it does fail, since this class is AutoCloseable, the constructor call is meant to be in a try/resource block.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
usmansaleem: yes, this should be private. I was playing with the idea of possibility of re-init kms client from test classes, but we don't require it in current implementation.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
usmansaleem: 1. the `kmsClient != null` invariant is really not required as this initialization is meant to happen via constructor. 
2. We don't want this class to be a singelton. However, in future when we use bulk-loading, this class can be saved in a hashmap or a cache if needed, hence I implemented equals and hashcode methods.
3. In the AwsKMSSignerFactory, we need to obtain the ECPrimaryKey since that is required to construct Signer implementations (such as AzureSigner and AwsKMSSigner). The second construction happens in actual sign call (on-demand) as we need to call the sign method on it. Now that I re-review it, we might be able to avoid this construction by passing the same instance of `AwsKMS` inside `AwsKMSSigner`. (I adapted the logic from `AzureKeyVaultSigner` and `AzureKeyVaultSignerFactory`)

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
usmansaleem: for both :). The API talks about using `grantTokens` to further restrict the authentication ... however, I am not sure whether we want to support it or not. May be we can take it as optional parameters from our meta file?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
usmansaleem: I think that we should not expose the actual Azure or AWS library outside of keystorage module. There is additional logic around `sign`, such as converting to ethereum specific signature etc., that is correctly performed in signing module. `AzureKeyVaultSigner` and `Factory` should not directly use `CryptoClient` directly as well.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/AwsKMSMetadata.java
usmansaleem: oh this will be required for upcoming bulk loading support. But I should remove this method for now.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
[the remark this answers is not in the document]
siladu: If it isn't doing IO yet then should be fine.

Seems like auto close might not work if exception is during construction https://stackoverflow.com/questions/34678538/is-the-close-method-on-a-try-with-resources-idiom-not-called-if-a-constructor-th?noredirect=1&lq=1

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
siladu: Is this a more general AWS feature, i.e. could be used with AWSSecretsManager for eth2 as well? If so, then I'd say it should be split out to another issue, especially if it isn't blocking using eth1 with AWS auth in the same way as we already use it for eth2.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+      } else {
+        awsSdkCredentials =
+            AwsBasicCredentials.create(
+                awsCredentials.getAccessKeyId(), awsCredentials.getSecretAccessKey());
+      }
+
+      awsCredentialsProvider = StaticCredentialsProvider.create(awsSdkCredentials);
+    }
+    return awsCredentialsProvider;
+  }
```
jframe, line 109: Think would be common to both the secrets manager and kms?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
```diff
@@ -0,0 +1,172 @@
...
+    this.authMode = authMode;
+    this.awsCredentials = awsCredentials;
+    this.region = region;
+    this.endpointOverride = endpointOverride;
+
+    initKmsClient();
+  }
+
+  void initKmsClient() {
+    if (kmsClient != null) {
```
jframe, line 75: Why do we need this lazy initialisation of the kmsClient?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
jframe, line 133: Prefer to use the Bytes type instead of a raw byte[]

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/util/Eth1SignatureUtil.java
```diff
@@ -0,0 +1,101 @@
...
+    return new BigInteger(Arrays.copyOfRange(signature, startR + 2, startR + 2 + lengthR));
+  }
+
+  private static BigInteger extractS(byte[] signature) {
+    int startR = (signature[1] & 0x80) != 0 ? 3 : 2;
+    int lengthR = signature[startR + 1];
+    int startS = startR + 2 + lengthR;
+    int lengthS = signature[startS + 1];
+    return new BigInteger(Arrays.copyOfRange(signature, startS + 2, startS + 2 + lengthS));
+  }
```
jframe, line 100: Can we use the ASN decoding from bouncy castle instead? We are using this `FcSecpArtifactSigner` to encode Filecoin public keys

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/awskms/AwsKeyManagerService.java
usmansaleem: correct, this piece is common for both secrets manager and kms. Let me try to refactor it out as a utility class.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
```diff
@@ -0,0 +1,110 @@
...
+    try (final KmsClient kmsClient =
+        AwsKMSClientFactory.createKMSClient(
+            awsCredentialsProvider,
+            awsKMSMetadata.getRegion(),
+            awsKMSMetadata.getEndpointOverride())) {
+      this.ecPublicKey = getECPublicKey(kmsClient, awsKMSMetadata.getKmsKeyId());
+    }
+  }
+
+  public static ECPublicKey getECPublicKey(final KmsClient kmsClient, final String kmsKeyId) {
```
siladu, line 62: nit: add @VisibleForTesting

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
```diff
@@ -0,0 +1,110 @@
...
+      throw new RuntimeException("Unsupported key spec from AWS KMS: " + keySpec.toString());
+    }
+
+    final X509EncodedKeySpec encodedKeySpec =
+        new X509EncodedKeySpec(publicKeyResponse.publicKey().asByteArray());
+    try {
+      KeyFactory keyFactory = KeyFactory.getInstance("EC");
+      return (ECPublicKey) keyFactory.generatePublic(encodedKeySpec);
+    } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+      throw new RuntimeException(e);
```
siladu, line 80: Are the error states tested somewhere?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
```diff
@@ -0,0 +1,110 @@
...
+  private final boolean applySha3Hash;
+
+  public AwsKMSSigner(final AwsKMSMetadata awsKMSMetadata, boolean applySha3Hash) {
+    this.awsKMSMetadata = awsKMSMetadata;
+    this.applySha3Hash = applySha3Hash;
+    awsCredentialsProvider =
+        AwsCredentialsProviderFactory.createAwsCredentialsProvider(
+            awsKMSMetadata.getAuthenticationMode(), awsKMSMetadata.getAwsCredentials());
+    try (final KmsClient kmsClient =
+        AwsKMSClientFactory.createKMSClient(
```
siladu, line 54: nit: not sure it matters but wondering if this would be more testable if a pre-constructed AwsKmsClient instance was passed into this constructor.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
```diff
@@ -0,0 +1,110 @@
...
+import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.services.kms.KmsClient;
+import software.amazon.awssdk.services.kms.model.GetPublicKeyRequest;
+import software.amazon.awssdk.services.kms.model.GetPublicKeyResponse;
+import software.amazon.awssdk.services.kms.model.KeySpec;
+import software.amazon.awssdk.services.kms.model.MessageType;
+import software.amazon.awssdk.services.kms.model.SignRequest;
+import software.amazon.awssdk.services.kms.model.SigningAlgorithmSpec;
+
+public class AwsKMSSigner implements Signer {
```
siladu, line 40: nit: inconsistent initialisation capitalisation (AwsKmsSigner I think is convention over AWSKMSSigner)

signing/src/main/java/tech/pegasys/web3signer/signing/config/AwsCredentialsProviderFactory.java
```diff
@@ -0,0 +1,79 @@
...
+    return awsCredentialsProvider;
+  }
+
+  /**
+   * Create static credentials provider using session credentials or basic credentials
+   *
+   * @param awsCredentials AwsCredentials
+   * @return instance of AwsCredentialsProvider
+   */
+  private static AwsCredentialsProvider getStaticCredentialsProvider(
```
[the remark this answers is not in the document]
usmansaleem: yes, for historical reason, we had chosen SPECIFIED instead of STATIC. Since this is a private method, I use this naming to make the intent clear.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
```diff
@@ -0,0 +1,110 @@
...
+    this.applySha3Hash = applySha3Hash;
+    awsCredentialsProvider =
+        AwsCredentialsProviderFactory.createAwsCredentialsProvider(
+            awsKMSMetadata.getAuthenticationMode(), awsKMSMetadata.getAwsCredentials());
+    try (final KmsClient kmsClient =
+        AwsKMSClientFactory.createKMSClient(
+            awsCredentialsProvider,
+            awsKMSMetadata.getRegion(),
+            awsKMSMetadata.getEndpointOverride())) {
+      this.ecPublicKey = getECPublicKey(kmsClient, awsKMSMetadata.getKmsKeyId());
```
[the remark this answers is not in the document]
usmansaleem: I concur .... let me take it out and pass constructed EC Public Key instead to constructor from the factory.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
usmansaleem: This is very unlikely to happen error. I've updated the code to add a comment explaining as such. Since these are checked exceptions, I am wrapping them in a runtime exception. Additionally, I've now used Bouncycastle provider instead of using the default Sun implementation.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
usmansaleem: 👍🏼. Refactored the code to use `AwsKms...` naming convention.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKMSSigner.java
usmansaleem: yeah this one is tricky. I don't know how long we can keep around kmsClient. Since it is a "closeable" resource, it ought to go in try/resource block.

All of tests involve this class (and Aws Kms remote signing feature) are integration tests, talking to external service.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/util/Eth1SignatureUtil.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.web3j.crypto.ECDSASignature;
+import org.web3j.crypto.Sign;
+import org.web3j.utils.Numeric;
+
+public class Eth1SignatureUtil {
+  private static final Logger LOG = LogManager.getLogger();
+
+  public static Signature deriveSignatureFromDerEncoded(
```
jframe, line 31: Can you add some tests for this

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSignerFactory.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @VisibleForTesting
+  public static ECPublicKey getECPublicKey(final KmsClient kmsClient, final String kmsKeyId) {
+    // kmsClient can be null/closed if close method has been called.
+    checkArgument(kmsClient != null, "KmsClient is not initialized");
+
+    final GetPublicKeyRequest getPublicKeyRequest =
+        GetPublicKeyRequest.builder().keyId(kmsKeyId).build();
+    final GetPublicKeyResponse publicKeyResponse = kmsClient.getPublicKey(getPublicKeyRequest);
+    KeySpec keySpec = publicKeyResponse.keySpec();
```
jframe, line 68: nit: final

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSigner.java
```diff
@@ -0,0 +1,71 @@
...
+        SignRequest.builder()
+            .keyId(awsKmsMetadata.getKmsKeyId())
+            .signingAlgorithm(SigningAlgorithmSpec.ECDSA_SHA_256)
+            .messageType(MessageType.DIGEST)
+            .message(SdkBytes.fromByteArray(dataToSign))
+            .build();
+
+    final byte[] signature;
+    try (final KmsClient kmsClient =
+        AwsKmsClientFactory.createKmsClient(
```
jframe, line 57: Is this expensive to create a KmsClient each time we need sign? Looks like it would need to redo the authentication each time

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSigner.java
```diff
@@ -0,0 +1,71 @@
...
+    final byte[] dataToSign = applySha3Hash ? Hash.sha3(data) : data;
+    final SignRequest signRequest =
+        SignRequest.builder()
+            .keyId(awsKmsMetadata.getKmsKeyId())
+            .signingAlgorithm(SigningAlgorithmSpec.ECDSA_SHA_256)
+            .messageType(MessageType.DIGEST)
+            .message(SdkBytes.fromByteArray(dataToSign))
+            .build();
+
+    final byte[] signature;
```
jframe, line 55: Does this need to be defined outside of the try? we could just return the signature in the last line of the try I think

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/AwsKmsMetadataDeserializer.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+
+public class AwsKmsMetadataDeserializer extends StdDeserializer<AwsKmsMetadata> {
```
jframe, line 30: Can you add some tests for the deserializer

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -110,6 +124,68 @@ public void signDataWithKeyInAzure() {
...
+        named = "AWS_SECRET_ACCESS_KEY",
+        matches = ".*",
+        disabledReason = "AWS_SECRET_ACCESS_KEY env variable is required"),
+  })
+  public void remoteSignWithAwsKMS() {
+    final String roAwsAccessKeyId = System.getenv("AWS_ACCESS_KEY_ID");
+    final String roAwsSecretAccessKey = System.getenv("AWS_SECRET_ACCESS_KEY");
+    final Optional<String> awsSessionToken =
+        Optional.ofNullable(System.getenv("AWS_SESSION_TOKEN"));
+    final String region = Optional.ofNullable(System.getenv("AWS_REGION")).orElse("us-east-2");
```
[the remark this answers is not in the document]
usmansaleem: mainly due to our test data is created in `us-east-2` region.

acceptance-tests/src/test/java/tech/pegasys/web3signer/tests/signing/SecpSigningAcceptanceTest.java
```diff
@@ -110,6 +124,68 @@ public void signDataWithKeyInAzure() {
...
+    // can be pointed to localstack
+    final Optional<URI> awsEndpointOverride =
+        System.getenv("AWS_ENDPOINT_OVERRIDE") != null
+            ? Optional.of(URI.create(System.getenv("AWS_ENDPOINT_OVERRIDE")))
+            : Optional.empty();
```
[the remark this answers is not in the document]
usmansaleem: nice ....

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSignerTest.java
```diff
@@ -0,0 +1,174 @@
...
+    ScheduleKeyDeletionRequest deletionRequest =
+        ScheduleKeyDeletionRequest.builder().keyId(testKeyId).pendingWindowInDays(7).build();
+    awsKMSClient.scheduleKeyDeletion(deletionRequest);
+
+    // close
+    awsKMSClient.close();
+  }
+
+  @Test
+  public void awsKmsSignerCanSignTwice() {
```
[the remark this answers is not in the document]
usmansaleem: this was one carried over from Azure tests ... I think we should remove it.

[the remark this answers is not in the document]
usmansaleem: or verify that we signature is signed through the same pubic key.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSigner.java
usmansaleem: Currently createKmsClient will create a new client everytime it is called. We can possibly introduce a cache/map in the factory that can return cached objects for same credentials (might be useful in bulk loading as well). What I am not entirely sure about is can we keep same kmsClient instance open?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSigner.java
usmansaleem: updated.

signing/src/main/java/tech/pegasys/web3signer/signing/config/metadata/AwsKmsMetadataDeserializer.java
usmansaleem: 👍🏼

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/util/Eth1SignatureUtil.java
usmansaleem: This methods of this class are indirectly tested by Azure and AWS integration test that uses "sign" operation.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/CachedAwsKmsClientFactory.java
```diff
@@ -13,25 +13,34 @@
...
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.kms.KmsClient;
 import software.amazon.awssdk.services.kms.KmsClientBuilder;
 
-/** Factory class that provide instance of KmsClient */
-public class AwsKmsClientFactory {
+/** Factory class that provide cached instance of KmsClient */
+public class CachedAwsKmsClientFactory {
+  private static final Map<AwsKmsClientKey, AwsKmsClient> CACHE = new ConcurrentHashMap<>();
```
siladu, line 27: I guess it doesn't make sense to ever clear this cache - object should be tied to the app lifecycle anyway?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsClientKey.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+
+/** This class acts as a key to identify Aws KmsClient from the cache. */
+final class AwsKmsClientKey {
+  final AwsCredentials awsCredentials;
+  final String region;
+  final Optional<URI> endpointOverrid;
```
jframe, line 25:
```suggestion
  final Optional<URI> endpointOverride;
```

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsClient.java
```diff
@@ -0,0 +1,99 @@
...
+import software.amazon.awssdk.services.kms.model.MessageType;
+import software.amazon.awssdk.services.kms.model.ScheduleKeyDeletionRequest;
+import software.amazon.awssdk.services.kms.model.SignRequest;
+import software.amazon.awssdk.services.kms.model.SigningAlgorithmSpec;
+
+/**
+ * Wraps KmsClient to allow the same instance to be cached and re-used. It exposes the methods that
+ * our code use.
+ */
+public class AwsKmsClient implements Closeable {
```
jframe, line 42: Does this need to be closable? Can't see where we call the close method on this

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsSignerTest.java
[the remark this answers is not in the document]
jframe: Testing that this is signing twice isn't useful but we should have a test that shows that we sign correctly when we have hashing enabled as the other test is testing the hashing is false case.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/CachedAwsKmsClientFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.kms.KmsClient;
+import software.amazon.awssdk.services.kms.KmsClientBuilder;
+
+/** Factory class that provide cached instance of KmsClient */
+public class CachedAwsKmsClientFactory {
+  private static final Map<AwsKmsClientKey, AwsKmsClient> CACHE = new ConcurrentHashMap<>();
```
jframe, line 27: Don't really like this being unbounded. We could use the Guava cache so we have a max number of clients that will be cached in memory.

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/util/Eth1SignatureUtilTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    final KeyPair keyPair = generateEC_SECPKeyPair();
+    final ECKeyPair web3jECKeyPair = ECKeyPair.create(keyPair);
+    // sign using web3j (which uses BouncyCastle ECDSASigner)
+    final ECDSASignature web3jSig = web3jECKeyPair.sign(dataToSign);
+
+    // concatenate r || s to create byte[]
+    final Bytes rBytes = Bytes.of(web3jSig.r.toByteArray()).trimLeadingZeros();
+    final Bytes sBytes = Bytes.of(web3jSig.s.toByteArray()).trimLeadingZeros();
+    final byte[] p1363Signature = Bytes.concatenate(rBytes, sBytes).toArray();
```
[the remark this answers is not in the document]
usmansaleem: web3j and bouncycastle returns signature (R,S) as BigIntegers. Our method signature takes byte[] (which is how Azure would return the signature [R||S]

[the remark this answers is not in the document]
siladu: OK so there's slightly different "Azure view" of P1363 compared to "web3j+BC's view" of P1363?
I use "view" as not sure "encoding" is accurate for this difference!

jframe (approved):

siladu (approved):
LGTM

jframe (approved):
LGTM

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/aws/AwsKmsClient.java
```diff
@@ -37,9 +36,10 @@
 
 /**
  * Wraps KmsClient to allow the same instance to be cached and re-used. It exposes the methods that
- * our code use.
+ * our code use. Since AwsKmsClient is meant to live for the duration of web3signer life, we have
+ * not implemented close method.
```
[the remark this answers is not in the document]
usmansaleem: Hopefully not ... but I will perform a manual test just in case :).

PR #853, opened 2023-07-19
eth_sign using eth1Address map

gfukushima:
## PR Description
This PR fixes a bug introduced during the migration of eth_sign to web3signer.
The current state of the codebase has a map of <PublicKey, Signer>. When eth_sign is called an eth1Address is provided.
For performance reasons we've introduced a second map <eth1Address, Signer> that allows a O1 look up. This new map wasn't being used in the signing methods.
This PR also removes the direct usage of signerProvider where possible by replacing it with A SignForIdentifier instance created using the eth1Address mapped signerProvider.
At last this PR adds a eth_sign integration test . 
   

## Documentation

## Changelog

## Testing

### DIFF
--- a/core/src/integrationTest/java/tech/pegasys/web3signer/core/jsonrpcproxy/EthSignIntegrationTest.java
+++ b/core/src/integrationTest/java/tech/pegasys/web3signer/core/jsonrpcproxy/EthSignIntegrationTest.java
@@ -1,0 +1,56 @@
+package tech.pegasys.web3signer.core.jsonrpcproxy;
+
+import io.netty.handler.codec.http.HttpHeaderValues;
+import io.vertx.core.json.Json;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.junit.jupiter.api.Test;
+import org.web3j.crypto.Keys;
+import org.web3j.protocol.core.Request;
+import org.web3j.protocol.core.methods.response.EthSign;
+import tech.pegasys.web3signer.core.service.jsonrpc.response.JsonRpcErrorResponse;
+import tech.pegasys.web3signer.core.service.jsonrpc.response.JsonRpcSuccessResponse;
+
+import java.security.InvalidAlgorithmParameterException;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.util.Map;
+
+import static java.util.Collections.singletonList;
+import static org.web3j.crypto.Keys.getAddress;
+import static tech.pegasys.web3signer.core.service.jsonrpc.response.JsonRpcError.SIGNING_FROM_IS_NOT_AN_UNLOCKED_ACCOUNT;
+
+public class EthSignIntegrationTest extends IntegrationTestBase {
+
+
+    @Test
+    void ethSignSignsMessageWhenAnUnlockedAccountIsPassed(){
+        final Request<?, EthSign> requestBody = jsonRpc().ethSign(unlockedAccount,"some random data to be signed");
+
+        final Iterable<Map.Entry<String, String>> expectedHeaders =
+                singletonList(ImmutablePair.of("Content", HttpHeaderValues.APPLICATION_JSON.toString()));
+
+        final JsonRpcSuccessResponse responseBody =
+                new JsonRpcSuccessResponse(requestBody.getId(), "0x22cf811d54835be906e45d728dfc8a7a32bb65ce146a4ccb6b689b8d8182a30f19fdb947229991b9dfd63b67fe4ebb59a3d3f8d9ed95b27e0c3d34e1800afbd81b");
+
+        sendPostRequestAndVerifyResponse(
+                request.web3Signer(Json.encode(requestBody)),
+                response.web3Signer(expectedHeaders, Json.encode(responseBody)));
+    }
+
+    @Test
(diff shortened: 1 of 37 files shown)


core/src/integrationTest/java/tech/pegasys/web3signer/core/jsonrpcproxy/EthSignIntegrationTest.java
siladu, line 1: 👍

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SignerForIdentifier.java
```diff
@@ -81,4 +81,14 @@ private String formatSignature(final ArtifactSignature signature) {
       throw new IllegalStateException("Invalid signature type");
     }
   }
+
+  /**
+   * Checks whether a signer for the passed identifiers is present
```
siladu, line 86:
```suggestion
   * Checks whether a signer for the passed identifier is present
```

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/internalresponse/EthSignResultProvider.java
```diff
@@ -50,13 +50,13 @@ public String createResponseResult(final JsonRpcRequest request) {
       throw new JsonRpcException(INVALID_PARAMS);
     }
 
-    final String identifier = params.get(0);
+    final String eth1Address = params.get(0);
```
siladu, line 53: 👍

signing/src/main/java/tech/pegasys/web3signer/signing/config/SecpArtifactSignerProviderAdapter.java
```diff
@@ -30,6 +30,10 @@
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+/**
+ * Adapts the provided ArtifactSignerProvider into a map of signers using their eth1 address as the
+ * identifier rather than the public key which is the default behaviour
```
siladu, line 35: 👍

siladu (approved):
LGTM

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
```diff
@@ -73,11 +76,12 @@ private static byte[] prependEip1559TransactionType(byte[] bytesToSign) {
         .array();
   }
 
-  private SignatureData sign(final byte[] bytesToSign) {
-    final SecpArtifactSignature artifactSignature =
```
jframe, line 77: Nice getting rid of that cast to SecpArtifactSignature

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
```diff
@@ -73,11 +76,12 @@ private static byte[] prependEip1559TransactionType(byte[] bytesToSign) {
...
-        (SecpArtifactSignature) signer.sign(Bytes.of(bytesToSign));
+  private SignatureData sign(final String eth1Address, final byte[] bytesToSign) {
+    final String bytesSigned =
+        secpSigner.sign(normaliseIdentifier(eth1Address), Bytes.of(bytesToSign)).orElseThrow();
```
jframe, line 81: We should throw JsonRpcException(SIGNING_FROM_IS_NOT_AN_UNLOCKED_ACCOUNT) so that the appropriate error is returned for the JsonRpc response

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
```diff
@@ -73,11 +76,12 @@ private static byte[] prependEip1559TransactionType(byte[] bytesToSign) {
...
 
-    final Signature signature = artifactSignature.getSignatureData();
+    final Signature signature =
```
jframe, line 83: nit: not ideal going from Bytes -> String -> byte[] -> SecpArtifactSignature. Do we need a sign method on SignerForIdentifier that will return Bytes instead?

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
gfukushima: Good catch

jframe, line 81: Adding a sign method that returns an ArtifactSignature would be a nicer way of doing this rather than converting to a hex string and back to Bytes. We already have the SignerIdentifier using generics so it could return a SecpArtifactSignature we could use directly.

gfukushima: done

jframe (approved):

Issue #718, opened 2023-02-20
Use Java 17 as default to compile and run Web3Signer code


usmansaleem:
Use Java 17 as default to compile and run Web3Signer code. Web3Signer uses Besu's Metrics service as a dependency which is now compiled in Java 17 and cannot be used with Web3Signer when later is compiled using Java 11.

Tasks: 

- Update gradle build file and update warning that Java 17 and upward can only be used for compilation
- Remove jdk11 docker image and distDocker logic in gradle build file
- consider updating signers library to build for Java 17 too (may not want to do this until Teku is Java 17 since they rely on it and may need security patches).
- Announce in Discord and Slack plans to Java 17 for next release.

PR #862, opened 2023-07-27
Sign using http api

### LINKED ISSUE #753: Fix eth1 mode Azure remote signing (Java 17)
eth1 Azure mode remote signing is not working in Java 17 due to a bug in Azure libraries [1]. 

The fix would need to go both in "signers" library and then bring in Web3Signer.

https://github.com/Azure/azure-sdk-for-java/issues/26689

gfukushima:
## PR Description
This PR replaces the sign method from the azure library for a http call to the the Azure REST API.
This unblocks #718.

## Fixed Issue(s)

Fixes #753

## Documentation

## Changelog

## Testing

### DIFF
--- a/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnection.java
+++ b/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnection.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2023 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.web3signer.keystorage.azure;
+
+import java.io.IOException;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.Map;
+
+public class AzureConnection {
+  private final HttpClient httpClient;
+
+  public AzureConnection(final HttpClient httpClient) {
+    this.httpClient = httpClient;
+  }
+
+  public Map<String, String> executeHttpRequest(final HttpRequest httpRequest) {
+
+    final HttpResponse<String> response;
+    try {
+      response = httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+    } catch (final IOException | InterruptedException | RuntimeException e) {
+      throw new RuntimeException("Error communicating with Azure vault: " + e.getMessage(), e);
+    }
+
+    if (response.statusCode() != 200 && response.statusCode() != 204) {
+      throw new RuntimeException(
+          String.format(
+              "Error communicating with Azure vault: Received invalid Http status code %d.",
+              response.statusCode()));
+    }
+
+    return AzureKVResponseMapper.from(response.body());
+  }
+}
--- a/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
+++ b/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
@@ -18,14 +18,11 @@ import java.net.URI;
 import java.net.http.HttpClient;
 import java.util.Optional;
 
-// TODO revisit this to cleanup unnecessary params for azure
-// Copied from hashicorp
 public class AzureConnectionParameters {
   private static final Long DEFAULT_TIMEOUT_MILLISECONDS = 10_000L;
   private static final Integer DEFAULT_SERVER_PORT = 8200;
   private final long timeoutMs;
   private final HttpClient.Version httpProtocolVersion;
-
   private final URI vaultURI;
 
   public static Builder newBuilder() {
@@ -41,8 +38,7 @@ public class AzureConnectionParameters {
     this.timeoutMs = timeoutMs.orElse(DEFAULT_TIMEOUT_MILLISECONDS);
     this.httpProtocolVersion = httpProtocolVersion.orElse(HttpClient.Version.HTTP_2);
     this.vaultURI =
-        URI.create(
-            String.format("%s:%d", serverHost, serverPort.orElse(DEFAULT_SERVER_PORT)));
+        URI.create(String.format("%s:%d", serverHost, serverPort.orElse(DEFAULT_SERVER_PORT)));
   }
 
   public long getTimeoutMilliseconds() {
--- a/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
+++ b/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
@@ -19,7 +19,7 @@ import java.net.http.HttpClient;
 import java.util.Optional;
 
 // TODO revisit this to cleanup unnecessary params for azure
-// Copied from hashicorp directory
+// Copied from hashicorp
 public class AzureConnectionParameters {
   private static final Long DEFAULT_TIMEOUT_MILLISECONDS = 10_000L;
   private static final Integer DEFAULT_SERVER_PORT = 8200;
--- a/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
+++ b/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2023 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.web3signer.keystorage.azure;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.util.Optional;
+
+// TODO revisit this to cleanup unnecessary params for azure
+// Copied from hashicorp directory
+public class AzureConnectionParameters {
+  private static final Long DEFAULT_TIMEOUT_MILLISECONDS = 10_000L;
+  private static final Integer DEFAULT_SERVER_PORT = 8200;
+  private final long timeoutMs;
+  private final HttpClient.Version httpProtocolVersion;
+
+  private final URI vaultURI;
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  /* Optional parameters will be set to their defaults when connecting */
+  private AzureConnectionParameters(
+      final String serverHost,
+      final Optional<Integer> serverPort,
+      final Optional<Long> timeoutMs,
+      final Optional<HttpClient.Version> httpProtocolVersion) {
+    this.timeoutMs = timeoutMs.orElse(DEFAULT_TIMEOUT_MILLISECONDS);
+    this.httpProtocolVersion = httpProtocolVersion.orElse(HttpClient.Version.HTTP_2);
+    this.vaultURI =
+        URI.create(
+            String.format("https://%s:%d", serverHost, serverPort.orElse(DEFAULT_SERVER_PORT)));
+  }
+
+  public long getTimeoutMilliseconds() {
+    return timeoutMs;
+  }
+
+  public URI getVaultURI() {
+    return vaultURI;
+  }
+
+  public HttpClient.Version getHttpProtocolVersion() {
+    return httpProtocolVersion;
+  }
+
+  public static final class Builder {
+    private String serverHost;
+    private Optional<Integer> serverPort = Optional.empty();
+    private Optional<Long> timeoutMs = Optional.empty();
+    private Optional<HttpClient.Version> httpProtocolVersion = Optional.empty();
+
+    Builder() {}
+
+    public Builder withServerHost(final String serverHost) {
+      this.serverHost = serverHost;
+      return this;
+    }
+
+    public Builder withServerPort(final Integer serverPort) {
+      this.serverPort = Optional.ofNullable(serverPort);
+      return this;
+    }
+
+    public Builder withTimeoutMs(final Long timeoutMs) {
+      this.timeoutMs = Optional.ofNullable(timeoutMs);
+      return this;
+    }
+
+    public Builder withHttpProtocolVersion(final HttpClient.Version httpProtocolVersion) {
+      this.httpProtocolVersion = Optional.ofNullable(httpProtocolVersion);
+      return this;
+    }
+
+    public AzureConnectionParameters build() {
+      checkNotNull(serverHost, "Azure host cannot be null");
+      return new AzureConnectionParameters(serverHost, serverPort, timeoutMs, httpProtocolVersion);
+    }
+  }
+}
--- a/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
+++ b/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
@@ -42,7 +42,7 @@ public class AzureConnectionParameters {
     this.httpProtocolVersion = httpProtocolVersion.orElse(HttpClient.Version.HTTP_2);
     this.vaultURI =
         URI.create(
-            String.format("https://%s:%d", serverHost, serverPort.orElse(DEFAULT_SERVER_PORT)));
+            String.format("%s:%d", serverHost, serverPort.orElse(DEFAULT_SERVER_PORT)));
   }
 
   public long getTimeoutMilliseconds() {
--- a/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKVResponseMapper.java
+++ b/keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKVResponseMapper.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2023 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.web3signer.keystorage.azure;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import io.vertx.core.json.DecodeException;
+import io.vertx.core.json.JsonObject;
+
+public class AzureKVResponseMapper {
+  public static final String ERROR_INVALID_JSON = "Invalid response returned from Azure Vault";
+
+  /**
+   * Convert Azure Sign HTTP JSON response to map of key/values
+   *
+   * @param json response from Azure REST API
+   * @return All key/value pairs
+   */
+  public static Map<String, String> from(final String json) {
+    if (json == null) {
+      throw new RuntimeException(ERROR_INVALID_JSON);
+    }
+    final JsonObject jsonResponse;
+    try {
+      jsonResponse = new JsonObject(json);
+    } catch (final DecodeException de) {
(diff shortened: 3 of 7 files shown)


keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnection.java
```diff
@@ -0,0 +1,46 @@
...
+    this.httpClient = httpClient;
+  }
+
+  public Map<String, String> executeHttpRequest(final HttpRequest httpRequest) {
+
+    final HttpResponse<String> response;
+    try {
+      response = httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+    } catch (final IOException | InterruptedException | RuntimeException e) {
+      throw new RuntimeException("Error communicating with Azure vault: " + e.getMessage(), e);
```
siladu, line 34: Will this be retryable/recoverable or crash the thread?

I'm guessing this is only executed during an external sign request sent to web3signer and so the user can always resend...will they get a 500 error in the response?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import tech.pegasys.web3signer.keystorage.azure.AzureConnectionParameters;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AzureConnectionFactory {
+  private final Map<URI, HttpClient> httpClientMap = new ConcurrentHashMap<>();
```
siladu, line 25: Does the entries here need to be cleaned up at all or is it OK to leave them there for the lifetime of the app?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
```diff
@@ -58,13 +66,14 @@ public class AzureKeyVaultSigner implements Signer {
...
-
-  @Override
-  public Signature sign(byte[] data) {
-    final AzureKeyVault vault;
+    final AzureConnectionFactory connFactory = new AzureConnectionFactory();
+    final AzureConnectionParameters connectionParameters =
+        AzureConnectionParameters.newBuilder()
+            .withServerHost(constructAzureKeyVaultUrl(config.getKeyVaultName()))
+            .build();
+    this.azureConnection = connFactory.getOrCreateConnection(connectionParameters);
```
siladu, line 74: Do we end up with an AzureConnection object per Signer for the entire lifetime of the app? What happens when there's 10,000 signers?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
```diff
@@ -74,12 +83,21 @@ public Signature sign(byte[] data) {
...
+  @Override
+  public Signature sign(byte[] data) {
 
     final CryptographyClient cryptoClient =
         vault.fetchKey(config.getKeyName(), config.getKeyVersion());
 
     final byte[] dataToSign = needsToHash ? Hash.sha3(data) : data;
-    final SignResult result = cryptoClient.sign(signingAlgo, dataToSign);
+
+    // 2023-07-27 - We can use the sign method from the azure
```
siladu, line 96: nit: maybe adding TODO will make this slightly easier to spot in the future?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
```diff
@@ -98,6 +108,41 @@ public CryptographyClient fetchKey(final String keyName, final String keyVersion
...
+      final String vaultName) {
+    final String keyName = cryptoClient.getKey().getName();
+    final String keyVersion = cryptoClient.getKey().getProperties().getVersion();
+    final String apiVersion = KeyServiceVersion.getLatest().getVersion();
+
+    final JsonObject jsonBody = new JsonObject();
+    jsonBody.put("alg", signingAlgo);
+    jsonBody.put("value", Bytes.of(data).toBase64String());
+
+    final String uriString = constructAzureSignApitUri(vaultName, keyName, keyVersion, apiVersion);
```
siladu, line 124: typo? ApitUrl -> ApiUrl?

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKVResponseMapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+    return Collections.unmodifiableMap(
+        jsonResponse.stream()
+            .filter(entry -> Objects.nonNull(entry.getValue()))
+            .collect(Collectors.toMap(Map.Entry::getKey, AzureKVResponseMapper::getValueToString)));
+  }
+
+  private static String getValueToString(final Map.Entry<String, Object> v) {
+    return v.getValue().toString();
+  }
```
siladu, line 51: think this could be simplified by using jsonResponse.getMap() and bubbling up the Map<String, Object> type.

Then just call toString on the values as you use them. This also saves having to convert all the fields to strings when we're only actually using two of them.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
```diff
@@ -23,42 +26,48 @@
...
+import org.apache.tuweni.bytes.Bytes;
 
 public class AzureKeyVault {
 
   private static final Logger LOG = LogManager.getLogger();
-
   private final TokenCredential tokenCredential;
   private final SecretClient secretClient;
   private final KeyClient keyClient;
+  private static final List<String> SCOPE = List.of("https://vault.azure.net/.default");
```
siladu, line 56: This looks odd to me, is it an azure vault convention? Do we need to make it configurable?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
```diff
@@ -58,13 +66,14 @@ public class AzureKeyVaultSigner implements Signer {
...
-    final AzureKeyVault vault;
+    final AzureConnectionFactory connFactory = new AzureConnectionFactory();
+    final AzureConnectionParameters connectionParameters =
+        AzureConnectionParameters.newBuilder()
+            .withServerHost(constructAzureKeyVaultUrl(config.getKeyVaultName()))
+            .build();
+    this.azureConnection = connFactory.getOrCreateConnection(connectionParameters);
     try {
-      vault =
+      this.vault =
```
siladu, line 76: I would consider connecting to the vault outside of the constructor. If it fails can it result in the object being left in an inconsistent state?

siladu (commented):
Seems pretty straight forward. Mostly just questions around failure states.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
gfukushima: Fixed

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
gfukushima: This is the recommended by Azure. Obviously this depends on what service you are trying to access. For instance in their example they use `https://graph.microsoft.com/.default` for the graph service. In our case think it's safe to hardcode that to vault since it's the only service our application will use the REST API for.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
gfukushima: The main reason of this change is that we don't want to keep a Signer in memory that failed to create the Vault object basically because they won't have any use afaik. Also we want to make sure that this is created only once which agree that we could do without having this in the constructor.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
gfukushima: If we kill the connection we lose the performance gain of having that cached. Which basically drops the signing time significantly overall.

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnection.java
gfukushima: They will get whatever is return by the azure API. When I was testing this with the worng token for instance I received a few 401 and 403.

siladu: But in this case, we are failing to communicate with Azure, so presumably won't be able to return Azure's specific response?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
siladu: What determines the URI key, is it effectively a single entry map (one azure url) or is there one per key, etc?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
siladu: think it could either go in an init() method called after constructor, or in a factory before construction

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
```diff
@@ -0,0 +1,89 @@
...
+      this.serverPort = Optional.ofNullable(serverPort);
+      return this;
+    }
+
+    public Builder withTimeoutMs(final Long timeoutMs) {
+      this.timeoutMs = Optional.ofNullable(timeoutMs);
+      return this;
+    }
+
+    public Builder withHttpProtocolVersion(final HttpClient.Version httpProtocolVersion) {
```
jframe, line 79: Doesn't look like these are ever set. Should these be configurable somewhere?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
jframe, line 25: Don't like this being unbounded. Although in practice we are unlikely to have 1000 different vaultUrls if we did don't think web3signer would handle that well.

Perhaps the answer is to set the max size on the Cache. Could use the Guava cache to do that.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
siladu: I was thinking a connection pool might be appropriate but not sure what perf tests you've already done, so might be unnecessary

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
gfukushima: one per vault

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnectionParameters.java
gfukushima: Removed the unused ones

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
gfukushima: Using caffeine (Guava successor) now to manage that.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
gfukushima: changed this. should be clearer now

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnection.java
gfukushima: Happy to remove that if you think "Azure vault" in the message is misleading?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
gfukushima: Changed it to use the connection pool that we're passing to that class now

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureConnection.java
siladu: I have no issue with the message, I'm just trying to understand what the UX is when this happens

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
siladu: I think this is the same comment/resolution: https://github.com/Consensys/web3signer/pull/862#discussion_r1277151220

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
```diff
@@ -44,46 +44,39 @@ public class AzureKeyVaultSigner implements Signer {
...
+  private final AzureKeyVault vault;
 
   AzureKeyVaultSigner(
       final AzureConfig config,
       final Bytes publicKey,
       final boolean needsToHash,
       final boolean useDeprecatedSignatureAlgorithm,
-      final AzureKeyVaultFactory azureKeyVaultFactory) {
+      final AzureKeyVault azureKeyVault,
+      final AzureConnection azureConnection) {
```
jframe, line 57: Will having this as a strong reference, prevent the cache from disposing of the underlying HttpClient?

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactoryTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    assertThat(azureConnFactory.getConnectionPool().asMap().size()).isEqualTo(1);
+
+    azureConnFactory.getOrCreateConnection(azureConnParams);
+
+    assertThat(azureConnFactory.getConnectionPool().asMap().size()).isEqualTo(1);
+
+    azureConnFactory.getOrCreateConnection(azureConnParams2);
+
+    assertThat(azureConnFactory.getConnectionPool().asMap().size()).isEqualTo(2);
```
jframe, line 48: Can we also assert that the same connection is returned for the same connection params. We are only checking the size in this test

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
```diff
@@ -103,6 +115,43 @@ public CryptographyClient fetchKey(final String keyName, final String keyVersion
...
+
+    final JsonObject jsonBody = new JsonObject();
+    jsonBody.put("alg", signingAlgo);
+    jsonBody.put("value", Bytes.of(data).toBase64String());
+
+    final String uriString =
+        constructAzureSignApiUri(vaultName, azureKeyName, azureKeyVersion, apiVersion);
+    final URI uri = URI.create(uriString);
+
+    final HttpRequest httpRequest =
```
jframe, line 135: nit: local variable isn't necessary we can just return the request

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
```diff
@@ -103,6 +115,43 @@ public CryptographyClient fetchKey(final String keyName, final String keyVersion
         .buildClient();
   }
 
+  public HttpRequest getRemoteSigningHttpRequest(
```
jframe, line 118: Why are we returning a HttpRequest rather having a method to just sign the request. The Azure logic is split between the `AzureKeyVault` and the `AzureVaultSigner`. I see that there was specific Azure code already in the signer but just trying understand the split/scope for the `AzureKeyVault`

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactoryTest.java
gfukushima: Done

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureKeyVaultSigner.java
gfukushima: Changed that along with some renames and some clean ups in the connection factory

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
gfukushima: done

keystorage/src/main/java/tech/pegasys/web3signer/keystorage/azure/AzureKeyVault.java
gfukushima: This is some tech debt that we could probably clean up once we tide up the AzureKeyVault class but for now I can't find a way to do it without exposing the tokenCredential which I think is a worse solution than just asking the AzureKeyVault to assemble the signing http request.

usmansaleem (commented):
LGTM

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureHttpClientFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+            .withServerHost(constructAzureKeyVaultUrl("Vault2"))
+            .build();
+    final AzureHttpClient conn1 = azureConnFactory.getOrCreateHttpClient(azureConnParams);
+    // assert a new client has been created
+    assertThat(azureConnFactory.getConnectionPool().asMap().size()).isEqualTo(1);
+
+    final AzureHttpClient conn2 = azureConnFactory.getOrCreateHttpClient(azureConnParams);
+
+    // assert a client1 and 2 are the same
+    assertThat(conn1).isEqualTo(conn2);
```
jframe, line 46: nit: using isSameAs would be better than isEqualTo so know that it is same instance as well

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureHttpClientFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.time.Duration;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import org.jetbrains.annotations.VisibleForTesting;
+
+public class AzureHttpClientFactory {
+  private final int CONNECTION_POOL_SIZE_LIMIT = 10;
```
siladu, line 27: would prefer something like CLIENT_CACHE_SIZE as don't think this is a connection pool in the traditional sense.

Also, maybe we want to make this configurable, but doesn't need to be in this PR.

jframe (approved):

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureHttpClientFactory.java
```diff
@@ -0,0 +1,53 @@
...
+          try {
+            return new AzureHttpClient(httpClientBuilder.build());
+          } catch (final Exception e) {
+            throw new RuntimeException("Unable to initialise connection to azure vault.", e);
+          }
+        });
+  }
+
+  @VisibleForTesting
+  protected Cache<URI, AzureHttpClient> getConnectionPool() {
```
siladu, line 50: same point about the name

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureConnectionFactory.java
```diff
@@ -12,17 +12,21 @@
...
-import java.util.concurrent.ConcurrentHashMap;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 
 public class AzureConnectionFactory {
-  private final Map<URI, HttpClient> httpClientMap = new ConcurrentHashMap<>();
+    private final int CONNECTION_POOL_SIZE_LIMIT = 10;
+  private final Cache<URI, HttpClient> httpClientMap =
+      Caffeine.newBuilder().maximumSize(CONNECTION_POOL_SIZE_LIMIT).build();
```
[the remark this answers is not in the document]
gfukushima: Good catch, I renamed a lot of classes to reflect better what we they are and forgot that constant

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/azure/AzureHttpClientFactoryTest.java
gfukushima: done

siladu (approved):
Just a naming gripe. Shouldn't block this PR.

PR #901, opened 2023-09-06
Docker tag prefixes for RC versions

usmansaleem:
## PR Description
Create `latest` and `x.y` docker tag prefixes only for non-develop and non-rc releases. This will allow to create releases such as 23.9.0-RC1 which doesn't update the latest and 23.9 docker tags.

## Documentation

## Changelog

## Testing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ plugins {
   id 'io.spring.dependency-management' version '1.0.11.RELEASE'
   id 'me.champeau.gradle.jmh' version '0.5.3' apply false
   id 'net.ltgt.errorprone' version '2.0.2'
-  id 'org.ajoberstar.grgit' version '4.1.1'
+  id 'org.ajoberstar.grgit' version '5.2.0'
 }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -506,22 +505,30 @@ task testDocker {
...
   def versionPrefixes = [dockerBuildVersion]
   if (project.hasProperty('branch') && project.property('branch') == 'master') {
     versionPrefixes.add('develop')
   }
 
-  if (!isDevelopBuild) {
+  // add `latest` tag for non-develop and non-rc versions
+  if (!rootProject.version.contains('develop') && !rootProject.version.toLowerCase().contains('-rc')) {
     versionPrefixes.add('latest')
+    // when docker build version is x.y.[a,b..] we would also like to use tag x.y
```
jframe, line 517: Guess this doesn't hurt, but probably don't need this either

jframe (approved):
LGTM

build.gradle
usmansaleem: It was already part of existing logic, I simply added the comment for readability.

PR #904, opened 2023-09-06
Non blocking upcheck

### LINKED ISSUE #891: Upcheck sometimes blocked by Pruning
...resulting in the service getting killed by k8s.

Reported by Yorick on Consensys Discord https://discord.com/channels/697535391594446898/725135157652226060/1145966681236193371

> It gets killed ... by k8s? 
> Normal   Killing    2m23s                  kubelet  Container web3signer failed liveness probe, will be restarted
So there's a liveness probe running, and when web3signer prunes, that's ... blocked? Prune can take over 2 mins

> 2 min 6 secs for manual prune. OK for now I will change web3signer: Prune every 6 hours, keep 100, prune on start. That should keep the pruning runs fast enough that the liveness probe block isn't an issue

```
       livenessProbe:
            httpGet:
              path: /upcheck
              port: http
            initialDelaySeconds: 900
            timeoutSeconds: 3
            periodSeconds: 30
            failureThreshold: 3
            successThreshold: 1
```
> So that sounds like after 90 secs of no /upcheck it gets killed. Prune takes longer than that and then it can never finish.
Which suggests /upcheck is blocked during prune, and it's usually so fast that's not seen - but Gnosis Chain with 10k validators and a restart of web3signer which restarts the 24 hour clock and well 😅
Prune on start and every 6 hours should help with this, even if /upcheck remains blocked
Particularly with that initialDelaySeconds, if it prunes on start it should always be able to finish, and then the runs after that will be faster

> 11 minutes ... maybe there are bits where it blocks and bits where it doesn't?
```
2023-08-29 07:21:40.445+00:00 | pool-3-thread-1 | INFO  | DBSlashingProtectionPruner | Pruning slashing protection database for 9879 validators
2023-08-29 07:32:54.310+00:00 | pool-3-thread-1 | INFO  | DBSlashingProtectionPruner | Pruning slashing protection database complete
```

> This is Gnosis2, much faster there
```
2023-08-29 07:39:26.056+00:00 | pool-3-thread-1 | INFO  | DBSlashingProtectionPruner | Pruning slashing protection database for 4971 validators
2023-08-29 07:40:23.991+00:00 | pool-3-thread-1 | INFO  | DBSlashingProtectionPruner | Pruning slashing protection database complete
```

gfukushima:
## PR Description
This PR removes the usage of the BlockingHandlerDecorator from the upcheck endpoint. This some very particular scenarios can cause the upcheck to stop responding when the worker pool is exhausted.
 
## Fixed Issue(s)

Fixes #891

## Documentation

## Changelog

## Testing


jframe (approved):
LGTM

Issue #696, opened 2023-01-04
High-watermark feature


beetrootkid:
You guys have recently implemented the watermark-repair  subcommand which will specify that bellow a certain epoch, web3signer will refuse to sign. Wanted to ask, would it be simple to implement a similar feature, but instead web3signer would **not** sign beyond a certain epoch?

Context: we are running web3signer + vault + slashing db in an Azure cluster (lets call it `old cluster`). And now we'd like to move it to a `new cluster` in AWS. In theory, we could achieve near-zero downtime and safely if:

- old cluster  was running with --do-not-sign-after-epoch=160000
- new cluster  was running with --do-not-sign-before-epoch=160000  (not sure how to use the watermark-repair  but I assume it accomplishes the same as this illustrative flag)

post epoch 160000, we would naturally teardown old cluster and keep only new cluster. The end goal for us is to move keys from one vault to another. That means we'd have to load same set of validator public keys in 2 different vaults and 1) not be slashed 2) minimal downtime 3) simple process without data migrations. We would expect to perform this scenario only when strictly necessary. If there a better way to achieve the same result, let us know. thanks!

non-fungible-nelson:
@jframe let's look to prioritize this in the next few sprints.

beetrootkid:
Hi. Just following up on the status of this feature?

siladu:
@beetrootkid We are approaching a solution and would like some feedback please...

For safety reasons - specifically in order to protect against two web3signer instances that share the same database from being misconfigured - we prefer the use of the database for storing the high-watermark.

In order to protect against clock discrepancies in CL clients that still are running in the old cluster, we must leave the high-watermark in place until all instances using the high-watermark are decommissioned.

Therefore the current preference is to use an 'offline' command similar to the current watermark-repair, for example:
`web3signer eth2 set-high-watermark --epoch=<same source and target epoch> --slot=<slot>`

This would only need to be executed once on the old cluster's database.
The existing `web3signer eth2 watermark-repair --epoch=<same source and target epoch> --slot=<slot>` would need to be executed on the new cluster's database to set the low-watermark.

This is captured in the following ACs...

## Acceptance Criteria

1. Web3signer won't sign anything **at or beyond** the high-watermark (Web3signer already won't sign anything **below** the low-watermark)
1. Web3signer operator is responsible for setting a good high-watermark that is in the future (so good documentation is needed)
[...]

PR #908, opened 2023-09-11
GET /highWatermark endpoint

siladu:
Part of #696 

`GET http://127.0.0.1:9000/api/v1/eth2/highWatermark`
returns either:
```
{
   "epoch": 1,
   "slot": 2
}
```
or empty json object when no high watermark set.

Allow epoch or slot to be null and respond with null string for that value (manual updates to DB can lead to this, even if high-watermark subcommand prevents it).

No automated tests for the endpoint in this PR as it will be tested as part of the overall subcommand acceptance test

## Documentation

## Changelog

## Testing

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
@@ -25,6 +25,7 @@ import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.web3signer.core.config.BaseConfig;
 import tech.pegasys.web3signer.core.metrics.SlashingProtectionMetrics;
 import tech.pegasys.web3signer.core.service.http.SigningObjectMapperFactory;
+import tech.pegasys.web3signer.core.service.http.handlers.HighWatermarkHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.LogErrorHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.keymanager.delete.DeleteKeystoresHandler;
 import tech.pegasys.web3signer.core.service.http.handlers.keymanager.imports.ImportKeystoresHandler;
@@ -90,6 +91,7 @@ public class Eth2Runner extends Runner {
   public static final String KEYSTORES_PATH = "/eth/v1/keystores";
   public static final String PUBLIC_KEYS_PATH = "/api/v1/eth2/publicKeys";
   public static final String SIGN_PATH = "/api/v1/eth2/sign/:identifier";
+  public static final String HIGH_WATERMARK_PATH = "/api/v1/eth2/highWatermark";
   private static final Logger LOG = LogManager.getLogger();
 
   private final Optional<SlashingProtectionContext> slashingProtectionContext;
@@ -170,6 +172,13 @@ public class Eth2Runner extends Runner {
         .failureHandler(errorHandler);
 
     addReloadHandler(router, List.of(blsSignerProvider), errorHandler);
+
+    slashingProtectionContext.ifPresent(
+        protectionContext ->
+            router
+                .route(HttpMethod.GET, HIGH_WATERMARK_PATH)
+                .handler(new HighWatermarkHandler(protectionContext.getSlashingProtection()))
+                .failureHandler(errorHandler));
 
     if (isKeyManagerApiEnabled) {
       router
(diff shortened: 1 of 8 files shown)


core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
```diff
@@ -90,6 +91,7 @@ public class Eth2Runner extends Runner {
   public static final String KEYSTORES_PATH = "/eth/v1/keystores";
   public static final String PUBLIC_KEYS_PATH = "/api/v1/eth2/publicKeys";
   public static final String SIGN_PATH = "/api/v1/eth2/sign/:identifier";
+  public static final String HIGH_WATERMARK_PATH = "/api/v1/eth2/highWatermark";
```
siladu, line 94: Should this be under this API prefix considering it should not be part of https://ethereum.github.io/remote-signing-api ?

core/src/main/resources/openapi-specs/eth2/signing/paths/high_watermark.yaml
```diff
@@ -0,0 +1,24 @@
+get:
+  tags:
+    - 'High Watermark'
+  summary: 'The High Watermark epoch and slot applicable to all validators'
+  description: 'Returns the uint64 epoch and slot of the high watermark. Signing of attestations or blocks are prevented at or beyond this high watermark'
```
gfukushima, line 5:
```suggestion
  description: 'Returns the uint64 epoch and slot of the high watermark. Signing of attestations or blocks are prevented at and beyond this high watermark'
```

jframe, line 5: Think we should make it clear that if there is no high watermark then an empty result is returned

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/HighWatermarkHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class HighWatermarkHandler implements Handler<RoutingContext> {
+  private final SlashingProtection slashingProtection;
+
+  public HighWatermarkHandler(final SlashingProtection slashingProtection) {
+    this.slashingProtection = slashingProtection;
+  }
+
+  @Override
+  public void handle(final RoutingContext context) {
```
jframe, line 34: Is it worth a test to check that the handler returns the correct response?

siladu: As mentioned in PR description, planning on testing this via the WatermarkRepair subcommand AT...do you think this is have a separate test instead?

core/src/main/resources/openapi-specs/eth2/signing/paths/high_watermark.yaml
siladu: I think "or" is more correct here because current object having its signature prevents will either by at or beyond the high watermark. Also used this terminology in test names

siladu: done

gfukushima: no worries, I think it's worth clarifying in which situation each could happen. When I first read this I thought this was a mistake.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/HighWatermarkHandler.java
jframe: think it's fine to leave to be done as as part of the AT

jframe (approved):

core/src/main/resources/openapi-specs/eth2/signing/paths/high_watermark.yaml
siladu: Could reverse it: "Signing of attestations or blocks are only allow when they are lower than this high watermark"?

core/src/main/java/tech/pegasys/web3signer/core/Eth2Runner.java
jframe: Think it is fine to have it under this API prefix

core/src/main/resources/openapi-specs/eth2/signing/paths/high_watermark.yaml
gfukushima: much clearer

PR #920, opened 2023-09-29
Make vertx worker pool size configurable

jframe:
## PR Description
Make vertx worker pool size configurable. Web3Signer uses the worker pool for signing requests. Increasing the size of the pool might help if there is a spike in signing requests.

It can be configured by using the hidden option --Xworker-pool-size.

Default is set to the same the default in Vertx which is 20. See https://vertx.io/docs/apidocs/io/vertx/core/VertxOptions.html#DEFAULT_WORKER_POOL_SIZE

## Documentation

## Changelog

## Testing


siladu (approved):
LGTM. Do we need code to make the config and env vars work or is that just for the ATs?

jframe:
> LGTM. Do we need code to make the config and env vars work or is that just for the ATs?

That's just for ATs. We automatically get env and config through pico environment providers.

Issue #926, opened 2023-10-11
Build fails with Java 19 on fresh machine


usmansaleem:
The web3signer build fails with Java 19 on fresh machine with following error:
```
A problem occurred configuring root project 'web3signer'.
> java.util.concurrent.ExecutionException: org.gradle.api.GradleException: Failed to create Jar file /Users/**/.gradle/caches/jars-9/b868ab677410c01cf9b5a0e29a035e73/jackson-core-2.15.2.jar.
```
The cause of this error is due to gradle bug as reported here: https://github.com/FasterXML/jackson-core/issues/955

The fix is to upgrade gradle version to latest 7.6

PR #927, opened 2023-10-11
Fix web3signer build

usmansaleem:
## PR Description
- Upgrade gradle wrapper to 7.6 to allow build using Java 19 without existing `~/.gradle/caches` pre-populated. 
- Temporary suppression of resurfaced `CVE-2023-4586` as no current fix is available and our code is not directly using the netty handlers.

## Fixed Issue(s)

#926 

## Documentation

## Changelog

## Testing


gradle/owasp-suppression.xml
```diff
@@ -1,15 +1,15 @@
...
+    <suppress until="2023-12-12">
         <notes><![CDATA[
         Temporary suppression, as it's arguably a false positive: https://github.com/netty/netty/issues/8537#issuecomment-1527896917
         The correct thing to do is ensure hostname verification is enabled: https://codetinkering.com/enable-hostname-verification-netty/
         This is indeed the case for Azure, which is the netty-handler: https://github.com/Azure/azure-sdk-for-java/issues/18286#issuecomment-947958978
+        Also see: https://github.com/jeremylong/DependencyCheck/issues/5912
         The other vulnerable lib is besu-metrics, which might want looking at, will see if this CVE gets flagged in besu first
         file name: netty-handler-4.1.97.Final.jar
         ]]></notes>
-        <packageUrl regex="true">^pkg:maven/io\.netty/netty\-handler@.*$</packageUrl>
```
siladu, line 12: What was the reason for loosening the package filter? I'd be concerned that this CVE might crop up in other packages even though we've only considered netty-handler

usmansaleem: updated with more generic netty expression.

siladu (approved):

PR #929, opened 2023-10-16
Upgrade Azure libraries to fix CVE-2023-36415

usmansaleem:
## PR Description
Upgrade Azure libraries to fix CVE-2023-36415. Suppress CVE for azure-identity 1.10.2 to 1.10.9 as it is only applicable on 1.10.1 and lower.
## Documentation

## Changelog

## Testing

### DIFF
--- a/gradle/owasp-suppression.xml
+++ b/gradle/owasp-suppression.xml
@@ -15,7 +15,7 @@
      </suppress>
     <suppress>
         <notes><![CDATA[
-        Suppress CVE-2023-36415 as this should only be applicable on version excluding 1.10.2.
+        Suppress CVE-2023-36415 as this should only be applicable on version up to but excluding 1.10.2.
         ]]></notes>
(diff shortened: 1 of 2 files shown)


gradle/owasp-suppression.xml
```diff
@@ -13,6 +13,13 @@
         <packageUrl regex="true">^pkg:maven/io\.netty/netty*@*.*$</packageUrl>
         <vulnerabilityName>CVE-2023-4586</vulnerabilityName>
      </suppress>
+    <suppress>
+        <notes><![CDATA[
+        Suppress CVE-2023-36415 as this should only be applicable on version up to but excluding 1.10.2.
```
siladu, line 18: Is that a mistake in the metadata then? Might be worth adding `until` if we could eventually remove this suppression, just so we can keep the file clean

siladu (approved):

jframe (approved):

PR #930, opened 2023-10-17
Circleci: Update nightly trivy scan

usmansaleem:
## PR Description
Update CircleCI dockerScan (trivy) nightly task. Use circleci base image for more up-to-date tools. Scan both amd64 and arm64 variants of web3signer develop tag. 

## Documentation

## Changelog

## Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -322,6 +322,7 @@ jobs:
           name: Scan develop image with trivy
           shell: /bin/sh
           command: |
+            docker images
             $HOME/bin/trivy image --exit-code 1 --no-progress --severity HIGH,CRITICAL --ignorefile "gradle/trivyignore.txt" --timeout 10m "docker.io/consensys/web3signer:develop"
       - notify
  
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,7 +326,7 @@ jobs:
           command: |
             docker pull docker.io/consensys/web3signer:develop
             docker manifest inspect consensys/web3signer:develop
-            $HOME/bin/trivy image --clear-cache --exit-code 1 --no-progress --severity HIGH,CRITICAL --ignorefile "gradle/trivyignore.txt" --timeout 10m "docker.io/consensys/web3signer:develop"
+            $HOME/bin/trivy image --exit-code 1 --no-progress --severity HIGH,CRITICAL --ignorefile "gradle/trivyignore.txt" --timeout 10m "docker.io/consensys/web3signer:develop"
       - notify
  
   publishOpenApiSpec:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -313,6 +313,8 @@ jobs:
     executor: trivy_executor
     steps:
       - prepare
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -312,24 +312,16 @@ jobs:
...
           name: Scan with trivy
           shell: /bin/sh
           command: |
-            rm -rf docker/test*
-            for FILE in $(ls docker)
-            do
-              docker pull -q "consensys/web3signer:develop-$FILE"
-              trivy -q image --exit-code 1 --no-progress --severity HIGH,CRITICAL --ignorefile "gradle/trivyignore.txt" --timeout 10m "consensys/web3signer:develop-$FILE" 
-            done
+            $HOME/bin/trivy image consensys/web3signer:develop-arm64 --image-src remote --exit-code 1 --quiet --no-progress --severity HIGH,CRITICAL --ignorefile "gradle/trivyignore.txt" --timeout 10m
```
jframe, line 323: Why don't we need to do the docker pull anymore? Does the --image-src remote on trivy take care of that?

.circleci/config.yml
```diff
@@ -312,24 +312,16 @@ jobs:
...
-            curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin
-      # remove the test files and only use the folder names
+          command: curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b $HOME/bin
+      # scan both develop-arm64 and develop-amd64 images using remote src.
       - run:
           name: Scan with trivy
           shell: /bin/sh
           command: |
-            rm -rf docker/test*
-            for FILE in $(ls docker)
```
siladu, line 328: What were the previous values of $FILE?

.circleci/config.yml
usmansaleem: correct. Technically we don't need to specify the `--image-src remote` as trivy tries docker, podman, containerd and lastly remote, since I was removing `setup_remote_docker`, I explicitly set it to remote so that it always fetch the image from docker hub.

.circleci/config.yml
usmansaleem: we used to have `docker/jdk17/Dockerfile` and `docker/jdk11/Dockerfile`. Hence we used to create/publish `consensys/web3signer:develop-java11` and `consensys/web3signer:develop-java17`. Since we now default to java17 variant, the `develop` tag now points to `develop-java17`. In addition, we further publish platform specific variants as well i.e. `develop-amd64` and `develop-arm64`. Similarly, `develop-java17-amd64` and `develop-java17-arm64` 

In summary, we only need to test/scan `develop` tag unless we introduce `java21` tag along with `java17` tag. We were also only scanning `develop-amd64` variant in past as CircleCI trivy job runs on Linux amd64 environment. Now we make sure to scan both amd64 and arm64 variants (as both of them are built separately).

jframe (approved):

siladu (approved):

PR #933, opened 2023-10-27
Re-add suppression in for false positive CVE-2023-4586

siladu:
Upgrade besu and netty related deps
Remove some older dep overrides


jframe (approved):
LGTM

PR #935, opened 2023-10-31
Remove KZG as it is not needed by web3signer

jframe:
## PR Description
Remove KZG as it is not needed by web3signer. This also removes the need to specify the trusted setup when using a custom network.

- Removed the js dependency vulnerability scanning as we don't use js and it's broken the build
- Removed the .net assembly scanner as it logs warning messages and we don't use it
- Upgraded Teku to latest version 23.10.0
- Changed spec to be initialised with the kzgNoop enabled
- Removed -Xtrusted-setup command line option since we no longer need it

## Documentation

## Changelog

## Testing

### DIFF
--- a/commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
+++ b/commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
@@ -126,15 +126,6 @@ public class Eth2SubCommand extends ModeSubCommand {
   private UInt64 denebForkEpoch;
 
   @CommandLine.Option(
-      names = {"--Xtrusted-setup"},
-      hidden = true,
-      paramLabel = "<STRING>",
-      description =
-          "The trusted setup which is needed for KZG commitments. Only required when creating a custom network. This value should be a file or URL pointing to a trusted setup.",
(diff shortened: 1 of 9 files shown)


commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -190,7 +181,7 @@ private void logNetworkSpecInformation() {
 
   private Eth2NetworkConfiguration createEth2NetworkConfig() {
     Eth2NetworkConfiguration.Builder builder = Eth2NetworkConfiguration.builder();
-    builder.applyNetworkDefaults(network);
+    builder.applyNetworkDefaults(network).kzgNoop(true);
```
siladu, line 184: Would this just be ignored if the chain was pre-deneb?

siladu (approved):

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
jframe: Yeah, it will be

PR #937, opened 2023-11-03
Prepare CircleCI config to run GCP Acceptance tests

usmansaleem:
## PR Description
Prepare CircleCI config to run GCP Acceptance tests

## Documentation

## Changelog

## Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,6 +224,10 @@ jobs:
           name: Acceptance Tests
           no_output_timeout: 20m
           command: |
+            if [ -n "$GCP_SERVICE_KEY" ]; then
+              echo "$GCP_SERVICE_KEY" | base64 --decode > $HOME/gcp_service_key.json
+              export GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcp_service_key.json
+            fi
             ./gradlew --no-daemon clean installDist
             ./gradlew --no-daemon --parallel acceptanceTest
       - notify


.circleci/config.yml
```diff
@@ -224,6 +224,10 @@ jobs:
           name: Acceptance Tests
           no_output_timeout: 20m
           command: |
+            if [ -n "$GCP_SERVICE_KEY" ]; then
+              echo "$GCP_SERVICE_KEY" | base64 --decode > $HOME/gcp_service_key.json
+              export GOOGLE_APPLICATION_CREDENTIALS=$HOME/gcp_service_key.json
```
jframe, line 229: Is this just for our ATs with how they work or will users also have to set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable?

jframe: Could just have the AT use the GCP_SERVICE_KEY env variable instead?

usmansaleem: End users, as well as web3signer developers, have to use "Application Default Credentials ADC) to setup authentication. We are using service user approach which uses service user key and GOOGLE_APPLICATION_CREDENTIALS environment variable. https://cloud.google.com/docs/authentication/provide-credentials-adc

jframe: That's not nice. It means we can't use our typical config and env variable approach. And google becomes a special case that is more difficult to use.

jframe: For our ATs it's fine though

siladu: nit: think it would be clearer if GCP_SERVICE_KEY was called GCP_SERVICE_KEY_BASE64

jframe (approved):

Issue #934, opened 2023-10-31
Keymanager API should not start if there is an error loading validators


pablomendezroyo:
The keymanager API should not start if there has been an error loading validators. I have been able to reproduce this edge case when having a clock error and the signer was not able to load validators via files, the keymanager API started and was returning 0 validators loaded. 

This is critical since the web3signer could be restarted in any situation and have the validators properly loaded.

IMO the web3signer should not start (`exit 1`) if there is an error loading validators.

The error
```
2023-10-30 23:48:10.984+00:00 | main | ERROR | Runner | Error loading signers
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: durationMillis must not be negative
	at java.util.concurrent.FutureTask.report(Unknown Source) ~[?:?]
	at java.util.concurrent.FutureTask.get(Unknown Source) ~[?:?]
	at tech.pegasys.web3signer.core.Runner.run(Runner.java:121) [web3signer-core-23.9.1.jar:23.9.1]
	at tech.pegasys.web3signer.core.Eth2Runner.run(Eth2Runner.java:382) [web3signer-core-23.9.1.jar:23.9.1]
	at tech.pegasys.web3signer.commandline.subcommands.ModeSubCommand.run(ModeSubCommand.java:33) [web3signer-commandline-23.9.1.jar:23.9.1]
	at picocli.CommandLine.executeUserObject(CommandLine.java:1939) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine.access$1300(CommandLine.java:145) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2314) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316) [picocli-4.6.2.jar:4.6.2]
	at picocli.CommandLine.execute(CommandLine.java:2078) [picocli-4.6.2.jar:4.6.2]
[...]

PR #941, opened 2023-11-07
Log warning when calculating time taken to load keys

usmansaleem:
## PR Description
Log warning instead of runtime exception if difference of time taken is negative i.e. system clock returned time in past.
## Fixed Issue(s)

#934 
## Documentation

## Changelog

## Testing

### DIFF
--- a/signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
+++ b/signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
@@ -43,6 +43,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.VisibleForTesting;
 
 public class SignerLoader {
 
@@ -120,7 +121,8 @@ public class SignerLoader {
     return MappedResults.newInstance(signingMetadataList, errorCount.get());
   }
 
-  private static Optional<String> calculateTimeTaken(final Instant start) {
+  @VisibleForTesting
+  static Optional<String> calculateTimeTaken(final Instant start) {
     final Instant now = Instant.now();
     final long timeTaken = Duration.between(start, now).toMillis();
     if (timeTaken < 0) {
--- a/signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
+++ b/signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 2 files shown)


signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -120,7 +120,12 @@ private MappedResults<SigningMetadata> convertConfigFileContent(
   }
 
   private static String calculateTimeTaken(final Instant start) {
-    return DurationFormatUtils.formatDurationHMS(Duration.between(start, Instant.now()).toMillis());
+    final Instant now = Instant.now();
+    final long timeTaken = Duration.between(start, now).toMillis();
+    if (timeTaken < 0) {
+      LOG.warn("System Clock returned time in past. Start: {}, Now: {}.", start, now);
+    }
+    return DurationFormatUtils.formatDurationHMS(Math.abs(timeTaken));
```
siladu, line 128: Think this could still throw the error when `timeTaken < 0`?

siladu (changes requested):
Would be good to expose current bug in a unit test as well

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
siladu: I see the Math.abs will avoid the error, but it will also mean we report inaccurate time taken so better not to report anything in this case

jframe: Agree ABS doesn't seem like the right solution. We should just not return anything in that case. So maybe just the return type to be Optional<String> and return the empty in the invalid case.

usmansaleem: addressed.

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -42,6 +43,7 @@
 import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.VisibleForTesting;
```
siladu, line 46: nit: other code uses `import com.google.common.annotations.VisibleForTesting`

signing/src/test/java/tech/pegasys/web3signer/signing/config/SignerLoaderTest.java
```diff
@@ -293,6 +296,20 @@ void callingLoadTwiceOnlyLoadSignersFromModifiedConfigFiles() throws IOException
...
 
+  @Test
+  void calculateTimeTakenWithFutureTimeShouldReturnEmpty() {
+    final Optional<String> timeTaken =
+        SignerLoader.calculateTimeTaken(Instant.now().plus(5, ChronoUnit.MINUTES));
+    assertThat(timeTaken).isEmpty();
+  }
+
+  @Test
+  void calculateTimeTakenWithPastTimeShouldReturnValue() {
```
siladu, line 307: nit: maybe boundary case of same time too?

siladu (approved):

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
usmansaleem: good catch ... fixed.

PR #980, opened 2024-03-15
fix: Make cached GVR thread-safe

### LINKED ISSUE #978: ERROR:  duplicate key value violates unique constraint "metadata_pkey"
ERROR:  duplicate key value violates unique constraint "metadata_pkey : This error is occurring after starting postgres container . We are facing this issue after upgrading to postgres:14.9 from postgres:12. The data from previous version has been deleted before starting the new version.

Web3signer or teku validator client does not show any errors during this timestamp. No attestations are missed. 

We are getting this error when postgres is initiated. Log file is attached below

[postgres_logs.txt](https://github.com/Consensys/web3signer/files/14540172/postgres_logs.txt)

usmansaleem:
## PR Description
Make cached GVR calculation thread-safe. During Web3Signer first boot, if multiple requests for signing comes in, it can result in multiple threads attempting to insert gvr in a fresh database. This would show an constraint violation in database logs.

## Fixed Issue(s)

Fixes #978 
## Documentation

## Changelog

## Testing

### DIFF
--- a/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidator.java
+++ b/slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidator.java
@@ -18,8 +18,10 @@ import tech.pegasys.web3signer.slashingprotection.dao.MetadataDao;
 
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.ExecutionException;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import net.jodah.failsafe.Failsafe;
 import net.jodah.failsafe.FailsafeExecutor;
 import net.jodah.failsafe.RetryPolicy;
@@ -37,8 +39,7 @@ public class GenesisValidatorRootValidator {
   private final MetadataDao metadataDao;
   private final FailsafeExecutor<Object> failsafeExecutor;
 
-  private final AtomicReference<Bytes32> cachedGVRRef = new AtomicReference<>();
-  private final Object lockForCachedGVR = new Object();
+  private final Cache<String, Bytes32> gvrCache = CacheBuilder.newBuilder().maximumSize(1).build();
 
   public GenesisValidatorRootValidator(final Jdbi jdbi, final MetadataDao metadataDao) {
     this.jdbi = jdbi;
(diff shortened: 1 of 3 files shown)


slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidator.java
```diff
@@ -65,6 +60,30 @@ public boolean checkGenesisValidatorsRootAndInsertIfEmpty(final Bytes32 genesisV
     }
   }
 
+  private Bytes32 insertAndCacheGVR(final Bytes32 genesisValidatorsRoot) {
+    var cachedGVR = cachedGVRRef.get();
+    if (cachedGVR == null) {
+      LOG.debug("No cached GVR found, fetching from database");
+      synchronized (lockForCachedGVR) {
```
jframe, line 67: Can one of the AtomicReference atomic operations be used to update the value to avoid the synchronized block?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidator.java
```diff
@@ -36,7 +37,8 @@ public class GenesisValidatorRootValidator {
   private final MetadataDao metadataDao;
   private final FailsafeExecutor<Object> failsafeExecutor;
 
-  private Bytes32 cachedGenesisValidatorRoot;
+  private final AtomicReference<Bytes32> cachedGVRRef = new AtomicReference<>();
```
jframe, line 40: Looks like we are using our cache, think it would be better to use the Guava cache rather than writing our thread safe implementation

usmansaleem: Hmm, correct. Guava Cache would achieve the same result.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidatorTest.java
```diff
@@ -64,6 +69,39 @@ void verifyCachedGVRIsUsedForNewGVR(final Jdbi jdbi) {
...
+            } finally {
+              latch.countDown(); // Ensure the latch is counted down even if an exception is thrown
+            }
+          });
+    }
+
+    latch.await(); // Wait for all threads to finish
+    executorService.shutdown(); // Shutdown the executor service
+
+    assertThat(allTrue.get()).isTrue();
```
siladu, line 100: Don't think this check or the latch is doing anything. 

When this fails, it's due to the number of times findGenesisValidatorsRoot is called

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidator.java
siladu: or ConcurrentHashMap?

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidatorTest.java
usmansaleem: `latch.await();` to make sure that all threads have joined. Agreed with allTrue. I added findByGVR assertions later on :). Will be removing allTrue assertion.

usmansaleem: decided to rename allTrue and still kept it.

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidatorTest.java
```diff
@@ -64,6 +68,36 @@ void verifyCachedGVRIsUsedForNewGVR(final Jdbi jdbi) {
...
+              allCachedGVRMatches.set(false);
+            }
+          });
+    }
+
+    // Shutdown the executor service, no new tasks will be accepted
+    executorService.shutdown();
+
+    // wait for all threads to finish
+    var successfulTerminate = executorService.awaitTermination(5, MINUTES);
```
siladu, line 93: Could probably even make this SECONDS so we don't slow down the tests too much if this were to happen

siladu (approved):
LGTM

slashing-protection/src/test/java/tech/pegasys/web3signer/slashingprotection/validator/GenesisValidatorRootValidatorTest.java
usmansaleem: Reduced it to 1 minute.

PR #981, opened 2024-03-18
build - suppress unrelated owasp warnings and update azure libraries

usmansaleem:
## PR Description
Update gradle build - suppress unrelated owasp warnings and update azure libraries.

## Documentation

## Changelog

## Testing

### DIFF
--- a/gradle/owasp-suppression.xml
+++ b/gradle/owasp-suppression.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
     <!-- See https://jeremylong.github.io/DependencyCheck/general/suppression.html for examples -->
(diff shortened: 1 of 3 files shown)


gradle/owasp-suppression.xml
```diff
@@ -1,40 +1,26 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
     <!-- See https://jeremylong.github.io/DependencyCheck/general/suppression.html for examples -->
-    <suppress until="2024-03-16">
+    <suppress until="2024-06-16">
```
jframe, line 4: Shame that there is still no fix for these

jframe (approved):

gradle/owasp-suppression.xml
usmansaleem: yes, I tried up update versions, but it is still reporting it incorrectly.

PR #982, opened 2024-03-20
Extension Signing request endpoint

### LINKED ISSUE #976: Let the user sign an arbitrary message with its keystore
Currently, web3signer only permits signing certain types of data (as body payloads) in the `/api/v1/eth2/sign/{identifier}` ETH2 endpoint or  `/api/v1/eth1/sign/{identifier}` ETH1 endpoint. See https://consensys.github.io/web3signer/web3signer-eth2.html

It would be nice to add an endpoint (probably in ETH2 mode or both ETH1 & ETH2) that signs an arbitrary payload sent by the user. Something like:

```
"message": {
    "platform": "dappnode",
    "timestamp": "185921877",
    "pubkey":
    "0xaf294f5663defdba078e0c16b349952ab762f8019f42f45fc0a369666e12d836247e4d7a3e3bd564826dcffcfdc50114"
},
```

This is outside web3signer main scope, it would let users sign messages not necessarily related to the eth2 protocol

usmansaleem:
## PR Description
An extension sign endpoint `/api/v1/eth2/ext/sign/:identifier` which can be enabled with cli option `--Xsigning-ext-enabled=true` The cli option is associated with `eth2` subcommand. For example `./web3signer eth2 --Xsigning-ext-enabled=true ...`. This extension covers some additional signing types and use-cases which are not covered by the remoting API specs and hence should be used at your own risk.

The format of request payload is:
```
{
  "type" : "PROOF_OF_VALIDATION",
  "platform" : "dappnode",
  "timestamp" : "1711338489397"
}
```
Only `PROOF_OF_VALIDATION` extension type is supported at the moment. Any other type would result in an error.

The signature is hex encoded string calculated as:

`signature = BLS.sign(payload.getBytes()).asHexString()`

For default accept type (`*/*`) and `JSON` accept type headers, the response is in following format:
**_Note_: 
Payload is returned in the response as Base64 encoded String.
Signature is returned as Hex encoded String**

```
{
    "payload": "eyJ0eXBlIjoiUFJPT0ZfT0ZfVkFMSURBVElPTiIsInBsYXRmb3JtIjoiQVQiLCJ0aW1lc3RhbXAiOiIxNzE0NTI0NTQ0NDA3In0=",
    "signature": "0xb792ace1bae87b02457957bfa753065094e153855049074c880b943e7239d4277f4b2384d6aa19a3c03043b92bc8e90302ae48b9ac6b0a5e50796fc5e1d3993e9df6621ad6906e7a754147a9ccacdc9e7c18c1be355a647db19591904575bd30"
}
```
This would allow the users of this endpoint to verify the payload is signed by appropriate private key.

## Fixed Issue(s)

Fixes #976 
## Documentation

## Changelog

## Testing


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/GenericSignForIdentifierHandler.java
```diff
@@ -0,0 +1,85 @@
...
+  private void respondWithSignature(
+      final RoutingContext routingContext, final String signature, final String body) {
+    final String acceptableContentType =
+        getAcceptableContentType(routingContext.parsedHeaders().accept());
+    final String responseBody;
+
+    if (acceptableContentType.equals(JSON_UTF_8)) {
+      responseBody =
+          new JsonObject()
+              .put(
```
jframe, line 59: Don't need the payload in the response. This is a different response format to all other remote signing requests

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/GenericSignForIdentifierHandler.java
```diff
@@ -0,0 +1,85 @@
...
+  public GenericSignForIdentifierHandler(final SignerForIdentifier<?> signerForIdentifier) {
+    this.signerForIdentifier = signerForIdentifier;
+  }
+
+  @Override
+  public void handle(final RoutingContext routingContext) {
+    final String identifier = normaliseIdentifier(routingContext.pathParam("identifier"));
+    final String body = routingContext.body().asString();
+    signerForIdentifier
+        .sign(identifier, Bytes.of(body.getBytes(StandardCharsets.UTF_8)))
```
jframe, line 44: Think we should use a custom domain and sign using that. I'm concerned that with it is right now as you create input that could be the equivalent of a block or attestation signing but with the slashing protection check being bypassed.

core/src/main/java/tech/pegasys/web3signer/core/config/BaseConfig.java
```diff
@@ -64,4 +64,6 @@ public interface BaseConfig {
   boolean keystoreParallelProcessingEnabled();
 
   int getVertxWorkerPoolSize();
+
+  boolean isGenericSigningExtensionEnabled();
```
jframe, line 68: This is going to apply to both eth1 and eth2 modes, we should move this up into eth2 config

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
```diff
@@ -282,6 +287,29 @@ private void registerHttpHostAllowListHandler(final Router router) {
     router.route().handler(new HostAllowListHandler(baseConfig.getHttpHostAllowList()));
   }
 
+  protected void addGenericSignHandler(
```
jframe, line 290: Why is this in the base runner? I thought this was specific to eth2 signing? We already have eth_sign RPC for eth1.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
```diff
@@ -66,6 +66,8 @@ public class Signer extends FilecoinJsonRpcEndpoint {
   public static final String ETH2_PUBLIC_KEYS = "/api/v1/eth2/publicKeys"; // bls keys
   public static final String RELOAD_ENDPOINT = "/reload";
 
+  public static final String GENERIC_SIGN_EXT_ENDPOINT = "/eth/v1/ext/sign/{identifier}";
```
jframe, line 69: Does this need its own endpoint? Thought it would be under the ETH2_SIGN_ENDPOINT but using a different domain. Or prefixed with `/api/v1/eth2/` path so it is consistent with other endpoints.

core/src/main/java/tech/pegasys/web3signer/core/Runner.java
usmansaleem: 👍🏼 , forgot about that. Will make it eth2 specific.

siladu (commented):
What do we now in terms of API docs for this?

I'm thinking it will be hard for a future user to understand how to use this API, e.g. what should they put for "platform"? Also, wondering if platform is strictly necessary in the request?

core/src/main/java/tech/pegasys/web3signer/core/service/http/OpenApiOperationsId.java
```diff
@@ -13,6 +13,7 @@
 package tech.pegasys.web3signer.core.service.http;
 
 /** Operation IDs as defined in web3signer.yaml */
+@Deprecated(forRemoval = true)
```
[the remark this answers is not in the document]
usmansaleem: this class was used when we had vertx "openapi router". We have since removed that router and this interface is unused.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/ProofOfValidationBody.java
```diff
@@ -0,0 +1,20 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.web3signer.core.service.http.handlers.signing;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record ProofOfValidationBody(
+    @JsonProperty(value = "type", required = true) SigningExtensionType type,
+    @JsonProperty(value = "platform", required = true) String platform,
+    @JsonProperty(value = "timestamp", required = true) String timestamp) {}
```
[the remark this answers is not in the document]
usmansaleem: done.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SigningExtensionHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+    // validate that we have correct incoming json body
+    try {
+      JSON_MAPPER.readValue(body, ProofOfValidationBody.class);
+    } catch (final JsonProcessingException | IllegalArgumentException e) {
+      routingContext.fail(BAD_REQUEST);
+      return;
+    }
+
+    final Bytes payload = Bytes.wrap(body.getBytes(UTF_8));
```
jframe, line 62: Is the conversion from to bytes well defined such that this conversion can be easily done in other languages to convert the payload and verify the signature?

usmansaleem: Tested above in NodeJS/Typescript. Seems to be working as expected. https://github.com/usmansaleem/node_bls_verify

jframe (approved):

PR #995, opened 2024-06-03
Upgrade to Java 21

### LINKED ISSUE #1013: Migrate Web3Signer build requirements to Java 21
Migrate Web3Signer build requirements to Java 21.

usmansaleem:
## PR Description
Upgrade build and runtime to Java 21. Following changes are also made to achieve this:
- Upgraded to Gradle 8.9
- Various plugin versions updated
- Various dependencies versions updated
  - Besu and Teku 
- Owasp suppressions updated
- Removed gatling performance test because it is not compatible with latest Gradle.

Logs after upgrading to Java 21:
```
Web3SignerApp | Version = web3signer/v24.2.0+20-gb2fe8d6/linux-aarch_64/-eclipseadoptium-openjdk64bitservervm-java-21
```

## Fixed Issue(s)

fixes #1013 

## Documentation

## Changelog

## Testing


jframe (commented):
We need give some notice about the breaking change of requiring Java 21 before merging this in. So for this upcoming release we can place a notice in our changelog and the announcement about the upcoming change and next monthly release change to using Java 21.

.circleci/config.yml
```diff
@@ -224,18 +224,11 @@ jobs:
       - capture_test_reports
 
   performanceTests:
```
jframe, line 226: can we just remove this step instead of doing the echo?

acceptance-tests/build.gradle
```diff
@@ -80,7 +80,7 @@ dependencies {
   testImplementation 'com.azure:azure-security-keyvault-keys'
   testImplementation 'com.azure:azure-security-keyvault-secrets'
 
-  gatlingImplementation configurations.getByName("testImplementation")
+  // gatlingImplementation configurations.getByName("testImplementation")
```
jframe, line 83: remove commented line

acceptance-tests/build.gradle
```diff
@@ -90,15 +90,15 @@ dependencies {
...
+  delete layout.buildDirectory.dir("besu")
 }
 
 task downloadBesu(type: Download) {
   src {
     // see gradle.properties for Besu URL
     return new SimpleTemplateEngine().createTemplate(besuDistroUrl).make(["besuVersion":"$besuVersion"]).toString()
   }
-  dest new File(buildDir, "besu-${besuVersion}.tar.gz")
+  dest layout.buildDirectory.file("downloads/besu-${besuVersion}.tar.gz")
```
jframe, line 101: Does the downloads directory get created if doesn't exist? And will work on Windows with the hardcoded `/` directory separator?

build.gradle
```diff
@@ -21,28 +21,27 @@ import java.text.SimpleDateFormat
...
+  id 'org.owasp.dependencycheck' version "10.0.3"
   id 'java-test-fixtures'
-  id 'com.diffplug.spotless' version '6.2.0'
-  id 'com.github.ben-manes.versions' version '0.27.0'
-  id 'com.github.jk1.dependency-license-report' version '2.0'
-  id 'io.spring.dependency-management' version '1.0.11.RELEASE'
-  id 'me.champeau.gradle.jmh' version '0.5.3' apply false
-  id 'net.ltgt.errorprone' version '2.0.2'
-  id 'org.ajoberstar.grgit' version '5.2.0'
+  id 'com.diffplug.spotless' version '7.0.0.BETA1'
```
jframe, line 34: Do need to use a beta version of spotless? it's not in production code so it's probably ok.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/LogErrorHandler.java
```diff
@@ -26,9 +26,24 @@ public class LogErrorHandler implements Handler<RoutingContext> {
...
     if (failureContext.failed()) {
-      LOG.error(
-          String.format("Failed request: %s", failureContext.request().absoluteURI()),
-          failureContext.failure());
+      String requestUri;
+      try {
+        requestUri = failureContext.request().absoluteURI();
+      } catch (final NullPointerException e) {
+        // absoluteURI can throw an NPE if host or port is missing
+        requestUri = "Error in calculating request URI";
```
jframe, line 34: More descriptive error would be nicer here along line of invalid URI due to missing host or port.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbLocker.java
```diff
@@ -17,13 +17,23 @@
 public class DbLocker {
 
   public enum LockType {
-    BLOCK,
-    ATTESTATION
+    BLOCK(0),
+    ATTESTATION(1),
+    ;
```
jframe, line 22: This formatting is a bit nasty, I'm guessing this spotless enforcing this though?

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/LogErrorHandler.java
usmansaleem: sure. This NPE happens when invalid host header is involved from the client. Latest vertx has started throwing an NPE. I'll update the error description.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/DbLocker.java
usmansaleem: Let me see if I am able to remove that `;`.

usmansaleem: removed ;

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/LogErrorHandler.java
usmansaleem: refactored it a bit.

build.gradle
usmansaleem: I was double minded on this one, and finally added it to catch any surprising stuff in advance. Since it is build-only as you mentioned, we should be alright about it.

acceptance-tests/build.gradle
usmansaleem: This works on nix without any problems. Need to check our whole build on Windows environment.

acceptance-tests/build.gradle
usmansaleem: done.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
```diff
@@ -204,16 +204,29 @@ public static String signPath(final KeyType keyType) {
...
   }
 
-  public Set<String> getMetricsMatching(final List<String> metricsOfInterest) {
+  public Map<String, String> getMatchedMetrics(final Set<String> metricsOfInterest) {
     final Response response =
         given().baseUri(getMetricsUrl()).contentType(ContentType.JSON).when().get(METRICS_ENDPOINT);
 
-    final List<String> lines =
-        Arrays.asList(response.getBody().asString().split(String.format("%n")).clone());
+    return convertToLines(response.getBody().asString()).stream()
```
jframe, line 211: This could be simplified to:
```response.getBody().asString().lines()```

Removing the need for convertToLines

File: CHANGELOG.md [hunk 8]
```diff
@@ -8,6 +8,9 @@ this release. The Web3Signer docker image will also use Java 21, however, binary
 require Java 21 to be available on the host machine.
 - This is the last Web3Signer release to use the "filecoin" mode. The "filecoin" mode will be removed in a future release.
 
+### Breaking Changes
+- Use Java 21 for build and runtime. Remove Java 17 variant of docker image. zip/tar.gz distributions will require Java 21 or above to run Web3Signer.
+
```
jframe, line 13: Also, we need to mention the changing of metrics to use openmetrics standard, users will need to update their dashboard

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
```diff
@@ -204,16 +204,29 @@ public static String signPath(final KeyType keyType) {
     return keyType == BLS ? ETH2_SIGN_ENDPOINT : ETH1_SIGN_ENDPOINT;
   }
 
-  public Set<String> getMetricsMatching(final List<String> metricsOfInterest) {
+  public Map<String, String> getMatchedMetrics(final Set<String> metricsOfInterest) {
```
jframe, line 207: nit: IMHO the previous name was clearer

build.gradle
```diff
@@ -135,26 +134,16 @@ allprojects {
         include '**/*.java'
         exclude '**/.gradle/**'
         exclude '**/build/**'
+        exclude '.openapidoc/**'
       }
-      removeUnusedImports()
+      // removeUnusedImports() // conflicts with googleJavaFormat for 7.0.0.BETA1
```
jframe, line 139: remove commented line

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/LogErrorHandler.java
```diff
@@ -24,15 +26,34 @@ public class LogErrorHandler implements Handler<RoutingContext> {
...
+      }
+
       // Let the next matching route or error handler deal with the error, we only handle logging
       failureContext.next();
     } else {
       LOG.warn("Error handler triggered without any propagated failure");
     }
   }
+
+  private static String getRequestUri(RoutingContext failureContext) {
```
jframe, line 49: I think the method name isn't quite right as the error case it doesn't return a requestUri. Would be better to return an empty and just include that error message in the handle method.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/signer/Signer.java
usmansaleem: nice suggestion, updated.

CHANGELOG.md
usmansaleem: Updated changelog, let me know if wording needs to be changed.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/LogErrorHandler.java
usmansaleem: agreed. Updated the method to return Optional.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/LogErrorHandler.java
```diff
@@ -24,15 +26,34 @@ public class LogErrorHandler implements Handler<RoutingContext> {
 
   @Override
   public void handle(final RoutingContext failureContext) {
-
     if (failureContext.failed()) {
-      LOG.error(
-          String.format("Failed request: %s", failureContext.request().absoluteURI()),
-          failureContext.failure());
+      final String requestUri =
```
jframe, line 30: Not going to block on this, but the variable name is a bit misleading as sometimes it is an error message sometimes it's the requestUri.

jframe (approved):

PR #1015, opened 2024-08-22
Moved teku deserializers from teku into web3signer

rolfyone:
Web3signer is referencing a few deserializers / serializers from teku, and I'm removing them from teku, so I've recreated them in web3signer to reduce integration issues...

I did have problems running tests, but I'm not sure why, this should be a relatively clean lift and shift...

## Documentation

## Changelog

- [X] I thought about adding a changelog entry, and added one if I deemed necessary.

## Testing

- [X] I thought about testing these changes in a realistic/non-local environment.

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/service/http/SigningObjectMapperFactory.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/service/http/SigningObjectMapperFactory.java
@@ -26,17 +26,17 @@ import tech.pegasys.teku.infrastructure.jackson.deserializers.bytes.DoubleSerial
 import tech.pegasys.teku.infrastructure.jackson.deserializers.uints.UInt64Deserializer;
 import tech.pegasys.teku.infrastructure.jackson.deserializers.uints.UInt64Serializer;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszBitvector;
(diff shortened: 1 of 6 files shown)


jframe (approved):
LGTM

core/src/main/java/tech/pegasys/web3signer/core/service/http/SigningObjectMapperFactory.java
```diff
@@ -26,17 +26,17 @@
 import tech.pegasys.teku.infrastructure.jackson.deserializers.uints.UInt64Deserializer;
 import tech.pegasys.teku.infrastructure.jackson.deserializers.uints.UInt64Serializer;
```
usmansaleem, line 27: @rolfyone what about uints deserializers? Do we need to migrate those as well?

usmansaleem (approved):
LGTM, one minor comment.

core/src/main/java/tech/pegasys/web3signer/core/service/http/SigningObjectMapperFactory.java
rolfyone: the ones in infra are fine, if we end up moving to TypeDefinitions they won't be referenced ultimately...

PR #1026, opened 2024-10-10
Add support for BlobTx(eip4844) for eth_signTransaction

NazariiDenha:
## Documentation

## Changelog

## Testing


usmansaleem (changes requested):
@NazariiDenha Thank you for your contribution. Can you also update the changelog entry please.

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
```diff
@@ -77,6 +82,35 @@ private static byte[] prependEip1559TransactionType(byte[] bytesToSign) {
         .array();
   }
 
+  private static final char[] HEX_ARRAY = "0123456789ABCDEF".toCharArray();
+  public static String bytesToHex(byte[] bytes) {
```
usmansaleem, line 86: you can use `Bytes toHexString` as that is used extensively in the code.

Issue #1017, opened 2024-09-03
Commit boost


jframe:
Add support for commit boost in web3signer. This will require adding a new API to sign commit boost requests.

CLI Options in Web3Signer to enable and use Commit Boost support:

 `--commit-boost-api-enabled=<true|false>`. To enable commit boost API
` --proxy-keystores-path <PATH>`. Path to the readable/writeable directory where encrypted proxy keys in v4 (BLS) and v3 (SECP) formats will be stored.
` --proxy-keystores-password-file <FILE>`. Path to the file containing password to encrypt/decrypt all proxy keys.

Resources:
https://github.com/Commit-Boost
API https://github.com/Commit-Boost/commit-boost-client/blob/main/api/signer-api.yml
Recording of the Ethereum Sequencing and Preconfirmations Call in which they discuss commit-boost:  https://www.youtube.com/watch?v=jrm4ZUoj9xY
https://ethresear.ch/t/commit-boost-proposer-platform-to-safely-make-commitments/20107

PR #1031, opened 2024-10-20
feat: Commit boost API - Get Public Keys

usmansaleem:
## PR Description
feat: Commit boost API - Get Public Keys. Implementing commit boost API as defined at https://commit-boost.github.io/commit-boost-client/api

### **Note:** This PR does not return compressed SECP public key. This will be handled in PR #1033 

 - [x] CLI options. 
 - [x] `--commit-boost-api-enabled=<true|false>`. To enable commit boost API 
 - [x] `--proxy-keystores-path`. Path to the directory that will read and store encrypted proxy keys in v4 (BLS) and v3 (SECP) formats.
 - [x] `--proxy-keystores-password-file`. The path to file that contains password to encrypt and decrypt proxy keystores.
 - [x] Implement route and handlers for `/signer/v1/get_pubkeys`
 - [x] Load proxy signers from local directories

## Fixed Issue(s)

See #1017

## Documentation

## Changelog

## Testing

### DIFF
--- a/commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
+++ b/commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
@@ -29,6 +29,7 @@ import tech.pegasys.web3signer.commandline.PicoCliAwsSecretsManagerParameters;
 import tech.pegasys.web3signer.commandline.PicoCliEth2AzureKeyVaultParameters;
 import tech.pegasys.web3signer.commandline.PicoCliGcpSecretManagerParameters;
 import tech.pegasys.web3signer.commandline.PicoCliSlashingProtectionParameters;
+import tech.pegasys.web3signer.commandline.config.PicoCommitBoostApiParameters;
 import tech.pegasys.web3signer.commandline.config.PicoKeystoresParameters;
 import tech.pegasys.web3signer.common.config.AwsAuthenticationMode;
 import tech.pegasys.web3signer.core.Eth2Runner;
@@ -164,6 +165,7 @@ public class Eth2SubCommand extends ModeSubCommand {
   @Mixin private PicoKeystoresParameters keystoreParameters;
   @Mixin private PicoCliAwsSecretsManagerParameters awsSecretsManagerParameters;
   @Mixin private PicoCliGcpSecretManagerParameters gcpSecretManagerParameters;
+  @Mixin private PicoCommitBoostApiParameters commitBoostApiParameters;
   private tech.pegasys.teku.spec.Spec eth2Spec;
 
   public Eth2SubCommand() {
@@ -183,7 +185,8 @@ public class Eth2SubCommand extends ModeSubCommand {
         gcpSecretManagerParameters,
         eth2Spec,
         isKeyManagerApiEnabled,
-        signingExtEnabled);
+        signingExtEnabled,
+        commitBoostApiParameters);
   }
 
   private void logNetworkSpecInformation() {
@@ -261,6 +264,7 @@ public class Eth2SubCommand extends ModeSubCommand {
     validateKeystoreParameters(keystoreParameters);
     validateAwsSecretsManageParameters();
     validateGcpSecretManagerParameters();
+    commitBoostApiParameters.validateParameters();
   }
 
   private void validateGcpSecretManagerParameters() {
--- a/core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/commitboost/CommitBoostPublicKeysHandler.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/commitboost/CommitBoostPublicKeysHandler.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2024 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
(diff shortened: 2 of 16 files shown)


commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
```diff
@@ -164,6 +165,7 @@ private static class NetworkCliCompletionCandidates extends ArrayList<String> {
   @Mixin private PicoKeystoresParameters keystoreParameters;
   @Mixin private PicoCliAwsSecretsManagerParameters awsSecretsManagerParameters;
   @Mixin private PicoCliGcpSecretManagerParameters gcpSecretManagerParameters;
+  @Mixin private PicoCommitBoostApiParameters commitBoostApiParameters;
```
jframe, line 168: It's not quite right to add this into the eth2 API as it has secp signing as well. I think it would be better to add this to the default or base command rather than a subcommand as that's where we want all the other functionality eventually too

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/commitboost/CommitBoostPublicKeysHandler.java
```diff
@@ -0,0 +1,81 @@
...
+  public CommitBoostPublicKeysHandler(final List<ArtifactSignerProvider> artifactSignerProviders) {
+    this.artifactSignerProviders = artifactSignerProviders;
+  }
+
+  @Override
+  public void handle(final RoutingContext context) {
+    // obtain DefaultArtifactSignerProvider as that is the only one we are dealing in eth2 mode.
+    final ArtifactSignerProvider artifactSignerProvider =
+        artifactSignerProviders.stream()
+            .filter(provider -> provider instanceof DefaultArtifactSignerProvider)
```
jframe, line 50: Why does commit boost only work with the DefaultArtifactSignerProvider?

signing/src/main/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProvider.java
```diff
@@ -60,6 +71,49 @@ public Future<Void> load() {
...
+                }
+                signers.forEach(
+                    (identifier, __) -> {
+                      final Path identifierPath =
+                          keystoreParameter.getKeystoresPath().resolve(identifier);
+                      if (identifierPath.toFile().canRead()
+                          && identifierPath.toFile().isDirectory()) {
+                        final Path v4Dir = identifierPath.resolve("v4");
+
+                        if (v4Dir.toFile().canRead() && v4Dir.toFile().isDirectory()) {
```
jframe, line 88: Why are separate directories for each proxy key version needed? Pretty sure the file formats have a version

commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/Eth2SubCommand.java
usmansaleem: Because the upcoming GenerateProxyKey requires Spec (to calculate genesis fork version) and gvr to compute the root. The spec and gvr is specific to "eth2" mode, hence the commit boost API is specific to eth2 mode as well and won't work in eth1 mode at all. 

I can still move it to base level, but we will only be using it in eth2 mode.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/commitboost/CommitBoostPublicKeysHandler.java
usmansaleem: commit boost only works in eth2 mode. The DefaultArtifactSignerProvider will load both v3 and v4 proxy signers for each of the main BLS signer. The SECPArtifactSigner that is used in eth1 mode (the eth address is used as identifier there) is not applicable for commit boost routes and handlers.

signing/src/main/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProvider.java
usmansaleem: This allow easier loading using existing BLS and SECP bulkloading utility classes. Otherwise, I would have to parse json, check the version, and then invoke BLS or SECP decryption logic. Since the proxy keys Generate logic (the other PR) is controlled programmatically, the user is not involved in placing these files.

usmansaleem: each delegator (BLS Signer) can have BLS Proxy Signers and SECP Proxy Signers. Hence:
```
/commit_boost_keystores_dir
..../0xblskey0..n/
.........../0xblskey0..n/v3/0xproxy_secp_key0...n
.........../0xblskey0..n/v4/0xproxy_bls_key0...n

```

signing/src/test/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProviderTest.java
```diff
@@ -76,9 +99,171 @@ void signerProviderCanMapInTwoSigners() {
...
+
+    assertThat(key2ProxyPublicKeys.get(KeyType.BLS))
+        .containsExactlyInAnyOrder(getPublicKeysArray(key2ProxyKeyPairs));
+    assertThat(key2ProxyPublicKeys.get(KeyType.SECP256K1))
+        .containsExactlyInAnyOrder(getSecpPublicKeysArray(key2SecpKeyPairs));
+  }
+
+  @Test
+  void emptyProxySignersAreLoadedSuccessfully() {
+    // enable commit boose without existing proxy keys
```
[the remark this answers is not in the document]
usmansaleem: done

signing/src/main/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProvider.java
usmansaleem: as discussed, updated to use BLS and SECP256K1 as subdirectories for each signer public key.

jframe (approved):

PR #1037, opened 2024-11-21
Refactor EthPublicKeyUtils to convert public key between Java, BC and Web3J libraries

usmansaleem:
## PR Description
Refactor EthPublicKeyUtils to convert public key between Java, BC and Web3J libraries

## Documentation

## Changelog

## Testing


core/src/test/java/tech/pegasys/web3signer/core/service/jsonrpc/EthAccountsResultProviderTest.java
```diff
@@ -22,30 +22,39 @@
...
+  private ECPublicKey publicKeyB;
+  private ECPublicKey publicKeyC;
+
+  private String addressA;
+  private String addressB;
+  private String addressC;
+
+  @BeforeEach
+  public void init() throws GeneralSecurityException {
+    SecureRandom secureRandom = new SecureRandom();
```
jframe, line 50: Is this going to cause an issue with entropy in CI creating this multiple times?

core/src/test/java/tech/pegasys/web3signer/core/service/Eth1AddressSignerIdentifierTest.java
```diff
@@ -15,53 +15,89 @@
...
+    assertThat(signerIdentifier.toStringIdentifier()).isEqualTo(expectedAddress);
+    assertThat(signerIdentifier.toStringIdentifier()).doesNotStartWithIgnoringCase("0x");
+    assertThat(signerIdentifier.toStringIdentifier()).isLowerCase();
+  }
+
+  public static String convertAlphabetsToUpperCase(final String input) {
+    final char[] chars = input.toCharArray();
+    int count = 0;
+
+    for (int i = 0; i < chars.length && count < 2; i++) {
```
jframe, line 94: why do only the first two characters need to be uppercase?

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/EthPublicKeyUtils.java
```diff
@@ -12,63 +12,254 @@
...
+   * @return The generated java security key pair
+   * @throws GeneralSecurityException If there is an issue generating the key pair
+   */
+  public static KeyPair createSecp256k1KeyPair(final SecureRandom random)
+      throws GeneralSecurityException {
+    final KeyPairGenerator keyPairGenerator =
+        KeyPairGenerator.getInstance(EC_ALGORITHM, BC_PROVIDER);
+    final ECGenParameterSpec ecGenParameterSpec = new ECGenParameterSpec(SECP256K1_CURVE);
+    if (random != null) {
+      keyPairGenerator.initialize(ecGenParameterSpec, random);
```
jframe, line 86: why wouldn't we always create the key pair using a securerandom?

core/src/test/java/tech/pegasys/web3signer/core/service/jsonrpc/EthAccountsResultProviderTest.java
usmansaleem: probably it will, let me create a static instance.

core/src/test/java/tech/pegasys/web3signer/core/service/Eth1AddressSignerIdentifierTest.java
usmansaleem: The intent was just to test that if mixed case hex are used as identifiers, they are normalized to be lower case. Since we are using random public keys, we are just converting the first two alpha characters to upper case.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/EthPublicKeyUtils.java
usmansaleem: I think my original intent was that we might send in null SecureRandom. I am going to remove this check as we will always send-in an initialize SecureRandom class.

core/src/test/java/tech/pegasys/web3signer/core/service/Eth1AddressSignerIdentifierTest.java
usmansaleem: I've changed the method name and its access modifier to make the intent clear.

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/EthPublicKeyUtilsTest.java
```diff
@@ -62,26 +60,72 @@ public void createsPublicKeyFromBytes() {
...
+      keyPair = EthPublicKeyUtils.createSecp256k1KeyPair(new SecureRandom());
+    } catch (GeneralSecurityException e) {
+      throw new RuntimeException(e);
+    }
+    return Stream.of(
+        // Compressed (33 bytes)
+        Bytes.fromHexString(
+            EthPublicKeyUtils.toHexStringCompressed((ECPublicKey) keyPair.getPublic())),
+        // Uncompressed without prefix (64 bytes)
+        Bytes.fromHexString(EthPublicKeyUtils.toHexString((ECPublicKey) keyPair.getPublic())));
```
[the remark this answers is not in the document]
usmansaleem: You meant prefix+64 bytes key? I originally had it, but removed it during the recent refactoring. Let me add it back.

core/src/test/java/tech/pegasys/web3signer/core/service/jsonrpc/EthAccountsResultProviderTest.java
usmansaleem: fixed.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/EthPublicKeyUtils.java
usmansaleem: Updated to use a static final instance of SecureRandom in this class.

signing/src/main/java/tech/pegasys/web3signer/signing/secp256k1/EthPublicKeyUtils.java
```diff
@@ -12,63 +12,254 @@
...
+    }
+
+    byte[] publicKeyBytes = publicKeyValue.toByteArray();
+
+    // Ensure we have exactly 64 bytes
+    if (publicKeyBytes.length < 64) {
+      byte[] temp = new byte[64];
+      System.arraycopy(publicKeyBytes, 0, temp, 64 - publicKeyBytes.length, publicKeyBytes.length);
+      publicKeyBytes = temp;
+    } else if (publicKeyBytes.length > 64) {
```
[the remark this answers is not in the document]
usmansaleem: I've modified the method names to make the intent more clear. Also, refactored the logic slightly. Essentially we are dealing with Web3J Public Key represented as BigInteger. These BigInteger are created from uncompressed R+S bytes without prefix (64 bytes) from Bouncy Castle ECPoint.

signing/src/test/java/tech/pegasys/web3signer/signing/secp256k1/EthPublicKeyUtilsTest.java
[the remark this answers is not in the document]
usmansaleem: Test case updated.

jframe (approved):

PR #1040, opened 2024-11-22
Commit Boost Acceptance Test for List Pub Keys

usmansaleem:
## PR Description
Add commit boost acceptance test for list pub keys API. Also fix the public key identifiers used by commit boost by introducing K256ArtifactSigner.

## Documentation

## Changelog

## Testing


core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/signing/SignerForIdentifier.java
```diff
@@ -46,15 +46,13 @@ public Optional<String> sign(final String identifier, final Bytes data) {
   /**
    * Sign data for given identifier and return ArtifactSignature. Useful for SECP signing usages.
    *
-   * @param <T> The specific type of ArtifactSignature
    * @param identifier The identifier for which to sign data.
    * @param data Bytes which is signed
    * @return Optional ArtifactSignature of type T. Empty if no signer available for given identifier
    */
-  @SuppressWarnings("unchecked")
-  public <T extends ArtifactSignature<?>> Optional<T> signAndGetArtifactSignature(
```
jframe, line 55: Removing this means we need to cast in some places

usmansaleem: This is currently used only at one place, in `TransactionSerializer`.

usmansaleem: Let me try to see if we can eliminate this method altogether.

usmansaleem: @jframe I've eliminated this method. TransactionSerializer can reconstruct SignatureData from the hex String.

PR #1054, opened 2025-01-13
feat: Remove stale validator keys during reload

### LINKED ISSUE #1018: Reload endpoint to remove keys that have been deleted
Add support on the reload endpoint to remove keys that have been deleted from the filesystem when using the reload endpoint.

usmansaleem:
## PR Description
Remove stale validator keys during reload API call. The stale keys will not be listed in public keys API call after reload. They are also disabled in slashing database in eth2 mode. 

## Fixed Issue(s)

Fixes #1018

## Documentation

## Changelog

## Testing


signing/src/main/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProvider.java
```diff
@@ -62,20 +71,40 @@ public Future<Void> load() {
...
+          final Set<String> staleKeys;
+          if (reloadKeepStaleKeys) {
+            staleKeys = new HashSet<>();
+          } else {
+            staleKeys = new HashSet<>(oldSigners.keySet());
+            staleKeys.removeAll(signers.keySet());
+          }
+
+          // step 4: callback to register new keys and disable stale keys in slashing database
+          postLoadingCallback.ifPresent(callback -> callback.accept(signers.keySet(), staleKeys));
```
Matilda-Clerke, line 102: This comment is making an assumption about what happens in the callback, which means it could be incorrect in the future if the callback function is modified. It might be better to explain the callback function where it is defined and change this comment to just state that it's executing the post load callback if it's present

Matilda-Clerke (commented):
LGTM, just the one comment below

signing/src/main/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProvider.java
```diff
@@ -62,20 +71,41 @@ public Future<Void> load() {
...
-                        return signer1;
-                      }))
-              .forEach(signers::putIfAbsent);
-
-          // for each loaded signer, load commit boost proxy signers (if any)
+          // step 1: Create copy of current signers
+          final Map<String, ArtifactSigner> oldSigners = new HashMap<>(signers);
+          // step 2: Clear current signers and then load them via ArtifactSignerCollectionSupplier
+          if (!reloadKeepStaleKeys) {
+            signers.clear();
```
jframe, line 78: Doing this clear means there will be no signers loaded and will possibly fail some signing requests for keys aren't going to be unloaded. Can this be done without clearing the existing signers and just applying the delta to add and remove entries from the existing signer's map?

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/PostLoadingValidatorsProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+      return;
+    }
+
+    final RegisteredValidators registeredValidators =
+        slashingProtectionContext.getRegisteredValidators();
+    final Jdbi jdbi = slashingProtectionContext.getSlashingProtectionJdbi();
+    jdbi.useTransaction(
+        handle -> {
+          // disable the validators in the database
+          staleValidators.forEach(
```
jframe, line 66: Why are the stale validators disabled after removing from the signers map? This will affect all web3signer instances using the same database. The validator is already removed from the signers map so don't see why disabling them is necessary.

jframe (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
 
+## Next Release
+
+### Breaking Changes
+- The behavior of reload API endpoint has been modified due to issue [#1018][issue_1018] implemented by PR [#1054][pr_1054].
+The reload API call will remove all in-memory keys before loading them again using configuration files and bulk loading.
+This unfortunately results in behaviour change of local configuration files pointing to encrypted keystores. Previously 
+cached local configuration files were avoided being reloaded during reload API call, now they will be reloaded and 
+reparsed. To keep the old behavior `--reload-keep-stale-keys=true` option can be used which will not remove stale keys 
+during reload API call.
```
jframe, line 11: Is this still true after the latest changes?

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+
```
jframe, line 12: Should also add a note about the deprecated vertx pool size option being removed

CHANGELOG.md
usmansaleem: The wording need to change a bit, but we are still loading all the local keystores during the reload because otherwise we won't be able to figure out which keystores are removed. Let me go through the local keystores logic again.

signing/src/main/java/tech/pegasys/web3signer/signing/config/DefaultArtifactSignerProvider.java
```diff
@@ -25,35 +25,55 @@
...
+
+  private final boolean reloadKeepStaleKeys;
   private final Supplier<Collection<ArtifactSigner>> artifactSignerCollectionSupplier;
-  private final Map<String, ArtifactSigner> signers = new HashMap<>();
-  private final Map<String, Set<ArtifactSigner>> proxySigners = new HashMap<>();
-  private final ExecutorService executorService = Executors.newSingleThreadExecutor();
+  private final Optional<Consumer<Set<String>>> postLoadingCallback;
   private final Optional<KeystoresParameters> commitBoostKeystoresParameters;
 
+  private final ConcurrentMap<String, ArtifactSigner> signers = new ConcurrentHashMap<>();
```
jframe, line 65: What impact does switching to ConcurrentMap have on the startup time for Web3Signer when we have a lot of keys?

usmansaleem: I will create a performance test (outside of this PR) loading of about 5000 fully encrypted keystores load time at start up.

signing/src/main/java/tech/pegasys/web3signer/signing/config/SignerLoader.java
```diff
@@ -45,216 +47,204 @@
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+/**
+ * The SignerLoaded loads the metadata files and converts them to ArtifactSigners. This class keeps
```
jframe, line 51:
```suggestion
 * The SignerLoader loads the metadata files and converts them to ArtifactSigners. This class keeps
```

jframe (approved):

PR #1062, opened 2025-02-17
chore: Use info flag with owasp dependency to keep circleci job output

usmansaleem:
## PR Description
Use info flag with owasp dependency to keep circleci job output


jframe (approved):
LGTM

PR #1079, opened 2025-03-25
update to tuweni 2.7.0 

macfarla:
## Documentation

## Changelog

## Testing


macfarla:
I have read the CLA Document and I hereby sign the CLA

macfarla:
recheck

usmansaleem (changes requested):
We need to exclude `io.tmio` so that they are not resolved via transitive dependency. Running `./gradlew -q dependencyInsight --dependency com.nimbusds --configuration runtimeClasspath` will provide all the dependencies (in versions.gradle) that's bringing it in. For example: https://github.com/ConsenSys/web3signer/blob/9e8a90ee7a2d2dd9abf673cc199de8f8a4c1a474/gradle/versions.gradle#L97

macfarla:
@usmansaleem have updated with exclusions

PR #1087, opened 2025-06-05
Fix azure-tags option usage in AT

usmansaleem:
## PR Description
Fix azure-tags option usage in AT. Moved tags option in super class and removed Eth1 and Eth2 specific subclasses for Azure tag options.

## Documentation

## Changelog

## Testing


jframe (approved):
LGTM

PR #1094, opened 2025-06-19
Improve error messages in JacksonSerializers with input value context

menzudilat:
## PR Description
Improves error messages in JacksonSerializers by showing the actual problematic value when deserialization fails. Now instead of generic "Failed to decode" messages, developers will see exactly which value caused the error.

Added a helper method `safeGetValueAsString()` to handle value extraction safely, even when the parser is in an error state.
## Documentation

## Changelog

## Testing

### DIFF
--- a/common/src/main/java/tech/pegasys/web3signer/common/JacksonSerializers.java
+++ b/common/src/main/java/tech/pegasys/web3signer/common/JacksonSerializers.java
@@ -31,9 +31,12 @@ public class JacksonSerializers {
     @Override
     public Bytes deserialize(final JsonParser p, final DeserializationContext ctxt) {
       try {
-        return Bytes.fromBase64String(p.getValueAsString());
+        final String valueAsString = p.getValueAsString();
+        return Bytes.fromBase64String(valueAsString);
       } catch (final Exception e) {
-        throw new RuntimeException("Failed to decode as a Base64 string.", e);
+        final String valueAsString = safeGetValueAsString(p);
+        throw new RuntimeException(
+            String.format("Failed to decode '%s' as a Base64 string", valueAsString), e);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


common/src/main/java/tech/pegasys/web3signer/common/JacksonSerializers.java
```diff
@@ -31,9 +31,12 @@ public static class Base64Deserialiser extends JsonDeserializer<Bytes> {
     @Override
     public Bytes deserialize(final JsonParser p, final DeserializationContext ctxt) {
       try {
-        return Bytes.fromBase64String(p.getValueAsString());
+        final String valueAsString = p.getValueAsString();
+        return Bytes.fromBase64String(valueAsString);
```
usmansaleem, line 35: There is no point in extracting valueAsString as variable here. Also, we may not really want to print or log the invalid value at first place.

common/src/main/java/tech/pegasys/web3signer/common/JacksonSerializers.java
```diff
@@ -52,9 +55,12 @@ public static class HexDeserialiser extends JsonDeserializer<Bytes> {
     @Override
     public Bytes deserialize(final JsonParser p, final DeserializationContext ctxt) {
       try {
-        return Bytes.fromHexString(p.getValueAsString());
+        final String valueAsString = p.getValueAsString();
+        return Bytes.fromHexString(valueAsString);
```
usmansaleem, line 59: same as before, there is no point in extracting `valueAsString` here.

common/src/main/java/tech/pegasys/web3signer/common/JacksonSerializers.java
```diff
@@ -128,4 +142,18 @@ public void serialize(
...
+   * Safely get value as string from parser, handling potential exceptions.
+   *
+   * @param p JsonParser
+   * @return String representation of the value or placeholder if extraction fails
+   */
+  private static String safeGetValueAsString(final JsonParser p) {
+    try {
+      return p.getValueAsString();
+    } catch (Exception e) {
+      return "<unable to extract value>";
```
usmansaleem, line 156: This will eventually result in `Failed to parse <unable to extract value> as a UInt64` if p.getValueAsString fails.

usmansaleem:
@menzudilat Thank you for submitting this pull request. After reviewing the changes, we have determined that the modifications do not add substantial material or functional improvements to the codebase. While the changes introduce additional error handling and refactoring, they do not significantly enhance the existing functionality or address critical issues.

Additionally, we have concerns regarding the logging of invalid values, as this may inadvertently expose sensitive or critical data, potentially leading to security vulnerabilities. For this reason, we recommend avoiding logging raw input values unless necessary and ensuring that any logged data is appropriately sanitized.

We appreciate your effort and encourage you to continue contributing to the project. If you have further ideas or improvements that align with the project's goals, please feel free to submit another pull request.

Thank you for your understanding.

PR #1096, opened 2025-06-27
EIP-4844 signing support WIP

mskrzypkows:
I want to add EIP-4844 support. This signs now, but I have some mismatch of the `from` address when I do `send_raw_transaction` of the signed transaction (the nonce is not correct that's why I guess the address is wrong). Do you have some suggestions what's missing?

### DIFF
--- a/core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
+++ b/core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
@@ -41,7 +41,9 @@ public class TransactionSerializer {
   }
 
   public String serialize(final Transaction transaction) {
-    if (transaction.isEip1559()) {
+    if (transaction.isEip4844()) {
+      return toHexString(serializeEip4844(transaction));
+    } else if (transaction.isEip1559()) {
       return toHexString(serializeEip1559(transaction));
     } else {
       return toHexString(serializeFrontier(transaction));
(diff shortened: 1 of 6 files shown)


mskrzypkows:
I have read the CLA Document and I hereby sign the CLA

alexghr:
Hello, we're adding support for Web3Signer to the [Aztec node](https://github.com/AztecProtocol/aztec-packages/pull/16486) and would like to be able to sign EIP-4844 txs through its API. Any chance someone could take a look at this PR?

Thanks :)

core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java
```diff
@@ -59,6 +61,23 @@ private byte[] serializeFrontier(final Transaction transaction) {
...
+    byte[] bytesToSign = transaction.rlpEncode(null);
+    bytesToSign = prependEip4844TransactionType(bytesToSign);
+
+    final SignatureData signatureData = sign(transaction.sender(), bytesToSign);
+
+    byte[] serializedBytes = transaction.rlpEncode(signatureData);
+    return prependEip4844TransactionType(serializedBytes);
+  }
+
+  private static byte[] prependEip4844TransactionType(byte[] bytesToSign) {
```
usmansaleem: nit: final modifier in the method argument: `final byte[] bytesToSign)`

mskrzypkows:
@usmansaleem fixed, web3j 5.0.1 has been released, added usage of new version.

usmansaleem:
@mskrzypkows Can you add a changelog entry please (in CHANGELOG.md file).

alexghr:
Any updates on this? I'm wondering if this PR will have to be updated to support the post-Fusaka blob format (using cell proofs instead of blob proofs). The new format requires a new byte 0x01 to be inserted between the tx body and the blobs

https://eips.ethereum.org/EIPS/eip-7594#networking

Gabriel-Trintinalia:
https://github.com/Consensys/web3signer/issues/1103

usmansaleem:
Tested via https://github.com/usmansaleem/signers_docker_compose/tree/main/web3signer-eip-4844 @jframe probably need your review as I committed into this PR. LGTM.

Issue #1073, opened 2025-03-20
25.2.0 memory leak


jonanon20:
We have noticed since the upgrade from 24.12.0 to 25.2.0 a significant increase in memory usage that has caused several jvm OOM errors. Our web3signers are responsible for 10s of thousands of keys and this issue appeared after onboarding a new series of ~200keys leading to the OOM crash and missed attestations.

![Image](https://github.com/user-attachments/assets/086e65e0-d94c-4ec3-b630-79836f849be1)

You can see from the graph that with version 24.12.0 we didn’t have  this memory issue, as soon as we deployed 25.2.0 on 19/02 then memory usage spiked considerably. As a temporary measure we have set our  `-XX:MaxRAMPercentage=50` and given our nodes more memory, we generally see memory usage plateau but the plateau will only last a day or 2 before it climbs again. This gradual climb continues until we are alerted and have to restart. Originally we were at `-XX:MaxRAMPercentage=25` without any issues or memory spikes. Memory usage for the web3signer up until 25.2.0 has always been consistently low.

When we onboarded 200 new keys and we had missed attestations we noticed a considerable jvm garbage collection time as shown on the graph below across 2 of our 3 signers (jvm_gc_collection_seconds).  
![Image](https://github.com/user-attachments/assets/5883cc44-f90a-46ba-822f-579aabfec67e)

despite the long gc times we didnt see any reduction in memory for 2 of our signers during this period but this could be the result of memory pressure at the time.

We have noticed a worrying increase in memory and a gradual climb over several days. Where before 25.2.0 we were seeing a usage of ~1GB with no issues we now gradually climb beyond 6GB after 2 or 3 days.

usmansaleem:
@jonanon20 do you use the `/reload` API call or restart web3signer instances to onboard new keys?

jonanon20:
Yes that right, we use the /reload API

usmansaleem:
@jonanon20 Can you provide some more details for us to debug, such as:
- How Web3Signer is deployed, via docker image or via binary
- How the keys are loaded, using keystore files or via AWS/Azure?
- What is the total memory allocated to Web3Signer instance? Also, system memory/CPU.

Thanks/ Uzi

jonanon20:
[...]

PR #1167, opened 2026-04-15
Fix reload not offloading removed validators from slashing protection

usmansaleem:
Related: https://github.com/Consensys/web3signer/issues/1073

## Summary
- The `/reload` endpoint re-registered **all** validators on every call and never disabled removed validators in the slashing protection database or evicted them from the in-memory BiMap
- This caused ~19MB of unnecessary old gen heap allocation per reload for high key counts (~230MB/day at 12 reloads/day) and left stale validator entries that were never cleaned up
- Reload now computes a delta (added/removed keys), only registers new validators, disables removed validators in DB (`enabled=false`), and removes them from the BiMap. A previously-disabled row is **not** silently re-enabled on re-register — the reload path has no mechanism to supply fresh slashing-protection data, so explicit re-enable remains the responsibility of the keystore ADD API (which supports a `slashing_protection` import). Disable-on-remove is symmetric with the existing DELETE keystores behaviour introduced in #520.

## Changes
- **`DefaultArtifactSignerProvider`**: Changed `postLoadingCallback` from `Consumer<Set<String>>` to `BiConsumer<Set<String>, Set<String>>`. The `load()` method now computes the delta between old and new signer state and passes only added/removed keys to the callback.
- **`PostLoadingValidatorsProcessor`**: Updated to `BiConsumer`, registers only added validators and disables removed validators.
- **`RegisteredValidators`**: Added `disableAndRemoveValidators()` to batch-disable in DB and remove from BiMap. Introduced a dedicated `mutationLock` that serialises `registerValidators` and `disableAndRemoveValidators` against each other, preventing interleaved DB-vs-BiMap inconsistencies without blocking signing-path readers during DB I/O.
- **`ValidatorsDao`**: Added `setEnabledBatch()` for efficient batch enable/disable in a single SQL statement.
- **`Eth2Runner`**: Updated wiring with type witness for the `BiConsumer` callback.

## Test plan
- [x] Unit tests for delta computation in `DefaultArtifactSignerProviderTest` (first load, reload with changes, reload with same keys, reload with empty)
- [x] Unit tests for `disableAndRemoveValidators` in `RegisteredValidatorsTest` (BiMap removal, empty list no-op, unknown keys skipped)
- [x] Integration tests in `RegisteredValidatorsTest` (real DB via `DatabaseSetupExtension`): freshly-registered validators are enabled by the column default, and previously-disabled rows stay disabled after re-register.
- [x] DAO-level tests for `ValidatorsDao.setEnabledBatch` in `ValidatorsDaoTest` (empty-list no-op, batch disable, batch enable).
- [x] Full test suites pass: `:slashing-protection:test`, `:signing:test`, `:commandline:test`
- [ ] Acceptance test: `ReloadAcceptanceTest.reloadPicksUpFileSystemChanges` covers add/remove/reload scenario

## Empirical verification

Reproduced the leak and confirmed the fix using a docker-compose harness driving `/reload` cycles against two JDK-25 images built from the same Dockerfile: one from the tip of `master` (pre-fix baseline) and one from this branch (patched). JVM: `-Xmx3g -Xms2g`, G1GC.

### Partial-rotation workload (closer to real operator behaviour)

Each cycle removes 2500 random keys and adds 5000 new keys, so the expected working set grows by **+2500 per cycle** from an initial 5000.

| Cycle | Expected | Unpatched BiMap | Unpatched heap | Patched BiMap | Patched heap |
|------:|---------:|----------------:|---------------:|--------------:|-------------:|
| 0 | 5000  | 5000           | 45.5 MB | 5000  | 45.6 MB |
| 1 | 7500  | 10000 (+2500)  | 51.8 MB | 7500  | 53.1 MB |
| 2 | 10000 | 15000 (+5000)  | 64.3 MB | 10000 | 57.0 MB |
| 3 | 12500 | 20000 (+7500)  | 77.2 MB | 12500 | 65.1 MB |
| 4 | 15000 | 25000 (+10000) | 89.3 MB | 15000 | 68.9 MB |
| 5 | 17500 | 30000 (+12500) | **100.9 MB** | 17500 | **73.9 MB** |

At the same 17,500-key working set the unpatched build holds **27 MB more** than the patched build. The `HashBiMap$BiEntry` instance counts (30,000 unpatched vs 17,500 patched) match exactly the 5 × 2500 keys that were removed but never evicted on the unpatched build. The stale BiMap entries drag in ~17 MB of `byte[]` (BLS key material) that GC cannot reclaim, which accounts for most of the 27 MB delta.

### Full-replace workload (wipe + regenerate 5000 every cycle)

BiMap count should stay at 5000 across cycles if removed keys are correctly evicted.

| Cycle | Unpatched BiMap | Unpatched old-gen | Patched BiMap | Patched old-gen |
|------:|----------------:|------------------:|--------------:|----------------:|
| 0 | 5000  | 43.5 MB | 5000 | 43.4 MB |
| 5 | 22,994 | 61.2 MB | 5000 | 59.1 MB |

Patched build stays flat at 5000 BiMap entries; unpatched grows to 22,994 (**4.6×**).

### Artifacts captured per run

`summary.tsv`, per-cycle Prometheus `/metrics`, `jcmd GC.class_histogram`, `jstat -gc`, `jcmd GC.heap_info`, and `.hprof` dumps at cycle 0 and the final cycle — suitable for Eclipse MAT dominator-tree analysis if deeper inspection is needed.

---

> [!NOTE]
> **Medium Risk**
> Touches reload and slashing-protection DB mutation paths; incorrect delta or enable/disable handling could impact validator availability or safety, though changes are covered by new unit/integration tests.
> 
> **Overview**
> Fixes `/reload` to avoid re-registering all validators and leaking slashing-protection state by **processing only the delta** of validator keys.
> 
> `DefaultArtifactSignerProvider` now computes added/removed signer identifiers and invokes a `BiConsumer` callback; `PostLoadingValidatorsProcessor` uses this to register only added validators and **disable + evict removed validators** via new `RegisteredValidators.disableAndRemoveValidators()` (with batch DB updates through `ValidatorsDao.setEnabledBatch()` and serialized mutation locking). Tests were expanded to cover delta computation, batch enable/disable, and ensuring previously-disabled validators are not silently re-enabled on reload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fb0d2ba4e1e6094aea6ce990accae574497f1c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 26.4.1
+### Bugs Fixed
+- Fix reload endpoint not offloading removed validators from slashing protection database and in-memory cache. Reload now computes a delta (added/removed keys), disables removed validators in the DB, removes them from the BiMap, and re-enables previously disabled validators when their keys are re-added. Also eliminates unnecessary re-registration of all validators on every reload, reducing old gen heap pressure. PR [#1167][PR_1167].
```
jframe: Think message could be clearer that it fixes a memory leak with the reload endpoint and could also be more succinct we don't need details about using BiMap etc.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDao.java
```diff
@@ -74,6 +74,18 @@ public void setEnabled(final Handle handle, final int validatorId, final boolean
         .execute();
   }
 
+  public void setEnabledBatch(
```
jframe, line 77: Can you add a test for this

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/RegisteredValidators.java
```diff
@@ -85,13 +87,58 @@ public int mustGetValidatorIdForPublicKey(final Bytes publicKey) {
...
+    } finally {
+      validatorsLock.readLock().unlock();
+    }
+
+    if (knownValidators.isEmpty()) {
+      return;
+    }
+
+    // Batch disable in DB
+    final List<Integer> ids = knownValidators.values().stream().toList();
```
jframe: I think there is an edge case here where after getting the validator Ids registerValidators could be called and add the validator back since there is no lock being held when we do the DB update.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/RegisteredValidators.java
```diff
@@ -85,13 +87,58 @@ public int mustGetValidatorIdForPublicKey(final Bytes publicKey) {
...
+    jdbi.useTransaction(
```
jframe: This is change in behaviour for the reload endpoint. Before we didn't disable the validator at the database level. Is there a reason to do this? This will change the behaviour if multiple web3signers are using the same database. It might be a good thing need to think it through. Also if we had deleted a key and added it back again now the reload would reenable this possibly without any updated slashing protection data which could be dangerous.

CHANGELOG.md
usmansaleem: done.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/dao/ValidatorsDao.java
usmansaleem: 👍🏼

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/RegisteredValidators.java
usmansaleem: Added synchronisation between registration and removal.

slashing-protection/src/main/java/tech/pegasys/web3signer/slashingprotection/RegisteredValidators.java
usmansaleem: Good pick. I've modified it based on review points:

- `/reload` when new validators are added, they will be `enabled` in slashing-protection db by default (existing logic). There is no prior slashing data to consider.
- `/reload` when existing validators are removed, they will be `disabled`. This now matches existing `keystores API` delete functionality.
- `/reload` if validators keys which were previously offloaded, are re-added again, do not `enable` them in db. As you've noted, there is no slashing protection data associated with /reload. Users must use keystores API to to re add previously disabled validator which allows importing of slashing protection data

jframe (approved):

PR #1198, opened 2026-05-25
fix: verify keystore pubkey matches decrypted key on import

usmansaleem:
## PR Description

**`POST /eth/v1/keystores`** now validates that the keystore's JSON \`pubkey\` field matches the key derived by actually decrypting the ciphertext. A caller submitting a keystore with a mismatched pubkey receives \`status: "error"\` before any slashing-protection DB entry or signer registration occurs.

## Changes

- \`KeyStore.decrypt\` returns \`BLSKeyPair\` and validates the optional EIP-2335 \`pubkey\` field against the derived public key, throwing \`KeyStoreValidationException\` on mismatch. Previously returned raw \`Bytes\`.
- \`KeyStore.encrypt\` now accepts \`BLSKeyPair\` instead of separate private/public key \`Bytes\`, ensuring the stored pubkey is always cryptographically derived from the private key.
- \`DefaultValidatorManager\` replaces \`FileValidatorManager\` and \`InMemoryValidatorManager\`, unifying file-backed and in-memory-only validator lifecycle in a single class.
- \`DELETE /eth/v1/keystores\` now locates files by the \`<pubkey>.yaml\` naming convention rather than scanning all YAML files and matching on the untrusted \`pubkey\` field inside the keystore JSON. Before deleting, the keystore is decrypted and the derived pubkey is verified to match the requested key; a mismatch or decrypt failure leaves files untouched and logs a warning. See changelog for the breaking-change note.
- EIP-2335 model types (\`KeyStoreData\`, \`Cipher\`, \`Checksum\`, \`Kdf\`, \`KdfParam\`, etc.) converted from mutable classes with getters to Java records with eager validation in compact constructors; all call sites updated to record accessor syntax.
- Pubkey mismatch check moved from \`BLSKeystoreUtil\` into \`KeyStore\`; unit tests added directly to \`KeyStoreTest\` (\`decryptSucceedsWhenPubkeyIsAbsent\`, \`decryptThrowsExceptionWhenPubkeyMismatches\`).

## Documentation

## Changelog

## Testing

---

> [!NOTE]
> **High Risk**
> Changes authentication-critical keystore import/delete and slashing-protection registration paths, plus a breaking DELETE file-matching behavior.
> 
> **Overview**
> **Key Manager import** now decrypts each EIP-2335 keystore and derives the signer from the ciphertext; if the optional JSON `pubkey` does not match the derived key, import fails with an error status instead of registering under the wrong pubkey (e.g. slashing-protection poisoning).
> 
> **EIP-2335 layer:** `KeyStore.decrypt` returns `BLSKeyPair` and enforces pubkey consistency; `encrypt` takes `BLSKeyPair` and stores the derived compressed pubkey. Model types move to records with eager validation in compact constructors (replacing separate `validate()` chains). `CipherSpec` separates encryption-time cipher params from deserialized `Cipher`.
> 
> **Validator lifecycle:** `FileValidatorManager` / `InMemoryValidatorManager` are replaced by `DefaultValidatorManager` with optional `KeystoreFileManager`; `ValidatorManager.addValidator` takes `BlsArtifactSigner` + `KeystoreFileRecord`. Import path decrypts in `ImportKeystoresHandler` before persistence.
> 
> **DELETE keystores (breaking):** Deletion targets only `<pubkey>.yaml` metadata (not a directory scan matching JSON `pubkey`). Files are removed only after decrypting the on-disk keystore and confirming the derived pubkey matches the request.
> 
> Tests and call sites (bulk load, Commit Boost, acceptance helpers) are updated for the new APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11aa2fcac9b797e6fb5c1b905af37f6c762745f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


signing/src/main/java/tech/pegasys/web3signer/signing/ValidatorManager.java
```diff
@@ -12,11 +12,41 @@
...
 
   void deleteValidator(final Bytes publicKey);
 
-  void addValidator(final Bytes publicKey, final String keystore, final String password);
+  void addValidator(final BlsArtifactSigner signer);
+
+  void preAddValidator(
+      final BlsArtifactSigner signer, final String jsonKeystoreData, final String password);
+
+  default BlsArtifactSigner decryptKeystore(final String jsonKeystoreData, final String password) {
```
jframe: This doesn't need to be on the interface, it should need to be called from the `ImportKeystoresHandler`

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
```diff
@@ -186,22 +187,42 @@ private List<ImportKeystoreData> getKeystoreDataToProcess(
             i -> {
               final String jsonKeystoreData = requestBody.getKeystores().get(i);
               final String password = requestBody.getPasswords().get(i);
-              final String pubkey;
+
+              // Decrypt first — the derived pubkey is authoritative
+              final BlsArtifactSigner signer;
               try {
-                pubkey = parseAndNormalizePubKey(jsonKeystoreData);
+                signer = validatorManager.decryptKeystore(jsonKeystoreData, password);
```
jframe: There is no need to call a method on the validatorManager for this. We could just move the `decryptKeystoreAndCreateSigner` from the `FileValidatorManager` to this class and do the decrypt and pass the signer through like are doing.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
```diff
@@ -157,8 +157,9 @@ private void importValidators(final List<ImportKeystoreData> importKeystoreDataL
         .forEach(
             data -> {
               try {
-                final Bytes pubKeyBytes = Bytes.fromHexString(data.pubKey());
-                validatorManager.addValidator(pubKeyBytes, data.keystoreJson(), data.password());
+                validatorManager.preAddValidator(
```
jframe: Why is seperate preAddValidator step necessary? I think this complicating the API and it breaks the atomicity that we had before.

signing/src/main/java/tech/pegasys/web3signer/signing/ValidatorManager.java
```diff
@@ -12,11 +12,41 @@
...
+
+import java.util.Optional;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
 
 public interface ValidatorManager {
+  ObjectMapper getJsonMapper();
```
jframe: IMHO this doesn't belong on the interface, we could easily pass this into the constructor of the `ImportKeystoresHandler` where it's needed

signing/src/main/java/tech/pegasys/web3signer/signing/ValidatorManager.java
```diff
@@ -12,11 +12,41 @@
...
 
   void deleteValidator(final Bytes publicKey);
 
-  void addValidator(final Bytes publicKey, final String keystore, final String password);
+  void addValidator(final BlsArtifactSigner signer);
+
+  void preAddValidator(
```
jframe: Why is this needed? I think it was simpler before.

jframe (changes requested):

signing/src/main/java/tech/pegasys/web3signer/signing/ValidatorManager.java
usmansaleem: fixed, move to utility class.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
usmansaleem: fixed.

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
usmansaleem: removed.

signing/src/main/java/tech/pegasys/web3signer/signing/ValidatorManager.java
usmansaleem: agreed, removed.

signing/src/main/java/tech/pegasys/web3signer/signing/ValidatorManager.java
usmansaleem: yes...reverted to earlier logic.

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileRecord.java
```diff
@@ -0,0 +1,42 @@
...
+public record KeystoreFileRecord(String json, String password, String fileNameIdentifier) {
+  private static final String YAML_EXTENSION = ".yaml";
+  private static final String JSON_EXTENSION = ".json";
+  private static final String PASSWORD_EXTENSION = ".password";
+  public static final Set<String> KEYSTORE_FILE_EXTENSIONS =
+      Set.of(YAML_EXTENSION, JSON_EXTENSION, PASSWORD_EXTENSION);
+
+  public KeystoreFileRecord {
+    Objects.requireNonNull(json, "Keystore json must not be null");
+    Objects.requireNonNull(json, "Keystore password must not be null");
```
jframe: think this should be
```suggestion
    Objects.requireNonNull(password, "Keystore password must not be null");
```

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileRecord.java
```diff
@@ -0,0 +1,42 @@
...
+    Objects.requireNonNull(json, "Keystore fileNameIdentifier must not be null");
```
jframe: think this should be
```suggestion
    Objects.requireNonNull(fileNameIdentifier, "Keystore fileNameIdentifier must not be null");
```

signing/src/main/java/tech/pegasys/web3signer/signing/util/BLSKeystoreUtil.java
```diff
@@ -0,0 +1,59 @@
...
+  private BLSKeystoreUtil() {}
+
+  public static BlsArtifactSigner decryptKeystore(
+      final ObjectMapper jsonMapper, final String jsonKeystoreData, final String password) {
+    try {
+      final KeyStoreData keyStoreData = jsonMapper.readValue(jsonKeystoreData, KeyStoreData.class);
+      final Bytes privateKey = KeyStore.decrypt(password, keyStoreData);
+      final BLSKeyPair keyPair = new BLSKeyPair(BLSSecretKey.fromBytes(Bytes32.wrap(privateKey)));
+
+      // validate pubKey (if present) - both are Bytes, use compareTo directly
```
jframe: This comment belongs in throwIfPubkeyMismatch?

signing/src/main/java/tech/pegasys/web3signer/signing/util/BLSKeystoreUtil.java
```diff
@@ -0,0 +1,59 @@
...
+      return new BlsArtifactSigner(
+          keyPair, SignerOrigin.FILE_KEYSTORE, Optional.ofNullable(keyStoreData.getPath()));
+    } catch (final JsonProcessingException e) {
+      throw new KeyStoreValidationException("Failed to parse keystore JSON", e);
+    }
+  }
+
+  private static void throwIfPubkeyMismatch(
+      final Bytes decryptedPublicKey, final Bytes claimedPubkey) {
+    if (decryptedPublicKey.compareTo(claimedPubkey) != 0) {
```
jframe: Why is compareTo needed? Could `!decryptedPublicKey.equals(claimedPubkey)` be used instead? It would be clearer.

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileRecord.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public String yamlFileName() {
```
jframe: Rather than naming this after the file extension I think it would be better to name this after what it represents. So maybe something like `metadataFileName` instead?

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileRecord.java
```diff
@@ -0,0 +1,42 @@
...
+    return fileNameIdentifier + YAML_EXTENSION;
+  }
+
+  public String jsonFileName() {
```
jframe: Same here, maybe `keystoreFileName`?

signing/src/main/java/tech/pegasys/web3signer/signing/DefaultValidatorManager.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+package tech.pegasys.web3signer.signing;
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+
+import org.apache.tuweni.bytes.Bytes;
+
+/// Default [`ValidatorManager`] implementation that supports both file-backed and in-memory-only
```
jframe: I'm getting a warning here and on line 24 about an illegal character U+0060

signing/src/main/java/tech/pegasys/web3signer/signing/util/BLSKeystoreUtil.java
usmansaleem: stale comment, removed.

signing/src/main/java/tech/pegasys/web3signer/signing/util/BLSKeystoreUtil.java
usmansaleem: I incorrectly thought that Bytes.equals might not work between different subclasses of Bytes (i.e. `Bytes48.equals(Bytes)`, however, equals works as expected. I've removed compareTo logic.

signing/src/main/java/tech/pegasys/web3signer/signing/DefaultValidatorManager.java
usmansaleem: yes fixed. It was due to using both `[ and backquote`.

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileRecord.java
usmansaleem: done

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileRecord.java
usmansaleem: done

signing/src/main/java/tech/pegasys/web3signer/signing/util/BLSKeystoreUtil.java
```diff
@@ -0,0 +1,45 @@
...
+public final class BLSKeystoreUtil {
+  private BLSKeystoreUtil() {}
+
+  public static BlsArtifactSigner decryptKeystore(
+      final ObjectMapper jsonMapper, final String jsonKeystoreData, final String password) {
+    final KeyStoreData keyStoreData = parseKeystoreJson(jsonMapper, jsonKeystoreData);
+
+    final BLSKeyPair keyPair = KeyStore.decrypt(password, keyStoreData);
+
+    return new BlsArtifactSigner(keyPair, SignerOrigin.FILE_KEYSTORE, keyStoreData.path());
```
jframe: nit: This util method isn't really general purpose as returns a signer that has a file keystore origin.

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileManager.java
```diff
@@ -98,49 +76,23 @@ public void createKeystoreFiles(
...
-                    } catch (final Exception e) {
-                      LOG.error("Error reading config file: {}", path, e);
-                      return null;
-                    }
-                  })
-              .filter(Objects::nonNull)
-              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-      // return the matching file
-      return Optional.ofNullable(map.get(pubkey));
+  private Optional<KeystoreFiles> findKeystoreConfigFiles(final String pubkey) throws IOException {
```
jframe: How is this safer? Before we were trusting that pubKey field was correct which was wrong, but this is similarly just trusting the filename is correct. The safest would be to decrypt the keystore and only delete keystores that match the decrypted pubkey.

bls-keystore/src/main/java/tech/pegasys/web3signer/bls/keystore/model/KeyStoreData.java
```diff
@@ -16,84 +16,43 @@
...
-    return crypto;
-  }
-
-  public Bytes getPubkey() {
-    return pubkey;
-  }
+public record KeyStoreData(
+    @JsonProperty(value = "crypto", required = true) Crypto crypto,
+    @JsonProperty(value = "path", required = true) String path,
+    @JsonProperty(value = "uuid", required = true) UUID uuid,
```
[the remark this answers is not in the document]
usmansaleem: Reverted so that only verify the bare minimum needed. Also converted to `java record`.

bls-keystore/src/main/java/tech/pegasys/web3signer/bls/keystore/model/KeyStoreData.java
```diff
@@ -16,84 +16,43 @@
...
+   * Compact Constructor: Automatically executes validation for BOTH Jackson deserialization and
+   * manually invoked constructors.
+   */
+  public KeyStoreData {
+    if (version == null || version != KEYSTORE_VERSION) {
+      throw new KeyStoreValidationException("KeyStore version not supported: " + version);
+    }
+    if (crypto != null) {
+      crypto.validate();
     }
```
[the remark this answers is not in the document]
usmansaleem: Removed `validate` from model. Using `record` compact constructors to perform validation when constructing.

signing/src/main/java/tech/pegasys/web3signer/signing/util/BLSKeystoreUtil.java
usmansaleem: Deleted. Refactored to use existing logic.

signing/src/main/java/tech/pegasys/web3signer/signing/KeystoreFileManager.java
usmansaleem: Added check to decrypt/verify before deleting local keystore files.

acceptance-tests/src/test/java/tech/pegasys/web3signer/dsl/utils/MetadataFileHelpers.java
```diff
@@ -274,18 +268,14 @@ private void createPasswordFile(final Path passwordFilePath, final String passwo
...
+      final Path keyStoreFilePath, final BLSKeyPair blsKeyPair, final KdfFunction kdfFunctionType) {
     final KdfParam kdfParam =
         kdfFunctionType == KdfFunction.SCRYPT
             ? new SCryptParam(32, SALT)
             : new Pbkdf2Param(32, 262144, HMAC_SHA256, SALT);
-    final Cipher cipher = new Cipher(CipherFunction.AES_128_CTR, IV);
+    final Cipher cipher = new Cipher(CipherFunction.AES_128_CTR, IV, Bytes.EMPTY);
     final KeyStoreData keyStoreData =
-        KeyStore.encrypt(privateKey, publicKey, password, "m/12381/3600/0/0/0", kdfParam, cipher);
+        KeyStore.encrypt(blsKeyPair, "password", "m/12381/3600/0/0/0", kdfParam, cipher);
```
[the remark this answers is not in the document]
usmansaleem: test code.

signing/src/main/java/tech/pegasys/web3signer/signing/DefaultValidatorManager.java
```diff
@@ -0,0 +1,93 @@
...
+  public void deleteValidator(final Bytes publicKey) {
+    try {
+      // Remove active key from memory first, will stop any further signing with this key
+      signerProvider.removeSigner(publicKey.toHexString()).get();
+      // Then, delete the corresponding keystore files
+      if (keystoreFileManager.isPresent()) {
+        final boolean filesDeleted =
+            keystoreFileManager.get().deleteKeystoreFiles(publicKey.toHexString());
+        if (!filesDeleted) {
+          LOG.warn(
```
jframe: Maybe we should throw an exception here instead so that the `DeleteKeystoreProcessor` returns an error when this happens instead returning a deleted status

core/src/main/java/tech/pegasys/web3signer/core/service/http/handlers/keymanager/imports/ImportKeystoresHandler.java
```diff
@@ -226,12 +237,10 @@ private static boolean imported(ImportKeystoreData data) {
   }
 
   private static boolean failed(ImportKeystoreData data) {
+    // signer is null when decryptKeystore() threw (bad password, corrupt data, or pubkey mismatch)
+    // — addValidator was never called in those cases, so there is nothing to clean up
     return data.importKeystoreResult().getStatus() == ImportKeystoreStatus.ERROR
-        && data.pubKey() != null;
-  }
-
```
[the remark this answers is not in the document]
usmansaleem: no action required.

jframe (approved):

PR #1228, opened 2026-09-21
Add -X/--Xhelp to list unstable CLI options

usmansaleem:
## PR Description

Web3Signer declares its unstable CLI options (option names beginning `--X`) as `hidden = true`, so they never appear in `--help` and are effectively undiscoverable from the CLI. This adds a hidden picocli help subcommand `Xhelp` (aliases `-X`, `--Xhelp`) that walks the command tree and prints every unstable option, grouped by the command that owns it.

### Usage

| Invocation | Output |
|---|---|
| `web3signer -X` / `web3signer --Xhelp` / `web3signer Xhelp` | all 16 unstable options, under `Unstable options for web3signer eth2:` and `... eth1:` headings |
| `web3signer eth2 -X` | only the 13 options owned by `eth2` |
| `web3signer eth1 -X` | only the 3 options owned by `eth1` |

Output goes to stdout with exit code 0, and the `Starting Web3Signer version ...` banner is suppressed. The listing is prefixed by a one-time notice stating that no stability or compatibility guarantee applies to unstable options.

`--help` output is unchanged apart from one footer line pointing at the listing (root, `eth2` and `eth1`). The `Xhelp` command itself stays `hidden = true`, so the `Commands:` table is not polluted, and `web3signer eth2 export -X` remains a parse error (`Unknown option: '-X'`, exit 2) since the export/import/watermark-repair subcommands own no unstable options.

```
$ web3signer eth1 -X
No stability or compatibility guarantee applies to unstable options. They may be added, changed or removed between releases without announcement.

Unstable options for web3signer eth1:
      --Xazure-endpoint-override=<URI>
         Override Azure Key Vault endpoint.
      --Xazure-authority-host-override=<URI>
         Override the Microsoft Entra ID (Azure AD) authority host.
      --Xazure-trust-certificate-override=<FILE>
         Trust the given X.509 certificate file for TLS connections to the
           Azure Key Vault endpoint and Microsoft Entra ID authority, in place
           of the platform default trust store.
```

### Implementation notes

- `XHelpSubCommand` implements `IHelpCommandInitializable2` rather than the deprecated v1 interface so output is written to the parser's `PrintWriter` (capturable in tests) instead of `System.out`, and obtains its scope root from `self.getCommandSpec().parent()` so the same class serves root-level and per-mode invocation under plain annotation registration.
- The command is registered via `@Command(subcommands = ...)` on `Web3SignerBaseCommand`, `Eth2SubCommand` and `Eth1SubCommand`. Programmatic `addSubcommand` would not work: `CommandLine.setOut` only propagates to subcommands registered at the time the setter is called, and `parseCommandLine` calls it after its `addSubcommand` loop, so a later-registered command would write to `System.out`.
- Each subtree is scanned with `CommandSpec.options()`, which already contains options flattened in from `@Mixin` fields, so a single scan per spec covers both directly declared and mixin-contributed options.
- `sortOptions(false)` preserves declaration order, matching the root command's `sortOptions = false`.
- `CommandlineParser.isHelpOrVersionRequested` now uses `commandSpec().helpCommand()` instead of a `"help"` string comparison, so the startup banner is suppressed for `Xhelp` as well.
- Option descriptions are not prefixed with a marker: the per-command heading plus the one-time stability notice already carry that signal.

## Documentation

A new CLI command is user-facing, so `doc-change-required` is applied: the CLI reference in `doc.web3signer` should gain a section for `-X` / `--Xhelp`.

## Changelog

## Testing

- `./gradlew :commandline:test` — full suite green, including the new `XHelpSubCommandTest` (aliases, heading scoping, exact `eth1 -X` output, declaration order, `--help` equality against a reference build, footer pointer, and the `eth2 export -X` rejection).
- `./gradlew installDist` plus real-binary runs of `web3signer -X`, `web3signer --Xhelp`, `web3signer eth2 -X`, `web3signer eth1 -X` (exit 0, stdout only, no banner) and `web3signer eth2 export -X` (exit 2).
- `./gradlew spotlessCheck` clean.

---

> [!NOTE]
> **Low Risk**
> CLI help and discoverability only; no signing, auth, or runtime behavior changes beyond help output and banner suppression for `-X`.
> 
> **Overview**
> Adds a hidden picocli help subcommand (`-X` / `--Xhelp` / `Xhelp`) so operators can discover **unstable** CLI flags (`--X…`) that stay omitted from normal `--help`. The walker prints those options **grouped by command** (root lists `eth1` and `eth2`; `web3signer eth2 -X` scopes to `eth2` only), with a stability disclaimer and declaration order preserved.
> 
> Standard `--help` for the root, `eth1`, and `eth2` gains a footer pointing at the `-X` listing; `Xhelp` itself stays hidden from the commands table. `CommandlineParser` treats any `helpCommand` (including `Xhelp`) like help for **startup banner suppression**. Several hidden Azure/Key Manager option descriptions drop the **EXPERIMENTAL** prefix in favor of unstable wording; changelog and docker docs are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c128da3f2537ddb413a991f58a156d42ddcefb1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


commandline/src/main/java/tech/pegasys/web3signer/commandline/subcommands/XHelpSubCommand.java
```diff
@@ -0,0 +1,99 @@
...
+    aliases = {"-X", "--Xhelp"},
+    description = "This command provides help text for all early access options.",
+    hidden = true,
+    helpCommand = true)
+public class XHelpSubCommand implements Runnable, CommandLine.IHelpCommandInitializable2 {
+
+  public static final String COMMAND_NAME = "Xhelp";
+  private static final String EARLY_ACCESS_OPTION_PREFIX = "--X";
+  private static final String STABILITY_NOTICE =
+      "Early access options are not covered by any stability or compatibility guarantee. "
```
jframe: Don't think we should be calling these early access options. There should be no expectation they will eventually become full options. Calling them unstable or experimental options would be more accurate.

jframe (approved):

PR #815, opened 2023-06-22
Fix cves by forcing upgrade of netty-handler and bouncycastle

siladu:
## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

## Testing

- [ ] I thought about testing these changes in a realistic/non-local environment.

PR #1231, opened 2026-09-23
Upgrade dependencies and build plugins to latest releases

usmansaleem:
## PR Description

Upgrades all pinned dependencies and build plugins to their latest release versions, based on `./gradlew dependencyUpdates --no-parallel`.

Prerelease-only "updates" were deliberately not taken: log4j 3.0.0-beta*, assertj 4.0.0-M1, wiremock 4.0.0-beta.38, azure-identity 1.19.0-beta.2, azure-security-keyvault-keys 4.12.0-beta.1.

### Library versions (`gradle/versions.gradle`)

- **Vert.x** 4.5.26 -> 5.2.0, **web3j** 5.0.2 -> 6.0.0, **Flyway** 11.20.3 -> 13.7.0, **MockServer** 5.15.0 -> 8.0.0
- **JUnit** 5.14.3 -> 6.1.3, **rest-assured** 5.5.7 -> 6.0.1, **okhttp** 4.12.0 -> 5.5.0, **Jackson** 2.21.2 -> 2.22.3
- **AWS SDK** 2.42.37 -> 2.55.3, **Besu** 26.8.0 -> 26.8.1, **log4j** (bom) 2.25.4 -> 2.26.1, **netty** (bom) 4.2.17.Final -> 4.2.18.Final
- **guava** 33.6.0-jre -> 33.7.1-jre, **bouncycastle** 1.84 -> 1.86, **tuweni** 2.7.2 -> 2.8.0, **errorprone** 2.49.0 -> 2.50.0
- **jdbi** 3.52.1 -> 3.54.0, **HikariCP** 7.0.2 -> 7.1.0, **postgresql** 42.7.10 -> 42.7.13, **embedded-postgres-binaries-bom** 17.9.0 -> 18.6.0
- **google-cloud-secretmanager** 2.91.0 -> 2.97.0, **azure keyvault keys/secrets** 4.10.6 -> 4.11.2, **azure-identity** 1.18.2 -> 1.18.6
- **jblst** 0.3.15 -> 0.3.18, **caffeine** 3.2.3 -> 3.3.0, **commons-logging** 1.3.6 -> 1.4.0, **zt-exec** 1.12 -> 1.13.0, **okio** 3.17.0 -> 3.18.2
- **reactor-netty-http** 1.3.5 -> 1.3.7, **nimbus-jose-jwt** 9.48 -> 10.10, **java-diff-utils** 4.16 -> 4.17

`caffeine` moves out of `signing/build.gradle` into the catalogue so every version lives in one place.

The `netty-handler` exclusion on `vertx-core` is removed. It was added in #815 together with an explicit `netty-handler` pin for CVE-2023-34462; `netty-bom` has since replaced that pin and manages every `io.netty` module, so the exclusion no longer did anything useful. The shipped `:app` runtime classpath is identical with or without it; the only difference is that `:bls-keystore` no longer loses `netty-handler` from its own classpaths. The unused `vertx-codegen` and `vertx-unit` catalogue entries are removed as well.

### Build plugins (`build.gradle`)

OWASP dependency-check 12.2.1 -> 13.0.0, Spotless 8.4.0 -> 8.10.2, errorprone plugin 4.4.0 -> 5.1.1, google-java-format 1.35.0 -> 1.36.1, jacoco 0.8.14 -> 0.8.15, ben-manes versions 0.54.0 -> 0.64.0, dependency-license-report 3.1.2 -> 3.1.4. Already current: spring dependency-management 1.1.7, jmh 0.7.3, grgit 5.3.3, foojay 1.0.0, download 5.7.0, osdetector 1.7.3, license-reporter 1.2.0.

The Spotless bump rewrapped Markdown Javadoc in `DefaultValidatorManager`/`ValidatorManager`; those formatting-only changes are included in the same commit.

### Code changes required by the majors

**Vert.x 5**
- `PoolMetricsAdapter` implements `PoolMetrics<Q, T>` with `enqueue`/`dequeue`/`begin`/`end`; `VertxMetricsAdapter` returns `PoolMetrics<?, ?>`.
- `HttpServer.listen()` and `Vertx.close()` now return `Future`; `executeBlocking` returns `Future`, and callbacks that only handle failure use `onFailure`.
- Key/trust option setters unified: `setPfxKeyCertOptions` (already deprecated in 4.x, where it delegated to `setKeyCertOptions`) -> `setKeyCertOptions`.
- Internal `HeadersMultiMap` replaced with `MultiMap.caseInsensitiveMultiMap()`.
- `HealthCheckHandler` moved from `io.vertx.ext.healthchecks` to `io.vertx.ext.web.healthchecks` (import-only change).
- Routes that wrapped handlers in the internal `io.vertx.ext.web.impl.BlockingHandlerDecorator` now call the public `Route.blockingHandler(handler, false)`, which installs that same decorator; no behaviour change.

**web3j 6**
- `FileBasedSignerFactory` and `Secp256k1ArtifactSignerFactory` also catch the now-unchecked Jackson 3 `JacksonException`, so a missing or malformed keystore still surfaces as `SignerInitializationException`/`SigningMetadataException` instead of a raw `tools.jackson` exception escaping the declared `IOException`.
- EIP-4844 `eth_signTransaction` returns the canonical signed transaction `0x03 || rlp([...14 fields...])`. web3j 5 wrapped the signed payload in a blob network-wrapper list plus three empty sidecar lists, which was not a valid EIP-4844 signed transaction. The signing preimage, and therefore r/s, is unchanged; the fixture in `EthSignTransactionResultProviderTest` was updated accordingly.

**Jackson 3 pin**
`tools.jackson:jackson-bom:3.1.5` is imported as a BOM. web3j 6 parses keystore files relying on `As.EXTERNAL_PROPERTY`, which Jackson databind 3.2.1 no longer honours: the sibling type-id property is left unset, `WalletFile.Crypto.kdf` deserialises to `null`, and every V3 keystore load fails. MockServer 8.0.0 pulls `jackson-bom:3.2.1` transitively via `com.networknt:json-schema-validator`, so without this pin the integration-test classpath (and any consumer resolving >= 3.2.1) breaks keystore loading.

### Behaviour changes worth calling out

- `vertx_worker_pool_rejected_total` is no longer exported. Vert.x 5's pool metrics SPI replaced `submitted`/`rejected` with `enqueue`/`dequeue`, and `dequeue` fires both when a queued task starts executing and when it is rejected, so a rejection can no longer be told apart. (Vert.x's own Micrometer implementation never exported a rejected counter.) In both 4.x and 5.x a worker task is only rejected after its pool has been shut down, so the counter was 0 during normal operation. The shipped Grafana dashboard's "Worker Pool Rejections" panel is replaced by "Worker Pool Queue Delay (p95)" (`http_vertx_worker_queue_delay`), which shows worker-pool saturation.
- Vert.x 5 also instruments its virtual-thread worker pool, so an idle `poolName="vert.x-virtual-thread"` series now appears alongside the existing worker-pool series.
- Jackson 2 and Jackson 3 now coexist on the classpath: web3j 6 and Vert.x 5 (via Besu) bring Jackson 3. Vert.x itself keeps using Jackson 2, though. `vertx-core` 5.2.0 is a multi-release jar whose Java 21+ `JacksonFactory` tries the Jackson 2 `DatabindCodec` first and only falls back to its Jackson 3 codec when Jackson 2 databind is missing. Web3Signer ships `jackson-databind` 2.22.3, so `Json.CODEC` resolves to `io.vertx.core.json.jackson.DatabindCodec` (checked at runtime against the installed distribution), and Vert.x JSON encoding/decoding (`Json.encode`, `JsonObject.mapFrom`/`mapTo`, `DecodeException`/`EncodeException`) is unchanged from Vert.x 4. Migrating Web3Signer's own `ObjectMapper` usage to Jackson 3 is intentionally left to a separate PR; once Jackson 2 databind leaves the classpath, Vert.x switches to its Jackson 3 codec, and that switch needs its own verification.

## Documentation

## Changelog

Changelog entry to be added by the author.

## Testing

- `./gradlew build` - 740 unit tests (85 classes), 0 failures
- `./gradlew releaseAcceptanceTest` - 47 classes, 316 tests, 0 failures (16 skipped: AWS/GCP credentials unavailable)
- `./gradlew releaseIntegrationTest` - 20 classes, 124 tests, 0 failures
- `./gradlew dependencyUpdates --no-parallel` - all libraries now report as latest release versions
- Vert.x 5 follow-up (`e0d1c3f1`): `./gradlew build` (above), `:core:integrationTest` (61 tests, 0 failures), and an acceptance subset covering every rewired route (`KeyIdentifiers`, `BlsSigning`, `SecpSigning`, `ListKeys`, `DeleteKeystores`, `CommitBoostGetPubKeys`: 133 tests, 0 failures, 4 skipped for missing cloud credentials). A smoke run of the distribution confirmed that the rewired routes still dispatch to the worker pool and that `http_vertx_worker_queue_delay{quantile="0.95"}` is exported for the updated dashboard panel.
- A forced recompile of all source sets with `-Xlint:deprecation,removal` reports no deprecated or for-removal Vert.x API usage.

Two pre-existing conditions, not introduced here: `:app:javadoc` fails under doclint `-Werror` on an unmodified file at baseline HEAD (and is not part of `build`), and `ValueTransferAcceptanceTest.multipleValueTransfers` is intermittently red (1 of 4 full-suite runs; 6/6 green in isolation) because it submits 50 null-nonce transactions without awaiting.


signing/src/test/java/tech/pegasys/web3signer/signing/config/metadata/parser/YamlSignerParserTest.java
```diff
@@ -242,7 +242,8 @@ void keyStoreMetaDataInfoWithHttpURIFailsWithException() throws IOException {
 
     Assertions.assertThatThrownBy(
             () -> signerParser.parse(signerParser.readSigningMetadata(yamlMetadata)))
-        .hasRootCauseMessage("Provider \"https\" not installed");
+        .isInstanceOf(SigningMetadataException.class)
+        .hasMessageStartingWith("Invalid signing metadata file format");
```
jframe: Why is this failing with a different error? We should still be failing due http being used in the path.

usmansaleem: caused by Jackson version upgrade. The previous root cause message was `Provider "https" not installed`. The new root cause message is:
```
Cannot deserialize value of type `java.nio.file.Path` from String "https://consensys.net/keystore.json": scheme 'https' not
allowed for Path deserialization (allowed: ["file"])
 at [Source: (StringReader); line: 4, column: 7] (through reference chain:
tech.pegasys.web3signer.signing.config.metadata.FileKeyStoreMetadata["keystoreFile"])
```

I've updated the test assertion.

jframe (approved):

